PR #3, opened 2016-05-05
Fix/chef provisioning

jessesiu:
add the api controller and views
change the faq page


pli888:
There is now a new branch called new-feature/api which should be used to commit the GigaDB API code into the gigadb-website repo. Using the new-feature/api branch will keep the API code separate from the fixes which have been made to get the Chef provisioning code to work.

PR #20, opened 2016-05-12
Fix/chef provisioning

jessesiu:
Datasets in carousel


pli888:
Your code commit that fixes the carousel datasets is mixed with 3 other commits that are to do with the GigaDB API and the FAQ page. These 3 code commits have already been added for your API work.

We want to have source code for new features or bug fixes committed into specific branches as described in the BRANCHING.md file. Please create a new branch from the `develop` branch and call this new branch `fix/carousel-datasets` and only check in the `Datasets in carousel` commit into this branch.

Issue #42, opened 2016-08-18
fix dataset-file delete function


only1chunts:
In the admin pages it should be possible to delete files from the dataset-files table:
http://gigadb.org/adminFile/admin
The delete button is present, but when I click it, there is an error 500 message and the file is not deleted.

Issue #9, opened 2016-05-10
Fix File-Sample display


only1chunts:
It is possible for a file to linked to 0, 1 or more samples, the current database schema is fine with that. However, the current dataset webpages are not correct when there is >1 sample associated to a file.
See for example:
[http://gigadb.org/dataset/100053](http://gigadb.org/dataset/100053)
This dataset actually only has 9 files in it, but the file table is displaying 41 files! this is because its showing the same file name again for every different sample its associated to (which is wrong).
If a file has multiple Samples associated to it, it should still only appear in the file table once. The File table should list each file with its associated sample ID's in the column for samples, if >1 then there should be upto 3 displayed with a "+" to allow to display more.

PR #43, opened 2016-08-19
fixes issue#42 and issue#9

jessesiu:


only1chunts:
@jessesiu I think these need to be split into separate pull requests, every update should be done individually to enable complete transparency in tracking.
#42 appears to be fixed in the public server - I've not checked on a VM
#9 -not checked yet.

only1chunts:
@jessesiu , Issue #42 still exists when I try on the VM , i.e. I still get an error 500 when I try to delete files using the admin pages.
I followed the instructions to fetch the "develop" branch, but perhaps I've done something wrong?! or the fix hasn't worked?
I've not checked #9 yet because I cant remember how to upload the database to look at the example dataset.

only1chunts:
@jessesiu , Its OK, I was looking at the gigascience:develop branch, not the jessesiu:develop branch. Now that I'm looking at the correct branch:
Issue #42 is fixed.
Issue #9 is part fixed. The linked samples now displays multiple samples which is the important bit, but there is no limit set to how many are shown, i.e. if more than 3 add the "+" button for users to expand the view if required (same as sample attributes table).

jessesiu:
@only1chunts  the commit fixes issue #43 will only show 3 linked samples and the + button (same as sample attributes table)

pli888:
@jessesiu Can you check the conflict problem with your chef/chef-cookbooks in this merge please?

jessesiu:
@pli888  Fixed the conflict problem

PR #119, opened 2017-02-15
Develop

jessesiu:
Fix some issues


pli888:
@jessesiu When using vagrant to test your code on a local machine It's a bit confusing that the help page API examples are still showing URL calls displayed with the gigadb.org domain name, e.g. http://gigadb.org/api/dataset?doi=100051 but it calls the API on the locahost?

Also when there are no results available for an API call, you get a message containing plain/HTML text, e.g. `No items where found for dataset doi <b>100051</b>` being returned. I think the client calling your API would expect XML to be returned otherwise it might break their application code.

When submitting pull requests, its clearer if you submit one pull request for issue.

@only1chunts Can you check Jesse's other fixes that they do what you need?

jessesiu:
The new commit create the vagrant template to create the help.html page. I will send another pull request for changing the API no results return format.

only1chunts:
all 3 issues are fixed in jesse/gigadb-website, checked 23-Mar-2017

PR #127, opened 2017-09-15
Develop

jessesiu:
two changes in the dataset view page

### DIFF
--- a/protected/views/dataset/view.php
+++ b/protected/views/dataset/view.php
@@ -149,6 +149,7 @@ HTML;
                     <?php  
                         $types = array();
                         $protocol = array();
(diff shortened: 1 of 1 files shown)


pli888:
@jessesiu I have merged your changes into develop but the datasets generate an "Error 500
Trying to get property of non-object" error message when you click on them. I think it might be a problem with the test database. Please checkout this [branch](https://github.com/gigascience/gigadb-website/tree/jessesiu-develop) and fix this problem.

jessesiu:
@pli888 I tested it, It seems the browse cookies contain the gigadb users info. You can log out, then log in with the username: admin@gigadb.org.

PR #144, opened 2017-12-12
Develop tc issue130

bobbledavidson:
Using the CCaptcha system in the Yii Framework, i've created an on-the-fly Captcha system for the contact page that doesn't rely on calling pictures from a remote server (presumably the problem with Chinese access to existing system.)

There was a problem with Yii Framework 1.1 in that it cannot create an image directly so I've included some PHP code to 'paint' the Captcha image when the web form is called. Updating the Yii Framework could solve this but I didn't want to cause problems with existing systems.

There is need for a /images/tempcaptcha/ folder for storing these painted Captcha images. If the user completes the webform, then the .PNG is deleted. I am working on a cleanup system for the use-case where the user navigates away from the contact page without submitting the form. Should have that with you tomorrow. 

For this image manipulation, you need to install some PHP libraries (e.g. ImageMagick) - I've provided an INSTALL file with instructions for doing this manually, figuring that would be easiest for you to review, rather than incorporating it in Chef/Vagrant files. It was also quicker for me to write it up this way. If you want me to add these steps to provisioning scripts, let me know.


pli888:
> For this image manipulation, you need to install some PHP libraries (e.g. ImageMagick) - I've provided an INSTALL file with instructions for doing this manually, figuring that would be easiest for you to review, rather than incorporating it in Chef/Vagrant files. It was also quicker for me to write it up this way. If you want me to add these steps to provisioning scripts, let me know.

I followed your manual install instructions but I’m getting a `imagejpeg(images/tempcaptcha/rlAoljA.png): failed to open stream: Permission denied` message when I try to access the contact webpage. I think I have a permissions problem for the temp captcha directory in the shared folder at `/vagrant`. I will have another go and @jessesiu will have a try with your instructions too.

However, I think the installation of the PHP libraries and the creation of the tempcaptcha directory in the virtual machine should be incorporated into the Chef/Vagrant provisioning system. We want the GigaDB website to be deployed automatically for development and production purposes, and not go back to having to perform manual steps on the command line.

When you add this functionality into Chef/Vagrant, please use the php and imagemagick Chef cookbooks to do this and not via bash scripts. The php Chef cookbook is already used in the gigadb-website project but you will need to include imagemagick into the list of [chef-cookbooks](https://github.com/pli888/chef-cookbooks). Documentation on how to do this is in [DEVELOPMENT.md](https://github.com/gigascience/gigadb-website/blob/develop/docs/DEVELOPMENT.md) in the gigadb-website repo.

> Using the CCaptcha system in the Yii Framework, i've created an on-the-fly Captcha system for the contact page that doesn't rely on calling pictures from a remote server (presumably the problem with Chinese access to existing system.)

If I can fix the tempcaptcha directory permissions problem, we will test your new Captcha system from a China location while we wait for the Chef/Vagrant work.

> There is need for a /images/tempcaptcha/ folder for storing these painted Captcha images. If the user completes the webform, then the .PNG is deleted. I am working on a cleanup system for the use-case where the user navigates away from the contact page without submitting the form. Should have that with you tomorrow.

I reviewed the [changes](https://github.com/gigascience/gigadb-website/pull/144/commits/8c7164dad2d17c23e38618dc53899b3ee5d0af6f) you made in protected/views/site/contact.php for the cleanup system. Somehow, it looks the whole contact.php file has been replaced but I don’t think this is necessary. After doing a file diff between your new contact.php and the contact.php in the develop branch, only 28 out of 86 lines of code have been changed/added in your version. I think the commit for this file should only consist of these changes and additions, and not the entire file. There is the same problem with the protected/views/user/_form.php file.

bobbledavidson:
if you're running the system using Vagrant on e.g. your local desktop, then you need to chmod the permissions of the /vagrant/tempcaptcha/ folder in your host file structure i.e. from your host terminal, not just from logging into the vagrant client terminal. Or at least, that was the solution when I also had this error.

bobbledavidson:
I'll work on adding the provisioning - my focus was to get the solution agreed first and then provisioning could be added later. 
As for the problem with the file changes, I'll rollback the branch and then update the files to ensure that Git recognises the changes appropriately.

bobbledavidson:
I've fixed the file change issue - the latest commit (branch: develop-tc-issue130) shows only the relevant line changes instead of falsely showing whole file changes.

pli888:
> if you're running the system using Vagrant on e.g. your local desktop, then you need to chmod the permissions of the /vagrant/tempcaptcha/ folder in your host file structure i.e. from your host terminal, not just from logging into the vagrant client terminal. Or at least, that was the solution when I also had this error.

Finally, got it working by creating the tempcaptcha directory with world-writeable permissions on the host before spinning up the VM.

pli888:
There is sometimes a problem with not all of the Captcha image being displayed on the contact web page which means the user cannot type in the correct response. For example, see [here](https://drive.google.com/file/d/1MtCne1r2q7Yyuh-sp8cC7lGkQBy9UvAg/view).

bobbledavidson:
Ah, ok - the 'painted' image is to narrow for a high number of capitals in the random string. I'll increase the width.

bobbledavidson:
I've updated the CaptchaGenerator function in the SiteController.php file so that it now paints an image that is 600px wide - this provides more than enough room for all combinations of 7 characters. There shouldn't be any more cutting short of Captcha strings. 
I've also removed the instructions for installing ImageMagick -it's not actually necessary, just the php-gd library is required. I'll add that to Chef shortly

bobbledavidson:
php-gd install has been added to Chef.

pli888:
Does the tempcaptcha folder in the images directory need to be in version control? Can it not be incorporated into the Vagrant-Chef build?

PR #146, opened 2017-12-14
Initial map browsing function

bobbledavidson:
Adding basic map browsing capabilities. In this commit, the 'browse samples' button is only available to logged in users (I misread the brief in the issue log).

There is one test data for GeoLocation  added via the gigadb_testdata.sql file.

The Open Layer map presents all country/region names in the language of the country.


bobbledavidson:
Added aesthetics as requested by Chris.

bobbledavidson:
Hi Peter - the outstanding tweak has been added: there are now scrollbars on the popups for each item on the mapbrowse page and the map autopans to ensure the popup is not lost behind the boundary of the map.

only1chunts:
@bobbledavidson , we've installed and tested the functionality, and can see the new scroll bar, however there now seems to be an issue with links for locations with multiple samples (not sure if it was present before or not).
When there is just 1 sample at a location the link is correct and it points to the DOI of the dataset that the sample is from. However when there are multiple samples in the list at a location all the sample links point to '<URL-prefix>/site/<DOI_number>' which is invalid.
Finally, please can the "browse samples" link be added to all pages that have the same row of buttons
screenshot_5

bobbledavidson:
OK - latest commit has the browse button included for admins, users and guests which should be all users.
I've fixed the links on the samples on the map browse page too.

Issue #167, opened 2018-02-13
Show full author names


pli888:
@only1chunts says when updating an author on the GigaDB Administration web pages, e.g. [here](http://localhost:9170/adminAuthor/update/id/3789), enable these pages to show the full author name including any middle name in the Display Name text field box.

[https://drive.google.com/open?id=16pBf-O1UsP78NRRRO6BiT1CtN0Ij5hOZ](https://drive.google.com/open?id=16pBf-O1UsP78NRRRO6BiT1CtN0Ij5hOZ)

only1chunts:
The display name field box should show the "display name" as previously defined by Rija's work, and it should be editable to overwrite that display name.

Issue #74, opened 2016-10-17
get complete citation from DOI ref


only1chunts:
In the dataset view we have a associated manuscripts section, currently this only displays the DOI as we enter it in the database, this should be able to look up that DOI and retrieve the text citation (i.e. title, authors etc...).
NB - this functionality will be used in #73 when completing the "citation" part of the dataset extension schema.
For ease it maybe worth considering adding those details to the database, so the lookup is only done once?
Both Crossref and datacite have APIs that could be useful for this?

jessesiu:
@only1chunts  This return citation is suitable?

curl -LH "Accept: text/x-bibliography; style=harvard3" https://doi.org/10.1093/gigascience/giy012

Zhang, Zhiyong et al., 2018. Draft Genome of the Protandrous Chinese Black Porgy, Acanthopagrus schlegelii. GigaScience. Available at: http://dx.doi.org/10.1093/gigascience/giy012.

only1chunts:
thats great, but doesn't Harvard style include the authors initials too? if you can find out how to include the initials that would be perfect, but as that is also fine. Thanks

only1chunts:
I have seen this working correctly in @jessesiu version

PR #168, opened 2018-02-14
Develop

jessesiu:


jessesiu:
The issues #167 #171 #74 already tested

PR #330, opened 2019-06-14
Change count numbers on the site index page

jessesiu:

The main page shows some count numbers from the database, but the data volume number is not linked with database

http://gigadb.gigasciencejournal.com:9170

屏幕快照 2019-06-14 下午4 52 01

The SiteController actionIndex function changed, add new sql statement to sum all published datasets size and convert to TB unit then show on the main page.

## Changes to the tests

Add unit test in protected/tests/functional/SiteTest.php, to assert count numbers are correct

### DIFF
--- a/protected/controllers/SiteController.php
+++ b/protected/controllers/SiteController.php
@@ -485,11 +485,8 @@ class SiteController extends Controller {
 
     public function formatBytes($bytes, $precision = 2) {
         $units = array('B', 'KB', 'MB', 'GB', 'TB');
-
         $base = log($bytes, 1024);
-        $suffixes = array('', 'K', 'M', 'G', 'T');
-
-        return round(pow(1024, $base - floor($base)), $precision) .' '. $suffixes[floor($base)];
+        return round(pow(1024, $base - floor($base)), $precision) ;
     }
 
     /**
--- a/protected/tests/functional/SiteTest.php
+++ b/protected/tests/functional/SiteTest.php
@@ -56,7 +56,7 @@ class SiteTest extends FunctionalTesting
     public function testItShouldShowCountNumber()
     {
         // this is the content we expect to be in
-        $expectations = ["4", "2", "16", "2 G"];
+        $expectations = ["4", "2", "16", "2"];
         $actual = [];
         $url = "http://gigadb.dev/site/" ;
         $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, null);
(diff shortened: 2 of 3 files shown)


protected/tests/functional/SiteTest.php
```diff
@@ -46,6 +46,26 @@ public function testItShouldShowLatestNews()
 
     }
```
pli888: I think it would be good to add a descriptive comment block for this function like the `testItShouldShowLatestNews()` function above.

protected/tests/functional/SiteTest.php
```diff
@@ -46,6 +46,26 @@ public function testItShouldShowLatestNews()
...
 
+    public function testItShouldShowCountNumber()
+    {
+
```
pli888: Delete blank line.

protected/tests/functional/SiteTest.php
```diff
@@ -46,6 +46,26 @@ public function testItShouldShowLatestNews()
...
+        $expectations = ["4", "2", "16", "2 G"];
+        $actual = [];
+        $url = "http://gigadb.dev/site/" ;
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, null);
+
+        //Find all count items on the page
+        $newsitemsElement = $this->findAllByCSS('html body div.content section div.container div.color-background div.home-color-background-block h4');
+
+        foreach ($newsitemsElement as $node) {
+
```
pli888: Delete blank line.

protected/controllers/SiteController.php
```diff
@@ -476,6 +483,15 @@ public function actionFeed(){
         exit;
     }
 
+    public function formatBytes($bytes, $precision = 2) {
+        $units = array('B', 'KB', 'MB', 'GB', 'TB');
+
+        $base = log($bytes, 1024);
+        $suffixes = array('', 'K', 'M', 'G', 'T');
```
pli888: @only1chunts When you look at how the volume of data is displayed in the home page, the units of data volume is displayed twice so currently on http://gigadb.org, it shows 43 T then underneath it also says `Data volume(TB)`. Perhaps it would be better just to display a number for the volume of data then underneath it could say `terabytes of data` so it is consistent with the numbers of datasets, samples and files.

only1chunts: The T after 43T is not required.
Words under the number 43 should be "Data volume (TB)"
The inclusion of commas every 3 decimal places would be good.

Issue #460, opened 2020-06-25
Functional test for admin page


pli888:
Recently, a bug allowed a non-logged in user to access the admin page at `/site/admin` which was caused by mis-configuration of the rules in `accessRules(`) function in the `SiteController` class. This bug has been fixed on the production server.

A functional test should be created to check this bug does not occur again in the future.

PR #463, opened 2020-06-30
Try to create a functional test for only logged in user can visit admin page

kencho51:
# Pull request for issue: #460 

## Changes to the tests
1. Create a functional_test folder.
2. Logged in user can visit `/site/admin` and see `Administration Page` in the content.
3. Not logged in user cannot visit `/site/admin` and can only see `Lost Password` in the `/site/login` content. 

##### Add function in `BrowserSignInPage.php`
1. To `assertFalse` that if an unlogged in user trying to login `/site/admin` page. 

##### Update the `unit_functional script`
1. Run unit_functional test on functional_test folder only.<br>
2. Add `--debug` parameter.


tests/unit_functional
```diff
@@ -12,20 +12,21 @@ set +a
```
pli888, line 12: I think the changes in this file have been included in this pull request by mistake?

tests/unit_functional
```diff
@@ -12,20 +12,21 @@ set +a
 
 printf "\nRunning PHPUnit functional tests\n"
 
-echo "Preparing the database for functional tests..."
-exec 3>&1
-exec 1> /dev/null
-pg_restore -h $GIGADB_HOST -U $GIGADB_USER -d $GIGADB_DB --clean --no-owner -v /var/www/sql/gigadb_testdata.pgdmp || true
-exec 1>&3
+#echo "Preparing the database for functional tests..."
```
pli888, line 15: I think the changes in this file have been included in this pull request by mistake?

protected/tests/support/BrowserSignInSteps.php
```diff
@@ -14,11 +14,22 @@ trait BrowserSignInSteps
 	public function loginToWebSiteWithSessionAndCredentialsThenAssert($email,$password,$to_assert)
 	{
 
-        $this->session->visit("http://gigadb.dev/site/login");
+        $this->session->visit("http://gigadb.dev/site/admin");
```
pli888, line 17: I'm not sure changes are needed in this file? For example, look at `AdminFileTest.php` for how to use `BrowserSignInSteps`.

protected/tests/functional_test/OnlyLoggedInTest.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
pli888, line 1: I would have added this file into the `protected/tests/functional` directory.

Maybe think about renaming this file to something more informative, maybe `AdminSiteTest.php` or `AdminSiteAccessTest.php`?

Also, look at the names of the functions in the other functional test files. The function names have a pattern: `testItShould...`. Its probably best to keep with this pattern for consistency.

protected/tests/functional_test/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,42 @@
...
+class AdminSiteAccessTest extends FunctionalTesting
+{
+    use BrowserSignInSteps;
+    use BrowserPageSteps;
+
+    public function testItShouldBeDisplayedToUsersWithAdminRole ()
+    {
+        // Logged in as Admin, and can visit to /site/admin page.
+        // 2 assertions, 2 passes.
+        $this->loginToWebSiteWithSessionAndCredentialsThenAssert("admin@gigadb.org","gigadb","Admin");
```
pli888, line 15: Instead of `Admin`, I think it is more specific to check for `Joe's GigaDB Page` after the user has logged in.

protected/tests/functional_test/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,42 @@
...
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Administration Page");
+    }
+
+    public function testItShouldNotBeDisplayedToUsersWithUserRole ()
+    {
+        // Logged in as User, but cannot visit to /site/admin page.
+        // 2 assertions, login pass, but visit to /site/admin fail.
+        $this->loginToWebSiteWithSessionAndCredentialsThenAssert("user@gigadb.org","vagrant","User");
```
pli888, line 24: I think it is more specific to check for `John's GigaDB Page` after the user has logged in. Also, confirm you have the correct password here too.

protected/tests/functional_test/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,42 @@
...
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Administration Page");
+
```
pli888, line 27: Checking for `Error 403` would be good to confirm that the user cannot access `/site/admin`.

protected/tests/functional_test/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,42 @@
...
+    public function testItShouldNotBeDisplayedToOrdinaryUsers ()
+    {
+        // No login was tested. 1 assertion,
+        // Visit to /site/admin, but would be redirected to /site/login, confirmed by seeing "Login in" as assert true,
+        // and the current url was /site/login.
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Login");
+        $out = $this->getCurrentUrl();
+        print($out);
```
pli888, line 38: Here, you could check that `Administration Page` should not be found on the page otherwise the user has access to the **Admin Page**.

protected/tests/functional/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Login");
+        //$current_site = $this->getCurrentUrl();
+        //print($current_site);
+        $this->session->visit($url);
+        $out = $this->session->getPage()->getContent();
+        //print($out);
+        //there is no "Administration Page' can be found in the $out content.
+        $this->assertFalse(strpos($out, "Adminstration Page"), "Ordinary User cannot visit admin page");
```
pli888, line 44: `assertFalse()` will report an error with the message if condition is true. Therefore test passes when `strpos` returns false when "Administration Page" is not found. The message should be `Ordinary User can visit admin page`.

pli888: You could also write a test to confirm that the user is re-directed to "/site/login" in this `testItShouldNotBeDisplayedToOrdinaryUsers()` function.

protected/tests/functional/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Functional test for only logged in admin user can access /site/admin.
+ */
+
+class AdminSiteAccessTest extends FunctionalTesting
+{
+    use BrowserSignInSteps;
+    use BrowserPageSteps;
+
+    public function testItShouldBeDisplayedToUsersWithAdminRole ()
```
pli888, line 11: You should remove the space between the function name and brackets here and in the other 2 functions.

protected/tests/functional/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,53 @@
...
+        // No login was tested.
+        // 2 assertions, visit to /site/admin, but would be redirected to /site/login, confirmed by seeing "Login in" as assertTrue is true.
+        // Fail to get "Administration Page" from the content of /site/admin, so assertFalse is false.
+
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Login");
+
+        // To confirm User has been re-directed to /site/login
+        $current_site = $this->getCurrentUrl();
+        print($current_site); // expect: "http://gigadb.dev/site/login"
```
pli888, line 41: You could use [`assertTrue()`](https://phpunit.readthedocs.io/en/9.2/assertions.html#asserttrue) to check the value of`$current_site` is "http://gigadb.dev/site/login".

pli888:
Hi @rija this pull request is ready for your review now.

rija:
Hi @kencho51

According to its function signature defined in
``vendor/phpunit/phpunit/src/Framework/Assert.php``

```
	/**
     * Asserts that a condition is true.
     *
     * @param bool   $condition
     * @param string $message
     *
     * @throws AssertionFailedError
     */
    public static function assertTrue($condition, $message = '')
    {
        static::assertThat($condition, static::isTrue(), $message);
    }
```

``assertTrue()`` takes a variable of PHP type ``bool`` as its first parameter and assert that its value is ``true`` and throws an exception otherwise.

Meanwhile, the function signature for ``hasContent()`` defined in
``vendor/behat/mink/src/Element/DocumentElement.php``

```
/**
     * Check whether document has specified content.
     *
     * @param string $content
     *
     * @return boolean
     */
    public function hasContent($content)
    {
        return $this->has('named', array('content', $content));
    }
 ```

 shows that ``hasContent()`` returns a value of PHP type ``bool`` ("boolean" is alias for "bool" in PHPDocs) to indicate whether it's ``true`` or ``false`` that the page has ``$content`` in it.

It means that to assert whether the document has the content "Login" it is enough to do 

```
 $this->assertTrue( $this->session->getPage()->hasContent("Login") ) ;
```

Which also reads well when you read it aloud in English.


The current code that it replaces is actually not correct, even though it doesn't crash and function in this particular context, it doesn't do what you think it does and would be cause of difficult to trace bugs in most contexts.

More specifically, this is the problematic snippet:
```
$out=="True"
```

* "True" is of the PHP type ``string``, it is not the PHP type that represents boolean values used for asserTrue() parameter and hasContent() return value
* "==" is the loose comparison operator, it's rarely the right one to use. Always use the strict "===" comparison operator unless you have a good reason not to. 

The reason it works here is because in loose comparisons, PHP converts any non-null string into ``true``
The test would have also passed if it said:

```
$out=="False"
```

you can read up more on comparison operators here:
https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison

and type declaration in functions here:
https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

**Comment by kencho51** [comment]
> Hi @kencho51
> 
> According to its function signature defined in
> `vendor/phpunit/phpunit/src/Framework/Assert.php`
> 
> ```
> 	/**
>      * Asserts that a condition is true.
>      *
>      * @param bool   $condition
>      * @param string $message
>      *
>      * @throws AssertionFailedError
>      */
>     public static function assertTrue($condition, $message = '')
>     {
>         static::assertThat($condition, static::isTrue(), $message);
>     }
> ```
> 
> `assertTrue()` takes a variable of PHP type `bool` as its first parameter and assert that its value is `true` and throws an exception otherwise.
> 
> Meanwhile, the function signature for `hasContent()` defined in
> `vendor/behat/mink/src/Element/DocumentElement.php`
> 
> ```
> /**
>      * Check whether document has specified content.
>      *
>      * @param string $content
>      *
>      * @return boolean
>      */
>     public function hasContent($content)
>     {
>         return $this->has('named', array('content', $content));
>     }
> ```
> 
> shows that `hasContent()` returns a value of PHP type `bool` ("boolean" is alias for "bool" in PHPDocs) to indicate whether it's `true` or `false` that the page has `$content` in it.
> 
> It means that to assert whether the document has the content "Login" it is enough to do
> 
> ```
>  $this->assertTrue( $this->session->getPage()->hasContent("Login") ) ;
> ```
> 
> Which also reads well when you read it aloud in English.
> 
> The current code that it replaces is actually not correct, even though it doesn't crash and function in this particular context, it doesn't do what you think it does and would be cause of difficult to trace bugs in most contexts.
> 
> More specifically, this is the problematic snippet:
> 
> ```
> $out=="True"
> ```
> 
> * "True" is of the PHP type `string`, it is not the PHP type that represents boolean values used for asserTrue() parameter and hasContent() return value
> * "==" is the loose comparison operator, it's rarely the right one to use. Always use the strict "===" comparison operator unless you have a good reason not to.
> 
> The reason it works here is because in loose comparisons, PHP converts any non-null string into `true`
> The test would have also passed if it said:
> 
> ```
> $out=="False"
> ```
> 
> you can read up more on comparison operators here:
> https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison
> 
> and type declaration in functions here:
> https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

Hi @rija , thank you for the details.
I have tried using  strict operator, `$out==="True"`, the test was passed. So would it be also one of the ways to fix the problem?

**Comment by rija** [comment]
No, it doesn't work, try for yourself
with the REPL ``php -a``:
```
php > var_dump(true == "True") ;
bool(true)
php > var_dump(true === "True") ;
bool(false)
```
And in AdminSiteAccessTest.php, when changing to "===":
```
$out = $this->session->getPage()->hasContent("Login"); // we assume that $out returns true if access control works
$this->assertTrue($out==="True", "Guest can see Administration Page");
```
will result in: 
```
1) AdminSiteAccessTest::testItShouldNotBeDisplayedToGuest
Guest can see Administration Page
Failed asserting that false is true.

/var/www/protected/tests/functional/AdminSiteAccessTest.php:59
```

**Comment by rija** [comment]
The following will work:
```
$out = $this->session->getPage()->hasContent("Login"); // we assume that $out returns true if access control works
$this->assertTrue($out === true, "Guest can see Administration Page");
```
but it still incorrect, because there is a duplication as ``assertTrue()``  is  doing the ``=== true`` part  already.

If you're wondering whether the test can be correctly expressed using explicit equality, yes, the two approaches below are correct and equivalent:

```
$this->assertTrue( $this->session->getPage()->hasContent("Login"), "Guest can see Administration Page");
```

```
$this->assertEquals( true,  $this->session->getPage()->hasContent("Login"), "Guest can see Administration Page");
```

**Comment by rija** [comment]
Also, there is a cost (memory and cpu) in instanciating variables, so we use variables when we intent to use/modify the value multiple times in the current context. Since ``$out`` is only used once, it's better to drop it and to "inline" the call to ``hasContent()`` instead.

**Comment by rija** [comment]
One more thing, in the list of changed files there is this file
``protected/tests/functional-test/AdminSiteAccessTest.php `` which seems to have been erroneously commited (notice the **"functional-test"** directory which shouldn't exist)

**Comment by kencho51** [comment]
Hi @rija 
Thank you for the detailed information, it is very useful.
The `protected/tests/functional-test/` is only for testing and saving time only, as a single run of `protected/tests/functional/` takes long time to complete.

**Comment by rija** [comment]
> Hi @rija
> Thank you for the detailed information, it is very useful.
> The `protected/tests/functional-test/` is only for testing and saving time only, as a single run of `protected/tests/functional/` takes long time to complete.

Glad it helps @kencho51,

you can run a single test using a command like this:

```
$ docker-compose run --rm test ./bin/phpunit --testsuite functional --bootstrap protected/tests/functional_custom_bootstrap.php --verbose --configuration protected/tests/phpunit.xml --no-coverage protected/tests/functional/AdminSiteAccessTest.php
```

File: protected/tests/functional/AdminSiteAccessTest.php [hunk 12]
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @uses \BrowserSignInSteps::loginToWebSiteWithSessionAndCredentialsThenAssert()
+     * @uses \BrowserPageSteps::visitPageWithSessionAndUrlThenAssertContentHasOrNull()
+     *
+     */
+    public function testItShouldBeDisplayedToUsersWithAdminRole()
+    {
+        $this->loginToWebSiteWithSessionAndCredentialsThenAssert("admin@gigadb.org", "gigadb", "Joe's GigaDB Page");
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Administration Page");
```
pli888, line 23: Based on Rija's comment re: cost of instantiating variables, the `$url` variable can be dropped and replaced with the `http://gigadb.dev/site/admin` link string.

protected/tests/functional/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @uses \BrowserSignInSteps::loginToWebSiteWithSessionAndCredentialsThenAssert()
+     * @uses \BrowserPageSteps::visitPageWithSessionAndUrlThenAssertContentHasOrNull()
+     *
+     */
+    public function testItShouldNotBeDisplayedToUsersWithUserRole()
+    {
+        $this->loginToWebSiteWithSessionAndCredentialsThenAssert("user@gigadb.org", "gigadb", "John's GigaDB Page");
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Error 403");
```
pli888, line 37: Based on Rija's comment re: cost of instantiating variables, the `$url` variable can be dropped and replaced with the `http://gigadb.dev/site/admin` link string.

protected/tests/functional/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     */
+    public function testItShouldNotBeDisplayedToGuest()
+    {
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Login");
+
+        // To confirm User has been re-directed to /site/login
+        $current_site = $this->getCurrentUrl();
+        $this->assertTrue($current_site == "http://gigadb.dev/site/login", "The current site has not been re-directed.");
```
pli888, line 54: Based on Rija's comment re: cost of instantiating variables, the `$current_site ` variable can be dropped and replaced with the call to `getCurrentUrl()`.

protected/tests/support/BrowserSignInSteps.php
```diff
@@ -21,4 +21,5 @@ public function loginToWebSiteWithSessionAndCredentialsThenAssert($email,$passwo
 
         $this->assertTrue($this->session->getPage()->hasContent($to_assert));
 	}
-}
\ No newline at end of file
+
```
pli888, line 24: Looks like the changes to this file are a couple of newlines. Maybe we don't need to merge these changes?

protected/tests/functional/AdminSiteAccessTest.php
kencho51: `$url` has been dropped.

protected/tests/support/BrowserSignInSteps.php
kencho51: Yes, please don't merge them.

protected/tests/functional/AdminSiteAccessTest.php
kencho51: `$url` has been dropped.

protected/tests/functional/AdminSiteAccessTest.php
kencho51: `$current_site` has been dropped.

protected/tests/functional/AdminSiteAccessTest.php
```diff
@@ -0,0 +1,59 @@
...
+     * @uses \BrowserPageSteps::getCurrentUrl()
+     *
+     */
+    public function testItShouldNotBeDisplayedToGuest()
+    {
+        $url = "http://gigadb.dev/site/admin";
+        $this->visitPageWithSessionAndUrlThenAssertContentHasOrNull($url, "Login");
+
+        // To confirm User has been re-directed to /site/login
+        $this->assertTrue($this->getCurrentUrl() == "http://gigadb.dev/site/login", "The current site has not been re-directed.");
```
pli888, line 51: Based on @rija's comment re: comparison operators, you need to use the strict === operator here to compare URLs?

Issue #462, opened 2020-06-30
change the default table settings in dataset pages


only1chunts:
**Is your feature request related to a problem? Please describe.**
Current the default view of the Files table on a dataset page (e.g. http://www.gigadb.org/dataset/100767) doesn't show the best selection of columns. 
The default view is of:
File Name | Sample ID | Data Type | File Format | Size | Release Date | Download
I would like to change the default view to be:
File Name | Description | Data Type | Size |  File Attributes | Download

It is possible for a user to define the columns displayed in the files table, but very few users are aware of that fact, so they don't change them.

**Describe the solution you'd like**
The default setting on the file table view needs to be changed to the above list.

**Describe alternatives you've considered**
making all columns visible by default is also an option, but that makes the table wider than the screen/browser window so users would need to scroll right to see the download button which is not good.

**Additional context**
This issue may need to be taken into consideration for the file browser widget (#63 ) too.

pli888:
>I would like to change the default view to be:
File Name | Description | Data Type | Size | File Attributes | Download

FYI @only1chunts not all all files have descriptions. Based on the output of `select count(*) from file where description = ''`, there are 20,289 files with no descriptions.

only1chunts:
Thanks Peter, descriptions weren't required when we first started so that
vast majority of those without description will be from datasets submitted
in the first year or so.
Cheers
Chris

PR #464, opened 2020-07-09
Update default columns in the table setting

kencho51:
# Pull request for issue: #462 

1. To check the column boxes by updating the `in_array` return value.

## Problem
Please help to have a look.
1. Items in the `$setting`:
image

2. `description` and `attribute` are in the `$setting`,  but in `in_array` returns `false`.

### DIFF
--- a/features/bootstrap/DatasetViewContext.php
+++ b/features/bootstrap/DatasetViewContext.php
@@ -282,20 +282,26 @@ class DatasetViewContext implements Context
                     $this->minkContext->getSession()->getPage()->hasContent($row['File name']), "File name match"
                 );
                 PHPUnit_Framework_Assert::assertTrue(
-                    $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "Sample ID match"
+                    $this->minkContext->getSession()->getPage()->hasContent($row['Description']), "Description match"
                 );
+//                PHPUnit_Framework_Assert::assertTrue(
+//                    $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "Sample ID match"
+//                );
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['Data Type']), "Data Type match"
                 );
-                PHPUnit_Framework_Assert::assertTrue(
-                    $this->minkContext->getSession()->getPage()->hasContent($row['File Format']), "File Format match"
-                );
+//                PHPUnit_Framework_Assert::assertTrue(
+//                    $this->minkContext->getSession()->getPage()->hasContent($row['File Format']), "File Format match"
+//                );
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['Size']), "Size match"
                 );
                 PHPUnit_Framework_Assert::assertTrue(
-                    $this->minkContext->getSession()->getPage()->hasContent($row['Release date']), "Release date match"
+                    $this->minkContext->getSession()->getPage()->hasContent($row['File Attributes']), "File Attributes match"
                 );
+//                PHPUnit_Framework_Assert::assertTrue(
+//                    $this->minkContext->getSession()->getPage()->hasContent($row['Release date']), "Release date match"
+//                );
                 if ($link) {
                     $this->minkContext->assertSession()->elementExists('css',"a.download-btn[href='$link']");
                 }
--- a/features/bootstrap/DatasetViewContext.php
+++ b/features/bootstrap/DatasetViewContext.php
@@ -275,11 +275,11 @@ class DatasetViewContext implements Context
             }
         }
         elseif("Files" == $arg1) {
-            //| File name                                        | Sample ID  | Data Type         | File Format | Size      | Release date | link |
+            //| File Name                                        | Sample ID  | Data Type         | File Format | Size      | Release date | link |
             foreach($table as $row) {
                 $link = $row['link'];
                 PHPUnit_Framework_Assert::assertTrue(
-                    $this->minkContext->getSession()->getPage()->hasContent($row['File name']), "File name match"
+                    $this->minkContext->getSession()->getPage()->hasContent($row['File Name']), "File Name match"
                 );
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['Description']), "Description match"
@@ -333,22 +333,22 @@ class DatasetViewContext implements Context
      */
     public function iShouldNotSeeTabWithTable($arg1, TableNode $table)
     {
-        //| File name                                        | Sample ID  | Data Type         | File Format | Size      | Release date | link |
+        //| File Name                                        | Sample ID  | Data Type         | File Format | Size      | Release date | link |
         foreach($table as $row) {
             if ("Files" == $arg1) {
                 PHPUnit_Framework_Assert::assertFalse(
-                    $this->minkContext->getSession()->getPage()->hasContent($row['File name']), "File name match"
+                    $this->minkContext->getSession()->getPage()->hasContent($row['File Name']), "File Name match"
                 );
             }
             elseif("Sample" == $arg1) {
                 if ($row['Sample ID']) {
                     PHPUnit_Framework_Assert::assertFalse(
-                        $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "File name match"
+                        $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "File Name match"
                     );
                 }
                 if ($row['Common Name']) {
                     PHPUnit_Framework_Assert::assertFalse(
-                        $this->minkContext->getSession()->getPage()->hasContent($row['Common Name']), "File name match"
+                        $this->minkContext->getSession()->getPage()->hasContent($row['Common Name']), "File Name match"
                     );
                 }
             }
--- a/features/bootstrap/DatasetViewContext.php
+++ b/features/bootstrap/DatasetViewContext.php
@@ -284,24 +284,15 @@ class DatasetViewContext implements Context
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['Description']), "Description match"
                 );
-//                PHPUnit_Framework_Assert::assertTrue(
-//                    $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "Sample ID match"
-//                );
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['Data Type']), "Data Type match"
                 );
-//                PHPUnit_Framework_Assert::assertTrue(
-//                    $this->minkContext->getSession()->getPage()->hasContent($row['File Format']), "File Format match"
-//                );
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['Size']), "Size match"
                 );
                 PHPUnit_Framework_Assert::assertTrue(
                     $this->minkContext->getSession()->getPage()->hasContent($row['File Attributes']), "File Attributes match"
                 );
-//                PHPUnit_Framework_Assert::assertTrue(
-//                    $this->minkContext->getSession()->getPage()->hasContent($row['Release date']), "Release date match"
-//                );
                 if ($link) {
                     $this->minkContext->assertSession()->elementExists('css',"a.download-btn[href='$link']");
                 }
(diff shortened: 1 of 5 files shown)


pli888:
Take a look at line 82 in the `DatasetController`:
```
$setting = array('name','size', 'type_id', 'format_id', 'location', 'date_stamp','sample_id'); // 'description','attribute' are hidden by default
```

The `$setting` array seems to control what columns are displayed in the `Files` table?

pli888:
@kencho51 I think I have worked out what is happening. Which columns are displayed in the Files table is defined by the `$setting` array in `DatasetController`. You have it set as:
```
$setting = array('name','size', 'type_id', 'format_id', 'location', 'date_stamp','sample_id', 'description', 'attribute'); // 'description','attribute' are hidden by default
```

The display order of the columns is configured in `protected/views/dataset/view.php`:
```
"columns": [
                { "visible": <?= in_array('name', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('description', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('sample_id', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('type_id', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('format_id', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('size', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('date_stamp', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('attribute', $setting)? 'true' : 'false' ?> },
                { "visible": <?= in_array('location', $setting)? 'true' : 'false' ?> },
              ]
```

The information in `$setting` can be changed by the user in `Table Settings` which is then saved in a cookie. To disregard this cookie, open http://gigadb.gigasciencejournal.com:9170/dataset/100004 in a private browser window then you will see a table showing:

File Name | Description | Sample ID | Data Type | File Format | Size | Release Date | File Attributes | Download
-- | -- | -- | -- | -- | -- | -- | -- | --
CS-master.tar.gz | compressed archive of the Analysis scripts (CS) files |   | Script | TAR | 114 B | 2014-06-06 |   |  
GD-master.tar.gz | compressed archive of the mock data and scripts GD files |   | Mixed archive | TAR | 163 B | 2014-06-06 |   |  

I think you need to edit `$setting` array in `DatasetController` to what @only1chunts requests in issue [462](https://github.com/gigascience/gigadb-website/issues/462).

pli888:
@kencho51 It's possible that Behat acceptance tests in `dataset-view.feature` will break because of this pull request. Check if this is the case and try to fix the test(s).

kencho51:
@pli888 I tried to run `behat` in my local gigadb server, and here is last part of the output:
```behat
--- Failed scenarios:
...
    features/dataset-view.feature:11
    features/dataset-view.feature:100
    features/dataset-view.feature:108
    features/dataset-view.feature:121
    features/dataset-view.feature:127
    features/dataset-view.feature:135
    features/dataset-view.feature:142
    features/dataset-view.feature:148
    features/dataset-view.feature:157
    features/dataset-view.feature:164
    features/dataset-view.feature:170
    features/dataset-view.feature:177
    features/dataset-view.feature:184
    features/dataset-view.feature:198
    features/dataset-view.feature:206
    features/dataset-view.feature:235
    features/dataset-view.feature:253
    features/dataset-view.feature:267
    features/dataset-view.feature:281
    features/dataset-view.feature:289
    features/dataset-view.feature:296
    features/dataset-view.feature:318
    features/dataset-view.feature:335
    features/dataset-view.feature:351
...

120 scenarios (104 failed, 16 skipped)
1306 steps (30 passed, 104 failed, 1172 skipped)
0m37.22s (18.28Mb)
``` 
And I think the dataset-view.feature:`184, 206, 235, 253, 267`  were failed because of this PR, I will further look at it.

rija:
Hi @kencho51,

@pli888 is right pointing the ``$setting`` variable in ``DatasetController.php`` to be the place of change for this work.

Do note that the content of the array is fairly static. Although you've just been asked to change it, 
it's a one-off event and the PHP program is never going to modify the content of the default list, so it's best (helps encapsulation, is more readable and uses less system resources)
to assign the default list to a constant in its own business object first. This has additional benefits:

* you can reuse the business object constant in the automated tests when you need to refer to the list's content without duplication of the array content and without risking introducing mismatch: you know that the list you test against is what you want the production code to use.
* If the file table setting needs other changes in the future, there is a logical class where to put those changes
* If there are more viewing contexts in the future (e.g: in reviewer mockup dataset pages we will want a different default list), we can just create additional constant with a different make up of the list of file columns. We can even create a constant for all columns to accompany tests that need to see all the columns content.


It may, for example,  look like this:

```
class DatasetPageSettings
{
	const VIEW_DEFAULT_FILE_COLUMNS = ['name','size', 'type_id', 'format_id', 'location', 'date_stamp'] ;
}
```

which could then be used like this:

```
$setting = DatasetPageSettings::VIEW_DEFAULT_FILE_COLUMNS;
```

notice the "\*\_FILE\_\*", this is because a dataset page has also got a sample table which works identically to the file table, and we will at some point do a similar refactoring to the ``$columns`` variable (for samples) in ``DatasetController.php``

features/dataset-view.feature
```diff
@@ -185,24 +185,33 @@ Feature: a user visit the dataset page
...
-		| Anas_platyrhynchos.pep 							| Pekin duck 	| Protein sequence 	| FASTA      	| 7.80 MiB  	| 2015-08-03  | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/Anas_platyrhynchos.pep |
-		| Anas_platyrhynchos_domestica.RepeatMasker.out.gz 	| Pekin duck 	| Other 			| UNKNOWN    	| 7.79 MiB  	| 2015-03-23  | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/Anas_platyrhynchos_domestica.RepeatMasker.out.gz |
-		| duck.scafSeq.gapFilled.noMito 					| Pekin duck 	| Sequence assembly	| FASTA 		| 1.03 GiB 		| 2013-01-23  | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/duck.scafSeq.gapFilled.noMito |
-		| pre_03AUG2015_update 								|				| Directory 		| UNKNOWN 		| 50.00 MiB 	| 2015-08-03  | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/pre_03AUG2015_update |
-		| readme.txt 										|				| Readme 			| TEXT 			| 337 B 		| 2013-01-23  | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/readme.txt |
+		| File Name              							| Description  	                                                                    | Data Type       	| Size  		| File Attributes | link |
```
pli888, line 188: Try changing `File Name` to `File name` and re-run the this test to see if it passes.

kencho51: @pli888 The behat tests for all scenario have been passed after changing to `File name`. At first, I changed it to `File Name` because the `Name` appears in the files table of local gigadb web page.

kencho51:
> Hi @kencho51,
> 
> @pli888 is right pointing the `$setting` variable in `DatasetController.php` to be the place of change for this work.
> 
> Do note that the content of the array is fairly static. Although you've just been asked to change it,
> it's a one-off event and the PHP program is never going to modify the content of the default list, so it's best (helps encapsulation, is more readable and uses less system resources)
> to assign the default list to a constant in its own business object first. This has additional benefits:
> 
> * you can reuse the business object constant in the automated tests when you need to refer to the list's content without duplication of the array content and without risking introducing mismatch: you know that the list you test against is what you want the production code to use.
> * If the file table setting needs other changes in the future, there is a logical class where to put those changes
> * If there are more viewing contexts in the future (e.g: in reviewer mockup dataset pages we will want a different default list), we can just create additional constant with a different make up of the list of file columns. We can even create a constant for all columns to accompany tests that need to see all the columns content.
> 
> It may, for example, look like this:
> 
> ```
> class DatasetPageSettings
> {
> 	const VIEW_DEFAULT_FILE_COLUMNS = ['name','size', 'type_id', 'format_id', 'location', 'date_stamp'] ;
> }
> ```
> 
> which could then be used like this:
> 
> ```
> $setting = DatasetPageSettings::VIEW_DEFAULT_FILE_COLUMNS;
> ```
> 
> notice the "*_FILE_*", this is because a dataset page has also got a sample table which works identically to the file table, and we will at some point do a similar refactoring to the `$columns` variable (for samples) in `DatasetController.php`

Hi @rija , will try to implement this feature. Thank you for the suggestion.

features/dataset-view.feature
rija: Hi @kencho51 . 
Your thinking is correct. The scenario should reflect the business view or the reaility of the feature as seen by the user, and that in turn should drive the implementation. The error was just the mismatch between the implementation and the scenario. In my email response, I did say that the feature file had the wrong casing because of what the original case was before your change, but really, we should be using what's on the user interface (which is File Name).

rija:
@kencho51 

I forgot to mention, and it might not be obvious, but that  ``DatasetPageSettings`` class is to be placed in a file at 
``protected/components/DatasetPageSettings.php`` (not with models as it's not an ActiveRecord model).

features/bootstrap/DatasetViewContext.php
```diff
@@ -275,26 +275,23 @@ public function iShouldSeeTabWithTable($arg1, TableNode $table)
             }
         }
         elseif("Files" == $arg1) {
-            //| File name                                        | Sample ID  | Data Type         | File Format | Size      | Release date | link |
+            //| File Name                                        | Sample ID  | Data Type         | File Format | Size      | Release date | link |
```
pli888, line 278: Might as well update this comment line with how the File table looks now.

features/bootstrap/DatasetViewContext.php
```diff
@@ -332,26 +329,26 @@ public function iShouldSeeTabWithTable($arg1, TableNode $table)
     }
 
     /**
-     * @Then I sould not see :arg1 tab with table
+     * @Then I should not see :arg1 tab with table
      */
-    public function iSouldNotSeeTabWithTable($arg1, TableNode $table)
+    public function iShouldNotSeeTabWithTable($arg1, TableNode $table)
```
pli888, line 334: You should check if any Behat tests were using the `iSouldNotSeeTabWithTable` function before you corrected its name. I think there are 2 tests in `dataset-view.feature` that used the `iSouldNotSeeTabWithTable` function. These tests need to be fixed if they are broken from your function name change.

kencho51: I think before I corrected it, `Scenario: Call To Actions - not logged in` has correct word `should`, and  only `Scenario: Files - Items per page` contains `sould`, so I changed it. 
And I have ran Behat test for `Scenario: Call To Actions - not logged in`, it passes too.

kencho51: But the rest of the `iSouldNotSeeTabWithTable` was used in the Samples table scenarios. I have updated them and Behat tests were passed.

features/dataset-view.feature
```diff
@@ -213,43 +213,43 @@ Feature: a user visit the dataset page
...
 		And I should see "Download Link"
 		And the "location" checkbox is checked
 		And I should see "File Attributes"
-		And the "attribute" checkbox is unchecked
+		And the "attribute" checkbox is checked
 		And I should see a button "Save changes" with no link
 		And I should see a button "Close" with no link
 
-	@ok @pageSize @files @javascript
+	@wip @pageSize @files @javascript @pr464
```
pli888, line 234: I think the `@wip` tag should be removed here. When we use `@wip` in our future work with Behat tests then this test will also run which we won't want it to.

features/bootstrap/DatasetViewContext.php
```diff
@@ -332,26 +329,26 @@ public function iShouldSeeTabWithTable($arg1, TableNode $table)
...
                 PHPUnit_Framework_Assert::assertFalse(
-                    $this->minkContext->getSession()->getPage()->hasContent($row['File name']), "File name match"
+                    $this->minkContext->getSession()->getPage()->hasContent($row['File Name']), "File Name match"
                 );
             }
             elseif("Sample" == $arg1) {
                 if ($row['Sample ID']) {
                     PHPUnit_Framework_Assert::assertFalse(
-                        $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "File name match"
+                        $this->minkContext->getSession()->getPage()->hasContent($row['Sample ID']), "File Name match"
```
pli888, line 346: The `"File Name match"` String message seems wrong here when we compare this and above. The String message should probably be `"Sample ID match"`.

features/bootstrap/DatasetViewContext.php
```diff
@@ -332,26 +329,26 @@ public function iShouldSeeTabWithTable($arg1, TableNode $table)
...
                     );
                 }
                 if ($row['Common Name']) {
                     PHPUnit_Framework_Assert::assertFalse(
-                        $this->minkContext->getSession()->getPage()->hasContent($row['Common Name']), "File name match"
+                        $this->minkContext->getSession()->getPage()->hasContent($row['Common Name']), "File Name match"
```
pli888, line 351: The `"File Name match"` String message seems wrong here when we compare this and above. The String message should probably be `"Common Name match"`.

Issue #185, opened 2018-05-04
add column description to sample table


only1chunts:
A request from a user:

> "Regarding the [sample] table contained in the [dataset] page we would like to have further detailed information about each column since we cannot interpret the correct meaning of some columns (common name, scientific name, ect.), can you provide us a description for such columns?"

This seems fair, we should add a (i) information icon in the headers that users can hover over to get a popup of the column descriptions.

kencho51:
Hi @pli888 whats the `information icon` look like? Any example that I can look for it?

only1chunts:
Good point @kencho51 ! just looking around our website the only place we actually use the (i) icon is in the wizard pages, elsewhere we've just added hover over pop-ups to explain some things. So perhaps for consistency on the website we should just use the hover-over pop-up text. Then at some point update the submission wizard to match, what do you think? The alternative is to add (i) icons to various other places!
Do you have the descriptions of the columns yet, they are not in this ticket, so I'll add them: 
```
File Name - The name of the file. Click header to sort by A-Z/Z-A. 
Description - Short description of file contents. Click header to sort by A-Z/Z-A. 
Sample ID - Name or ID of sample used to generate this file.
Data Type - The type of data in the file, see [help](http://gigadb.org/site/help#vocabulary) page for definitions of individual data types.  Click header to sort by A-Z/Z-A. 
File Format - The format of the file, see [help](http://gigadb.org/site/help#vocabulary) page for definitions of individual file formats. Click header to sort by A-Z/Z-A. 
Size - The size on disk of the file. Click header to sort by A-Z/Z-A. 
Release Date - Date of release of the file, see the history log for details of any changes made after initial release date. Click header to sort by A-Z/Z-A. 
File Attributes - Additional information about the file presented as Key:Value pairs. 
Download link - The direct link to the files server location.
```
Note- the link to the data-type and file-format vocaularies page may need to be changed when we update the help pages since the current public list of terms is out of date and incomplete!

kencho51:
Hi @only1chunts Which table do you want to put the hover over pop-up text? The description you gave fits the `Files table`, not the `Sample table`. If you want pop-up text for both tables, please provide the description for `Sample table` as well.

only1chunts:
Sorry! yes please on both tables, and the funding table:
**Sample Table:**
```
Sample ID - User-specified name or identifier of the sample object. Note: a DNA sample and an RNA sample from the same donor are classed as two separate samples.
[...]

PR #473, opened 2020-07-30
Allow mouseover pop up text

kencho51:
# Pull request for issue: #185 
1. Change <del>`th`</del> to `td` that allows mouse over pop-up text.
2. The following picture shows the outcome:
popuptext

## Questions:
1. This issue is at first requested to fix `Sample table`, but the description given in the #185 conversation fits the `Files table`. So the sample table description will be available and add this feature to both tables? 
2. Is it Ok to change the table head `th` to `td`?

### DIFF
--- a/protected/views/dataset/view.php
+++ b/protected/views/dataset/view.php
@@ -319,15 +319,24 @@ $this->pageTitle="GigaDB Dataset - DOI 10.5524/".$model->identifier." - ".$title
                             <table id="files_table" class="table table-striped table-bordered" style="width:100%">
                                 <thead>
                                     <tr>
-                                        <th>File Name</th>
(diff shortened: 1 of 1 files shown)


pli888:
> Is it Ok to change the table head th to td?

@kencho51  Why do you need to change the `th` tag to the `td` tag?

And how do you display the sort icons on each of the column headers?

Continue this work on the Sample table then I can deploy this functionality onto a staging server for the curators to look at.

kencho51:
> @kencho51 Why do you need to change the `th` tag to the `td` tag?
> And how do you display the sort icons on each of the column headers?

`td` is for table data cell, and it allows the display the sort icons. But it is not requested, I will revert it back to `th` in the future commit.

Issue #437, opened 2020-05-29
Viewing/finding files from very large datasets


ScottBGI:
Feedback from Gane Wong on usability of the 1KP dataset, and how we need to change the way we view files for very large datasets. This example has >7000 files and currently its not very usable to trying to pull out specific files you want. And going forward we are likely to have more and more projects with very large numbers of files.

See the GigaDB entry for the 1KP data:

http://dx.doi.org/10.5524/100627

And for comparison see the corresponding Cyverse dataset:

https://doi.org/10.25739/8m7t-4e85.

Feedback from Gane:
Right now the list is a mess as there 7289 files and even if we show 200/page it is not possible to step to an arbitrary page. You have to step through sequentially. At least on cyverse you can guess which page you need and jump directly to it. Compared to the Cyverse site this also presents a list of a thousand plus files but there are broken into 15 pages and you could jump to any one you want.

The reason GigaDB has over 7000 was they did not group by species. Again that information was provided, right in the file names. Obvious just by looking. In any case the problem is even if you show 200 files/page there are still about 40 pages and the buttons at the bottom only go from 1 to 10 or something like that. You cannot jump ahead to page 35. You have to do it one-by-one. 

Here’s an easy solution. At the bottom of the webpage there are tabs labelled 1 to 10 and a “next” button. The problem is “next” only advances by one webpage, and each “next” takes a few minutes to download. If there are 80 webpages you can imagine the problem. I suggest we simply get rid of the “next” button and generate a tab for EVERY possible page. Default to 100 files/page so the total number of tabs is more reasonable. But don’t make people manually push “next” forever.

Even better idea. Since the list is alphabetic, instead of having to guess which page to go to, let the user enter the first few letters of the file and go to the first page with such files.

only1chunts:
There is a long-standing ticket for improving the file browser (#63 ), I've added this ticket on user feedback there.

only1chunts:
Another example of a dataset with too many files to view properly is www.doi.org/10.5524/100310 , with 19981 files! The webpage actually fails to load most of the time! We may need to bundle and tar.gz some of those.

ScottBGI:
Gane has been chasing up and asking if we could do one of his workaround options first as they'd quite like to issue a correction to the paper but don't want to highlight the data till it's more usable. Any timeline I can give him for a temp or full fix?

pli888:
Another option is to have something like this at the bottom of the dataset webpage:

[...]

PR #495, opened 2020-08-04
Add pagination for large dataset 

kencho51:
# Pull request for issue: #437 

### This is a pull request for the following functionalities:
1. Show `First` and `Last` buttons.
2. Try to implement `Jump to page` functionality.

### Problems
1. Click `First`, `Previous` and `1` will redirect to, for example, `/dataset/view/id/100002`, but not `/dataset/view/id/100002/Files_page/1`.
2. The extension `CLinkPager` should be able to create paginator like this `<< First | < Previus | 1 | ... | 5 | Next > | Last >>`, but it is not showing in any dataset page.


pli888:
@kencho51 In the input text box field, is it possible to go to the relevant URL when the return button is pressed after a user has entered a page number?

kencho51:
>  In the input text box field, is it possible to go to the relevant URL when the return button is pressed after a user has entered a page number?

@pli888 What do you mean go to the relevant URL? When `Go to page` button is pressed, users will be re-directed to the page according to the number they have input.

pli888:
@kencho51 I am thinking of how to improve the `Go to page` functionality. If a user enters a page number into the text box then it would be reasonable to be able to go the relevant page URL when they press the return key?

I can deploy your work onto the staging server and let the curators comment on the `Go to page` functionality.

kencho51:
> @kencho51 I am thinking of how to improve the `Go to page` functionality. If a user enters a page number into the text box then it would be reasonable to be able to go the relevant page URL when they press the return key?

@pli888 I was also thinking about it, like showing the total pages by default, for example: 
`Go to page` [ ] of `total page no.`.

kencho51:
> > I am thinking of how to improve the `Go to page` functionality. If a user enters a page number into the text box then it would be reasonable to be able to go the relevant page URL when they press the return key?
> 
> @pli888 I was also thinking about it, like showing the total pages by default, for example:
> `Go to page` [ ] of `total page no.`.

Like this:
image

pli888:
@kencho51 I did some more testing of your pagination work. Can you perform the steps below to see if you get the same problem?

1. Go to http://gigadb.gigasciencejournal.com:9170/dataset/100006
2. Click on Files tab
3. Enter 2 in page number text box and press "Go to page" button
4. Look at the URL in the browser, you will see http://gigadb.gigasciencejournal.com:9170/dataset/100006///Files_page/2 
5. Press the "1" button to go back to page 1
6. The browser URL is http://gigadb.gigasciencejournal.com:9170/dataset/view/slug/%2F%2FFiles_page%2F2/id/100006/Files_page/1 and I get a Error 400 your request is invalid message in the browser

kencho51:
>  I did some more testing of your pagination work. Can you perform the steps below to see if you get the same problem?
> 
> 1. Go to http://gigadb.gigasciencejournal.com:9170/dataset/100006
> 2. Click on Files tab
> 3. Enter 2 in page number text box and press "Go to page" button
> 4. Look at the URL in the browser, you will see http://gigadb.gigasciencejournal.com:9170/dataset/100006///Files_page/2
> 5. Press the "1" button to go back to page 1
> 6. The browser URL is http://gigadb.gigasciencejournal.com:9170/dataset/view/slug/%2F%2FFiles_page%2F2/id/100006/Files_page/1 and I get a Error 400 your request is invalid message in the browser

@pli888 I found this error only reproducible for case 1.

Case1:
When we visited the File tab and input "2" in the `Go to page` box, the URL in the browser would become:  `http://gigadb.gigasciencejournal.com:9170/dataset/101001///Files_page/2`.

Case 2:
But if we visited the File tab and press the "1" or "2" buttons first, then we input "2" in the `Go to page` box, the URL would be: `http://gigadb.gigasciencejournal.com:9170/dataset/view/id/101001/Files_page/2`.

The problem because `window.location.pathname` only output the current browser path, in the case 1, it will output `"/dataset/100006"`. But in case 2, it will output `"/dataset/view/id/100006/Files_page/2"`

I solve this by making an array with these default values: `"", "dataset", "view", "id", pageID`, and the navigation of the pages works well in my branch, details please refer to my pushed code.

I will make a Behat test on this feature.

pli888:
@kencho51 One way to simulate typing the Enter/Return key in the `pageTarget` text field in an acceptance test:
```
    /**
     * @Given I manually press return
     */
    public function iManuallyPressReturn()
    {
        $xpath = '//*[@id="pageTarget"]';
        $this->minkContext->getSession()->getDriver()->getWebDriverSession()->element('xpath', $xpath)->postValue(['value' => ["\r\n"]]);
    }
```

kencho51:
> ```
>     /**
>      * @Given I manually press return
>      */
>     public function iManuallyPressReturn()
>     {
>         $xpath = '//*[@id="pageTarget"]';
>         $this->minkContext->getSession()->getDriver()->getWebDriverSession()->element('xpath', $xpath)->postValue(['value' => ["\r\n"]]);
>     }
> ```

@pli888 I followed your code, the Behat test was passed.

features/dataset-view.feature
```diff
@@ -277,6 +277,60 @@ Feature: a user visit the dataset page
 		| pre_03AUG2015_update 								| folder containing originally submitted data files, prior to update Aug 3rd 2015.	| Directory 		| 50.00 MiB 	|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/pre_03AUG2015_update |
 		| readme.txt 										|				                                                                    | Readme 			| 337 B 		|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/readme.txt |
 
+	@ok @files @javascript @pr437
+	Scenario: Files - Pagination
+		Given I am not logged in to Gigadb web site
+		And I am on "/dataset/101001"
+		And I follow "Files"
```
pli888, line 284: On my dev environment, using `chrome headless` breaks Behat tests at this step.

features/bootstrap/DatasetViewContext.php
```diff
@@ -217,6 +217,15 @@ public function iShouldSeeAButton($arg1)
...
     }
 
+    /**
+     * @Then I manually hit return
+     */
+    public function iManuallyHitReturn()
+    {
+        $xpath = '//*[@id="pageNumber"]';
+        $this->minkContext->getSession()->getDriver()->getWebDriverSession()->element('xpath', $xpath)->postValue(['value' => ["\r\n"]]);
+    }
```
pli888, line 227: At the moment, this function seems quite restrictive as it can only be used to simulate a Return key press in a text field with id = "pageNumber". The function could be made more generic to simulate any key press provided as a parameter on any field which can be located with an XPath string.

Maybe the php-doc annotation could be something like this: `@And I send the :arg1 field with keypress :arg2` and the function name `iSendTheFieldWithKeyPress($arg1, $arg2)`.

features/dataset-view.feature
pli888, line 284: Behat tests with the @javascript tag breaks at this step for me. I think this is because the chrome container needs to be built and started by adding it to the `depends_on` section for the `webapp` service in `docker-compose.yml`.

ops/deployment/docker-compose.yml
```diff
@@ -148,13 +148,27 @@ services:
     extra_hosts:
       - "gigadb.dev:172.16.238.10"
 
+  chrome:
```
pli888, line 151: Need to add chrome to the depends_on section in webapp service.

kencho51: `chrome` is added to `webapp` service, and the Behat tests pass after the change. Should I remove `PhantomJS` from the `webapp`? And  running a Behat test, for example, `1 scenario 9 steps` takes `4m14.62s`,  it is too time consuming?

pli888: I think we should remove PhantomJS since it is not being maintained anymore and it has been replaced with Chrome in this PR for running Behat tests involving Javascript. Perhaps @rija can confirm we can do this?

protected/components/FilesPagination.php
```diff
@@ -0,0 +1,17 @@
...
+Yii::import("system.web.CPagination");
+
+class FilesPagination extends CPagination {
+    public function createPageUrl($controller,$page)
+    {
+        $params=$this->params===null ? $_GET : $this->params;
+//        if($page>0) // page 0 is the default
+            $params[$this->pageVar]=$page+1;
+//        else
+//            unset($params[$this->pageVar]);
```
pli888, line 12: If the code is commented out then it should not be present?

protected/components/FormattedDatasetFiles.php
```diff
@@ -91,7 +91,8 @@ public function getDataProvider(): CArrayDataProvider
 		    'pagination' => null
 		    )
 		);
-		$files_pagination = new CPagination(count($files));
+//		$files_pagination = new CPagination(count($files));
```
pli888, line 94: If the code is commented out then it should not be present?

protected/views/dataset/view.php
```diff
@@ -746,3 +748,40 @@ class="btn btn-green btn-block claim-button"
...
+        if (userInput >= min && userInput <= max) {
+            console.log("Valid page number!");
+        }else if (userInput > max) {
+            targetPageNumber = max;
+            console.log("Error, return to " + max);
+        } else if (userInput < min) {
+            targetPageNumber = min;
+            console.log("Error, return to " + min);
+        }
+        // var targetUrlArray = Array.apply(null, Array(5)).map(function(_,i) { return window.location.pathname.split("/")[i]});]
```
pli888, line 770: If the code is commented out then it should not be present?

protected/views/dataset/view.php
```diff
@@ -746,3 +748,40 @@ class="btn btn-green btn-block claim-button"
...
+        // Create array with default values
+        //TODO: This array is hardcoded
```
pli888, line 772: Is there more work to do here?

protected/views/dataset/view.php
```diff
@@ -746,3 +748,40 @@ class="btn btn-green btn-block claim-button"
...
+        let targetUrlArray = ["", "dataset", "view", "id", pageID];
+        targetUrlArray.push('Files_page', targetPageNumber);
+        window.location = window.location.origin + targetUrlArray.join("/");
+        // Uncomment will show the target url in console.
+        // console.log(window.location.origin + targetUrlArray.join("/"))
+    }
+
+    function pressEnter() {
```
pli888, line 780: Is this JS function required?

protected/views/dataset/view.php
kencho51: The code is removed.

protected/views/dataset/view.php
kencho51: The JS function is removed.

protected/views/dataset/view.php
kencho51: The hardcoded array serves the purpose and works well, no more work is needed  at this stage.

protected/components/FormattedDatasetFiles.php
kencho51: The code is removed.

protected/components/FilesPagination.php
kencho51: The code is removed.

features/bootstrap/DatasetViewContext.php
rija: I would keep the step syntax and the name of the implementing function specific to scenario as it is for now (although I'd drop the "manually") but factor out the logic into a more generic ``sendKeyPressFromField()`` helper function. 

```
/**
 * @When I hit return
 */
function IHitReturn() {
	$this->sendKeyPressFromField("\r\n","pageNumber")
}
```

the scenario reads better that way from user/business perspective, yet the main logic is generic:
```
When I fill in "pageNumber" with "2"
And I hit return
```

Later, on future work, if a similar need for pressing a key on a different field arises, we could always revisit this and come up with better abstraction that fit that future context and this one. But until then, it's still hypothetical (there are many ways to abstract and generize, doing it too early may get us the wrong one)

kencho51: @rija I followed your suggestion, and ran the behat test, but it failed and thrown the following error:
```
 And I hit return                                           # DatasetViewContext::iHitReturn()
      Fatal error: Call to undefined method DatasetViewContext::sendKeyPressFromField() (Behat\Testwork\Call\Exception\FatalThrowableError)
    │
    │  http://gigadb.dev/dataset/101001
    │
    └─ @AfterStep # GigadbWebsiteContext::debugStep()
```

pli888: @kencho51 It looks like you need to create the `sendKeyPressFromField($keypress, $field_name)` function in your DatasetViewContext class.

kencho51: @pli888 `sendKeyPressFromField($key, $field_name)` has been created and passed the Behat test.

rija:
Hi @kencho51,

When I run all behat tests on your branch, I've got some tests failing

```
@admins-attach-author-to-user @issue-56 @ok-docker
Feature: a curator can fill in user id in an author record
  As a curator,
  I want to connect a user identity to an author record
  So that I can enable gigadb users direct access to the dataset they have authored

  Background:                                                                 # features/admins-attach-author-to-user.feature:7
    Given Gigadb web site is loaded with "gigadb_testdata.pgdmp" data         # GigadbWebsiteContext::gigadbWebSiteIsLoadedWithData()
      │ Initializing the database with gigadb_testdata.pgdmp... Terminating DB Backend... Restarting php container for  project...
      │ 
    And default admin user exists                                             # GigadbWebsiteContext::defaultAdminUserExists()
    When I go to "/dataset/100002"                                            # Behat\MinkExtension\Context\MinkContext::visit()
    Then I should see "Genomic data from Adelie penguin (Pygoscelis adeliae)" # Behat\MinkExtension\Context\MinkContext::assertPageContainsText()
      The text "Genomic data from Adelie penguin (Pygoscelis adeliae)" was not found anywhere in the text of the current page.
      
      +--[ http://gigadb.dev/dataset/100002 | Selenium2Driver ]
```
It seems that Chrome (at least on my setup) launches itself by default with a browsing window of small size of 800x600 [1] that causes the dataset page to not show the title [2] which is what that failing step is trying to find.

If you are not seeing this problem, you may be able to reproduce it by forcing Chrome to starts with the lower sized window by updating ``behat.yml`` configuration:

```
              chrome:
                switches:
                  - "--headless"
                  - "--disable-gpu"
                  - "--window-size=800,600"
```

The fix is to configure the settings above with a higher size: ``"--window-size=1024,768"``.
Even if your setup is not showing the problem, it's a good habit to make these settings deterministic so they are garanteed to be the same for everyone.

>[1] you can find out by preceding the step with a ``I take screenshot "x"`` and check the image property of the screenshot file
>[2] that's a responsive design/accessbility bug in itself but it's not related to your work or this ticket. It's going to be fixed on my branch.

kencho51:
Hi @rija, I have updated the `behat.yml` as below:
```
              chrome:
                switches:
                  - "--headless"
                  - "--disable-gpu"
                  - "--window-size=1024,768"
``` 
And I have added `And I take a screenshot named`  in this `@admins-attach-author-to-user @issue-56 @ok-docker`, the behat test passed.

rija:
> Hi @rija, I have updated the `behat.yml` as below:
> 
> ```
>               chrome:
>                 switches:
>                   - "--headless"
>                   - "--disable-gpu"
>                   - "--window-size=1024,768"
> ```
> 
> And I have added `And I take a screenshot named` in this `@admins-attach-author-to-user @issue-56 @ok-docker` scenario, the behat test passed.

Hi @kencho51,

thanks for following up on my comment.

Couple of observations:


```
And I take a screenshot named "testwinsize_2"
```

is useful for debug, but it should be removed once it has served its purpose (here verifying behat config works),
as it has nothing to do with that feature's scenario


Also that debug step only works with a Javascript browser (chrome) so when using it
don't forget to temporarily add a "javascript" tag above the scenario's header, otherwise you will get this warning and no screenshot:
```
 And I take a screenshot named "testwinsize_2"                                             # GigadbWebsiteContext::itakeAScreenshot()
      │ cannot take screenshot with this driver/var/www/features/bootstrap/GigadbWebsiteContext.php:232:
      │ class Behat\Mink\Driver\GoutteDriver#3798 (7) {
```

as the default browser (Goutte) doesn't support the implementation for that debugging step. 
(it should be feasible to make it work with both drivers, but never had the need to do that so far, and it's
easy to add the "javascript" tag when it's needed)

Other than that, all tests pass. Good Job!

features/bootstrap/DatasetViewContext.php
```diff
@@ -217,6 +217,34 @@ public function iShouldSeeAButton($arg1)
         PHPUnit_Framework_Assert::assertTrue( $this->minkContext->getSession()->getPage()->hasLink($arg1) );
     }
 
+    /**
+     * @Then I manually hit return
+     */
+    public function iManuallyHitReturn()
```
pli888, line 223: This function can be deleted now because it is replaced by the `iHitReturn()` function?

features/dataset-view.feature
```diff
@@ -277,6 +277,61 @@ Feature: a user visit the dataset page
...
+		| duck.scafSeq.gapFilled.noMito 					| draft genome assembly                                                             | Sequence assembly	| 1.03 GiB 		|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/duck.scafSeq.gapFilled.noMito |
+
+	@wip @files @javascript @pr437
+	Scenario: Files - Pagination
+		Given I am not logged in to Gigadb web site
+		And I am on "/dataset/101001"
+		And I follow "Files"
+		And I have set pageSize to "5" on "files_table_settings"
+		When I fill in "pageNumber" with "2"
+#		And I manually hit return
```
pli888, line 327: This line can be removed since it is commented out and it's iManuallyHitReturn() function will be deleted.

features/bootstrap/DatasetViewContext.php
kencho51: `iHitReturn` is deleted now.

features/dataset-view.feature
kencho51: `And I manually hit return` is removed.

protected/views/dataset/view.php
pli888: The `pressEnter` JS function is still present.

kencho51: The `pressEnter` is required as it returns `goToPage()`, then user can to go a specific page of a `file table` by inputing the page number and hitting the physical `Return` key.

pli888: Ok now I understand that this function is responsible for detecting an Enter key press in the `pageNumber` text field and will then call the `goToPage()` JS function.

pli888: Maybe calling this function `detectEnterKeyPress` is more accurate?

kencho51: This function name has been updated.

features/dataset-view.feature
```diff
@@ -277,6 +277,60 @@ Feature: a user visit the dataset page
...
+		| Anas_platyrhynchos.gff 							| genome annotations	                                                            | Annotation 		| 10.10 MiB 	|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/Anas_platyrhynchos.gff |
+		| Anas_platyrhynchos.pep 							| amino acid translations of coding sequences                                       | Protein sequence 	| 7.80 MiB  	|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/Anas_platyrhynchos.pep |
+		| Anas_platyrhynchos_domestica.RepeatMasker.out.gz 	| repeat masker output 	                                                            | Other 			| 7.79 MiB  	|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/Anas_platyrhynchos_domestica.RepeatMasker.out.gz |
+		| duck.scafSeq.gapFilled.noMito 					| draft genome assembly                                                             | Sequence assembly	| 1.03 GiB 		|                 | ftp://climb.genomics.cn/pub/10.5524/101001_102000/101001/duck.scafSeq.gapFilled.noMito |
+
+	@wip @files @javascript @pr437
```
pli888, line 320: I think the `@wip` tag should now be `@ok`?

kencho51: This scenario has passed the test, so the tag is updated to `@ok` now.

features/dataset-view.feature
```diff
@@ -277,6 +277,60 @@ Feature: a user visit the dataset page
...
+	@ok @files @javascript @pr437
+	Scenario: Files - Pagination
+		Given I am not logged in to Gigadb web site
+		And I am on "/dataset/101001"
+		And I follow "Files"
+		And I have set pageSize to "5" on "files_table_settings"
+		And I fill in "pageNumber" with "2"
+		And I press "Go to page"
+		When I follow "1"
+		And I take a screenshot named "file_tab_p1"
```
pli888, line 311: This screenshot step needs to be removed as it is only used for debugging.

kencho51: This step is removed.

pli888:
@kencho51 If you look at your [Gitlab CI pipeline for your fork](https://gitlab.com/gigascience/forks/kencho51-gigadb-website/-/pipelines), you will see that this branch is not passing the tests. This should be fixed before I merge your work. This can be done by replacing `phantomjs` with `chrome` in `.gitlab-ci.yml` file:
```
- docker-compose run --rm webapp
- docker-compose up -d phantomjs <~~ change to - docker-compose up -d chrome
# - docker-compose top
```

Also, `ops/deployment/docker-compose.staging.yml` should contain the `chrome` service:
```
    volumes:
      - le_config:/etc/letsencrypt
        
  chrome:
    image: selenium/standalone-chrome:3.141.59-oxygen # latest version with Chrome 74
    shm_size: '1gb' # to avoid a known issue
    ports:
      # - "5900:5900" #for VNC access
      - "4444:4444" #for webdriver access
    networks:
      - web-tier
    extra_hosts:
      - "gigadb.test:172.16.238.10"
      - "gigadb.dev:172.16.238.10"
    environment: # to run headless, set false and comment out port 5900 above and make sure to pass --headless arg in acceptance.suite.yml
      START_XVFB: "false"


networks:
```

rija (commented):
I don't think it's necessary to add the Chrome service into the staging docker-compose file ``docker-compose.staging.yml``.
We don't need to run it on staging.
We only need it on development and CI environments for supporting acceptance testing automation.
On CI, Gitlab is able to pick up the service defined in ``docker-compose.yml`` thanks to this line of ``gitlab-ci.yml``:
```
COMPOSE_FILE: "ops/deployment/docker-compose.yml:ops/deployment/docker-compose.ci.yml"
```
Using that variable, docker-compose is able to merge the services defined in more than one files.

Issue #494, opened 2020-08-04
Local dev GigaDB has no guidelines option on dataset pages


pli888:
On the [GigaDB homepage](http://gigadb.gigasciencejournal.com:9170) in a local dev installation, the drop down `Help` menu has a `Guidelines` option. However, this `Guidelines` option is not present on dataset pages, for example, [/dataset/100004](http://gigadb.gigasciencejournal.com:9170/dataset/100004).

Rahul-coder01:
java -jar DVUploader-v1.0.0.jar -server=<Dataverse server URL> -did=<Dataset DOI> -key=<User's API Key> <file or directory list>

pli888:
> java -jar DVUploader-v1.0.0.jar -server= -did= -key=<User's API Key>

@Rahul-coder01 I am not sure how running a JAR packaged application will help this issue.

PR #497, opened 2020-08-19
Add guidelines to dataset page

kencho51:
# Pull request for issue: #494 

### Changes to the view template files
Allow `guidelines` page redirection in the dataset page by adding `site/guide` to `new_datasetpage.php`.

### Problem
Every changes in the view of `new_main.php`, that have to do the same in the `new_dataset.php` and `new_faq.php` , otherwise these two pages do work properly. Why these two files are separated from the main? Do we have to think about how the changes made in `new_main.php` effect in these two files too?

### DIFF
--- a/protected/views/layouts/new_datasetpage.php
+++ b/protected/views/layouts/new_datasetpage.php
@@ -146,6 +146,7 @@
                             <ul class="dropdown-menu">
                                 <li><a href="/site/help">Help</a></li>
                                 <li><a href="/site/faq">FAQ</a></li>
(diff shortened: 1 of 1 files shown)


pli888:
>Every changes in the view of new_main.php, that have to do the same in the new_dataset.php and new_faq.php , otherwise these two pages do work properly. Why these two files are separated from the main? Do we have to think about how the changes made in new_main.php effect in these two files too?

For the time being, I think we have to keep on having to make these multiple changes until we start work on integrating a front end framework with the GigaDB back end Yii application.

Issue #457, opened 2020-06-23
no ability to remove file attributes


only1chunts:
## User Story

>As a curator,
>I want to manage file attributes from the update form
>So that I can associate various attributes to files

## Acceptance Criteria

>Given there file attributes associated with a dataset
>When I navigate to the update form
>Then I can delete file attributes

## Additional Info

**Describe the bug**
In the Dataset-File admin pages (e.g. http://gigadb.org/adminFile/update/id/417577) we have the ability to add and update File Attributes. However, we have just noticed that we dont have the ability to remove them! In the example above we had the issue whereby the license attribute got added twice somehow, when we tried to remove the duplicate we found there was no delete button.

**To Reproduce**
1. Go to 'http://gigadb.org/adminFile/update/id/417577'
2. Scroll down to the Attributes table
4. Look for the delete button to remove any of the file attributes!

**Expected behavior**
As per most other admin pages we should have the ability to remove attributes from files when required.

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Code is complete
* [ ] Automated tests related to the changes are implemented and passing
* [ ] All automated test suites are passing locally
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] A Pull Request has been created and review requested
* [ ] Pull Request is reviewed and approved
* [ ] The item has been merged to the develop branch
* [ ] All automated test suites are passing on continuous Integration pipeline and item is ready to release

PR #503, opened 2020-09-04
Add delete feature in adminFile page

kencho51:
# Pull request for issue: #457 

1. Separate `deleteFileAttribute` action using Yii's method for routing urls to an action

1. Update `DatasetLog` to record the delete history of upload_status `Published` datasets
2. Update `CurationLog` to record the delete history of upload_status `nonPublished` datasets 

1. Create a `Delete` button in `adminFile/update` page 
2. Display `File Attribute` row in `adminFile/view` page, separate row if more than 1 value

## Example
1. `Delete` button in `/adminFile/update` page
image

2. `File Attribute` rows in `/adminFile/view` page 
image

## Progress
- [x] Create delete file attribute action
- [x] Create `Delete` button in `/adminFile/update` page
- [x] Get log entry in the `curation_log` table
     - [x] `createCurationLogEntry` function working
     - [x] Pass unit test
- [x] Get log entry in the `dataset_log` table
    - [x] `createDatasetLogEntry` function working
    - [x] Pass unit test
- [x] Update `adminFile/view` page to display `File Attribute` row
- [x] Pass `behat` test on 
    1. - [x] Delete button and its delete action
    2. - [x] Delete history of `Published` datasets
    3. - [x] Display file attribute value in `adminFile/view` page
- [x] GitLab CI/CD is passing


pli888:
>1.The actionDelete is not deleting anything, how can I call the delete function properly and delete attribute with file_id?

`actionDelete($id)` in `AdminFileController` class deletes a File and its FileAttributes.

protected/controllers/AdminFileController.php
```diff
@@ -440,6 +440,9 @@ public function actionUpdate($id)
 
                 $this->redirect(array('view', 'id' => $model->id));
             }
+        } elseif (isset($_POST['delete_attr'])) {
+            $args = $_POST['FileAttributes'];
+            FileAttributes::model()->findByPk($args['id'])->delete();
         }
```
kencho51, line 446: @pli888 I tried to create a condition in `actionUpdate`, when I clicked the `Delete!` button, it threw this error:
```
Error 500
Trying to get property of non-object
```

pli888:
Looking at this issue again, I was thinking a Delete button could be added to `adminFile/_form.php` next to the `Edit` button. Clicking the Delete button would execute JS code in a `$('.js-delete').click(function(e)` block, similar to the `js-edit` JS code block that is executed when the Edit button is clicked. The `$('.js-delete').click(function(e)` code block will call a new function `actionDeleteAttr()` in AdminFileController, passing it a file attribute id which it can use to delete the file attribute and return a success response to the JS code. If the JS code block receives a success response then it would reload the `/adminFile/update/id/xxxxx` page with the file attribute deleted probably by using `window.location.reload();`.

features/admin-file.feature
```diff
@@ -0,0 +1,22 @@
...
+  Background:
+    Given Gigadb web site is loaded with production-like data
+    When I go to "/adminFile/update/"
+    Then I should see "Login"
+
+  @wip @issue-457 @javascript
+  Scenario: Sign in as admin and visit admin file update page
+    Given I sign in as an admin
+    When I go to "/adminFile/update/id/88252"
+    Then I should see a button "New Attribute"
```
kencho51, line 15: Hi @rija, I am trying to pass a behat test on this scenario. It threw the following error:
image
It is strange that the same scenario is passing in @pli888 branch. Can you help have a look?

pli888: @kencho51 I did a new checkout of your code and managed to get your test working:
```
Running acceptance tests
+ [[ dev == \d\e\v ]]
+ bin/behat --tags '@wip&&~@facebook&&~orcid' -v --stop-on-failure
@admin-file @issue-457
Feature: An admin user can edit and delete attribute in admin file page
  As an admin user
  I can add/edit/save/delete attribute

  Background:                                                 # features/admin-file.feature:6
    Given Gigadb web site is loaded with production-like data # DatasetViewContext::gigadbWebSiteIsLoadedWithProductionLikeData()
      │ Initializing the database with production_like.pgdmp... Terminating DB Backend... Restarting php container for deployment project...
      │ 
    When I go to "/adminFile/update/"                         # Behat\MinkExtension\Context\MinkContext::visit()
    Then I should see "Login"                                 # Behat\MinkExtension\Context\MinkContext::assertPageContainsText()

  @wip @issue-457 @javascript
  Scenario: Sign in as admin and visit admin file update page # features/admin-file.feature:12
    Given I sign in as an admin                               # GigadbWebsiteContext::iSignInAsAnAdmin()
    When I go to "/adminFile/update/id/88252"                 # Behat\MinkExtension\Context\MinkContext::visit()
    Then I should see a button "New Attribute"                # DatasetViewContext::iShouldSeeAButton()

1 scenario (1 passed)
6 steps (6 passed)
```

kencho51:
@pli888 @rija The behat  tests pass in my branch now. And the `Delete` button is working as shown in the screenshots `before` and `after` pressing the `Delete` button.

This is the screenshot `before pressing Delete button`:
screenshot_before_delete

This is the screenshot `after pressing Delete button`:
screenshot_after_delete

behat.yml
```diff
@@ -18,5 +18,12 @@ default:
...
           selenium2:
-            wd_host: http://phantomjs:8910/wd/hub
+            wd_host: "http://chrome:4444/wd/hub"
+            browser: chrome
+            capabilities:
+              chrome:
+                switches:
+                  - "--headless"
+                  - "--disable-gpu"
+                  - "--window-size=1024,1024"
```
pli888, line 28: This line conflicts with the `develop` branch which has window-size=1024,768. Your tests in `admin-file.feature` pass with this window size so I think we should stick to 1024,768 unless there is a reason to have it as 1024,1024?

features/admin-file.feature
```diff
@@ -0,0 +1,34 @@
+@admin-file @issue-457
+  Feature: An admin user can edit and delete attribute in admin file page
+    As an admin user
+    I can add/edit/save/delete attribute
```
pli888, line 4: This feature is missing the last part:
`So that <something important to my job can happen>`

Maybe @only1chunts can provide his reason for wanting to be able to attach files with attributes?

features/admin-file.feature
```diff
@@ -0,0 +1,34 @@
...
+
+  Background:
+    Given Gigadb web site is loaded with production-like data
+    When I go to "/adminFile/update/"
+    Then I should see "Login"
```
pli888, line 9: I don’t think these last two statements are required in the `Background` section. They appear to be testing the security of the website to ensure that a guest user is not able to access the admin page for Files. This does not need to be tested before each scenario?

features/admin-file.feature
```diff
@@ -0,0 +1,34 @@
...
+    Then I should see a button input "Delete"
+
+  @ok @issue-457 @javascript
+  Scenario: Sign in as admin and delete attribute
+    Given I sign in as an admin
+    When I go to "/adminFile/update/id/13973"
+    And I take a screenshot named "before_delete"
+    And I press "Delete"
+    And I wait "3" seconds
+    Then I take a screenshot named "after_delete"
```
pli888, line 30: I think you need to define some outcomes of what you expected to be observed using `Then` steps. The screenshot steps can be removed when the pull request is ready for merging.

features/admin-file.feature
only1chunts: I guess you are looking for something like:
`So that I can associate various attributes to files`
Do that make sense in this context?
We add attributes like "License:GPLv3" or "PixelResolution:1024" etc the left part of the currie is from the Attributes table, the right part is the value.

behat.yml
kencho51: The screen size of `window-size=1024,768` is not enough to capture the `attribute` row which shows the `Delete` button.

features/admin-file.feature
kencho51: These two statements has been removed. And a new scenario has been created to test the security issue in a new commit.

features/admin-file.feature
kencho51: The scenario has been updated and the screenshot steps have been removed.

behat.yml
pli888: OK, having a common window size of 1920,1080 seems more appropriate then.

features/admin-file.feature
```diff
@@ -0,0 +1,36 @@
...
+    Then I should see "Login"
+
+  @ok @issue-457 @javascript
+  Scenario: Sign in as admin and visit admin file update page and see New Attribute, Edit, Delete buttons
+    Given I sign in as an admin
+    When I am on "/adminFile/update/id/13973"
+    Then I should see a button "New Attribute"
+    And I should see a button input "Edit"
+    And I should see a button input "Delete"
+
```
pli888, line 24: I think you need to test for the presence of the attribute name and its value. Can use `And I should see "attribute_name"` and `And I should see "attribute_value"`.

features/admin-file.feature
```diff
@@ -0,0 +1,36 @@
...
+  @ok @issue-457 @javascript
+  Scenario: Sign in as admin to delete attribute
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/13973"
+    And I should see a button input "Delete"
+    When I press "Delete"
+    And I wait "3" seconds
+    Then I should not see a button "Delete"
```
pli888, line 32: You can also test that the `attribute_name` and `attribute_value` is not present due to them being deleted here.

features/admin-file.feature
kencho51: Updated in the latest commit.

features/admin-file.feature
kencho51: Updated in the latest commit.

rija (commented):
Hi @kencho51

the general approach is fine, here are some suggestions

### 1. User story and scenarios

The user of this functionality is Chris and his team, so the Gherkin user is curator.
There is a precedent in ``features/admins-attach-author-to-user.feature``

Also, edit and delete are two different aspects of managing file attributes (alongside viewing and creating)  so the user story should goes more along the line of:

```
Feature: Managing file attributes 
As a curator
I want to manage file attributes from the update form
So that a file attribute's state stays up-to-date 
```

and the feature file should be called ``admin-file-attributes.feature`` because there are surely other user stories that are related to files administration  and each story should be in its own file.


### 2. AdminFileController

That file is really big, whenever we have to make change to such huge class, we should ask ourselves  whether there's an opportunity to break it down or at least not add more code to it. 

So the new action for deleting a file  attribute should probably be implemented in a separate class

The Yii framework documentation explains how to do that here:

https://www.yiiframework.com/doc/guide/1.1/en/basics.controller#action

and see below how I applied it to ``AdminDatasetController.php`` for the ``moveFiles`` action. 
Notice the ``actions()`` method in the controller, where the  value of the action reflects the path to the action class ``MoveFilesAction.php`` 


* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/controllers/AdminDatasetController.php
* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/controllers/adminDataset/MoveFilesAction.php

(Also note that you still need to keep the change you did in the ``accessRules()`` method)

Finally, I don't see the need to abbreviate attribute in ``deleteAttr``.
We should keep its descriptive name so when we read the code or search the codebase, there's less mental processing and quicker results.
Using ``deleteFileAttribute`` as action name is clearer and avoid confusion with other types of attributes (I can imagine a scenario where a developer do a global search/replace on ``deleteAttr`` only meaning to act on a certain type of attribute and accidentally change all of them)

rija (commented):
### AdminFileController.php

That file is really big, whenever we have to make change to such huge class, we should ask ourselves  whether there's an opportunity to break it down or at least not add more code to it. 

So the new action for deleting a file  attribute should probably be implemented in a separate class

The Yii framework documentation explains how to do that here:

https://www.yiiframework.com/doc/guide/1.1/en/basics.controller#action

and see below how I've applied it to ``AuthorisedDatasetController.php `` for the ``uploadFiles`` action. 

* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/controllers/AuthorisedDatasetController.php
* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/controllers/authorisedDataset/FilesUploadAction.php

Notice the ``actions()`` method in the controller, where the  value of the action reflects the path to the action class ``FilesUploadAction.php `` 

You still need to keep the changes you made to ``accessRules()`` to specifiy access controls for the new action.

kencho51:
Hi @rija ,

> ### User story and scenarios
> The user of this functionality is Chris and his team, so the Gherkin user is curator.
> There is a precedent in `features/admins-attach-author-to-user.feature`
> 
> Also, edit and delete are two different aspects of managing file attributes (alongside viewing and creating) so the user story should goes more along the line of:
> 
> ```
> Feature: Managing file attributes 
> As a curator
> I want to manage file attributes from the update form
> So that a file attribute's state stays up-to-date 
> ```
> 
> and the feature file should be called `admin-file-attributes.feature` because there are surely other user stories that are related to files administration and each story should be in its own file.

The feature contents has been updated and the file name has been updated to `admin-file-attributes.feature`.


> ### AdminFileController.php
> That file is really big, whenever we have to make change to such huge class, we should ask ourselves whether there's an opportunity to break it down or at least not add more code to it.

The `deleteFileAttribute` has been separated from the `AdminFileController`.

All above changes work properly and  `pass` the behat test.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,42 @@
...
+  Feature: A curator can manage file attributes in admin file update page
+    As a curator,
+    I want to manage file attributes from the update form
+    So that I can associate various attributes to files
+
+  Background:
+    Given Gigadb web site is loaded with production-like data
+
+
+  @ok @issue-457 @javascript
```
pli888, line 11: This scenario does not involve javascript-enabled features so does require the @javascript tag to pass?

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,42 @@
...
+  Scenario: Guest user cannot visit admin file update page
+    Given I am not logged in to Gigadb web site
+    When I go to "/adminFile/update/"
+    Then I should see "Login"
+
+  @ok @issue-457 @javascript
+  Scenario: Sign in as admin and visit admin file update page and see New Attribute, Edit, Delete buttons
```
pli888, line 18: This scenario does not involve javascript-enabled features either so does require the @javascript tag to pass?

pli888:
>Check if need to create an entry in curation_log table recording the file attribute deletion

@only1chunts If a curator deletes an attribute for a file then should this action be recorded in the `curation_log` table?

only1chunts:
If the file is in a public dataset then yes the action should be recorded
please.

pli888:
@only1chunts And the `dataset_log` table should also record this file attribution action too?

pli888:
> If the file is in a public dataset then yes the action should be recorded please.

@kencho51 I cannot see a generic function in the codebase that will create a new entry in the `curation_log` table with a given `action` description String. You might have to create a new function to do this in `CurationLogController`.

features/admin-file-attributes.feature
kencho51: @javascript tag removed and passed the test.

features/admin-file-attributes.feature
kencho51: @javascript tag removed and passed the test.

only1chunts:
> @only1chunts And the `dataset_log` table should also record this file attribution action too?

Sorry, I may have mis-understood the initial question. I'll try to clarify here:
If a File Attribute is deleted from a dataset with status PUBLISHED then the action needs to be recorded in the dataset_log (public history of the dataset), it does NOT need to be stored in the curation_log (private history of the dataset).
Ideally, if the dataset status is anything other than PUBLISHED then the action should be stored in the curation_log, but we are less worried about this action than we are about making sure the public history is correct.

rija:
Hi @pli888, @kencho51,

Making these logs a component service is good thinking in general but is not appropriate here, because from my understanding of what @only1chunts says, these two logging tables  are only used in relation to changes to the dataset entity. They are not generic logs that can be available as a service to other entities.

It's enough here to create an additional static method in  ``models/CurationLog.php`` that has the signature for the parameters you want.

During that process, there will be duplication with the existing log creation methods in that class. That's fine:  you can factor out the common code (mostly instantiation but as much as possible) into a static method shared by them all.

You don't need unit testing all the public methods, only unit test the common shared method.
The public methods can be covered by functional and acceptance tests steps that checks for the curation log entry on the dataset form.

Same logic goes for ``models/DatasetLog.php``

kencho51:
Hi @rija 

> It's enough here to create an additional static method in `models/CurationLog.php` that has the signature for the parameters you want.
> 
> During that process, there will be duplication with the existing log creation methods in that class. That's fine: you can factor out the common code (mostly instantiation but as much as possible) into a static method shared by them all.
> 
> You don't need unit testing all the public methods, only unit test the common shared method.
> The public methods can be covered by functional and acceptance tests steps that checks for the curation log entry on the dataset form.
> 
I have updated the `/models/CurationLog.php` and am working on the `unit test`.

protected/tests/unit/CurationLogTest.php
```diff
@@ -0,0 +1,14 @@
...
+class CurationLogTest extends CDbTestCase
+{
+    protected $fixtures=array(
+        'log'=>'CurationLog',
+        'dataset'=>'Dataset',
+    );
+
+    public function testCreateLogEntry()
+    {
+        $this->assertEquals(date("Y-m-d"), $this->log(0)->creation_date, "Different date format with CurationLog");
```
pli888, line 12: I think you need to call the static `createLogEntry($id)` function in the `CurationLog` model class from here. If the `createLogEntry($id)` function returns `false` because of some problem with saving the new curation log entry then you can test for a boolean value here.

kencho51: I am trying to make this code working from the `/models/CurationLog.php/`
```
    public static function createCurationLogEntry($id)
    {
        $model = Dataset::model()->findByPk($id);
        if ($model->upload_status === "Published") {
            $curationlog = new CurationLog;
            $curationlog->creation_date = date("Y-m-d");
            $curationlog->last_modified_date = null;
            $curationlog->dataset_id = $id;
            $curationlog->created_by = "System";
            $curationlog->action = "Status changed to stuff";
            if (!$curationlog->save())
                return false;
        }
``` 
When I click `Delete` in the `file attribute` page, there is no entry in the tbl `curation_log`, am I missing something in the configuration or calling that function not in proper way?

pli888: Have you called this `createCurationLogEntry($id)` function from somewhere after the file attribute is deleted?

pli888: It looks like you are deleting the file attribute in `DeleteFileAttributeAction` class. Try calling the `createCurationLogEntry($id)` function after `$attribute->delete();` to create a new entry in curation_log table after the file attribute is deleted.

kencho51: > ```
> CurationLog::createCurationLogEntry("8");
> ```
This is working. But the `id` is hardcoded, I will try to replace it.

protected/controllers/adminFile/DeleteFileAttributeAction.php
```diff
@@ -0,0 +1,27 @@
...
+            throw new CHttpException(404, "The requested page does not exist.");
+
+        if (isset($_POST['id'])) {
+            $attribute = FileAttributes::model()->findByPk($_POST['id']);
+
+            if ($attribute) {
+                $file = new File;
+                $file->id = $attribute->file_id;
+                $out = print_r($file->dataset_id, true);
+                file_put_contents('fileobj.log', $out);
```
kencho51, line 20: @pli888 I want to get `dataset_id` from  table `file` and then pass it to `createCurationLogEntry`,  but nothing is returned when I check the `fileobj.log`. Can you have a look for me?

pli888: I don't think you need to be creating a new `File` object on line 17 to get hold of the `dataset_id`. You should be able to get it by querying through the properties of objects starting from `$attribute` using `$out = $attribute->file->dataset_id;`

protected/tests/unit/CurationLogTest.php
kencho51: The hardcoded `id` has been replaced.

protected/models/CurationLog.php
```diff
@@ -81,29 +81,46 @@ public function attributeLabels()
...
+        if ($model->upload_status !== "Published") {
+            $curationlog = new CurationLog;
+            $curationlog->creation_date = date("Y-m-d");
+            $curationlog->last_modified_date = null;
+            $curationlog->dataset_id = $id;
+            $curationlog->created_by = "System";
+            $curationlog->action = "Status changed to stuff";
+            if (!$curationlog->save())
+                return false;
+        }
```
kencho51, line 123: Hi @rija I tried to extract the most common code from the static functions, but it did not work. Do you have reference/documentation on how to factor out common code?

protected/tests/unit/CurationLogTest.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+class CurationLogTest extends CDbTestCase
+{
+    public function testNoLastModifiedDate()
+    {
+        $expectation = null;
+//        $this->assertEquals($expectation, CurationLog::createCurationLogEntry(8)->last_modified_date, "Last modified date created");
+        $out = CurationLog::createCurationLogEntry(8);
+        file_put_contents('test.log', var_export($out, true));
```
kencho51, line 10: Hi @rija The `test.log` was an empty file, which means there is no output from `Curation::createCurationLogEntry(8)` function, I was expecting something like:
```
creation_date = 2020-10-19;
last_modified_date = null;
dataset_id = 8;
created_by = "System";
action = "Status changed to stuff";
```
So what is the proper way to call static function for unit test?

rija: Hi @kencho51 

have a look here:

* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/models/CurationLog.php
* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/tests/unit/CurationLogTest.php

protected/models/CurationLog.php
rija: Hi @kencho51 

have a look here:

* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/models/CurationLog.php
* https://github.com/rija/gigadb-website/blob/fuw-cicd/protected/tests/unit/CurationLogTest.php

protected/tests/unit/DatasetLogTest.php
```diff
@@ -0,0 +1,24 @@
...
+        $fileId = "/adminFile/update/id/664";
+        $datasetlog = new DatasetLog(); //Instantiate a new Object
+        $datasetlog->createDatasetLogEntry($datasetId, $fileName, $fileModel, $modelId, $fileId);
+        $this->assertNotNull($datasetlog);
+        $this->assertTrue(is_a($datasetlog, DatasetLog::class));
+//        $this->assertEquals($datasetId, $this->dataset_id);
+//        $this->assertEquals($fileName, $datasetlog->message);
+//        $this->assertEquals($fileModel, $datasetlog->model);
+//        $this->assertEquals($modelId, $datasetlog->model_id);
+//        $this->assertEquals($fileId, $datasetlog->url);
```
kencho51, line 21: Hi @rija I am going to pass this unit test for `createDatasetLogEntry()` function, do you know how to extract the value from `$datasetlog` to do the assertion? I tried a quite different ways, but  only `empty` value was returned.

rija: Hi @kencho51

Often, when testing is hard, it's a sign the design of the implementation is not optimal.

Here, use the factory approach like for CurationLog.

A factory method must return an instance of the class 
(It's like a real factory, whenever the production line is running, it outputs real base product, that can then be customised for specific markets further down the line)

Then you can make another method as part of the public API for DatasetLog (that's what ``createDatasetLogEntry()``seems to be)  that will make use of the factory method.

Also you must use specify return and parameter type definition in all  your functions' signature (I know I forgot to do it on the CurationLog public API methods). E.g:

```
public static function createDatasetLogEntry(int $id, string $fileName, object $fileModel, int $modelId, int $fileId): bool
```

Once you've separated instantiation and customisation and have articulated  the types of parameters and return value,  things (and what to test) will get a bit clearer.

kencho51: Hi @rija 
I have passed the acceptance test for the logging of addition and deletion in `features/dataset-log-file-attribute-add-delete.feature`, and I `assertTrue` for the return value for the function `createDatasetLogEntry()` in the unit test, do you think it is still needed to implement the `factory method`?

rija: Hi @kencho51,

Look at it this way:
As part of the same ticket, you've worked on CurationLog and DatasetLog whose usage mechanism are identical from a code perspective.
Their potential for changes also follow the same pattern (from somewhere in code we need to log info with parameters specific to a context).
How can we justify these two classes having differing implementation?

Often it's the right move to leave out refactoring for later when a new requirement comes up, 
but we should also maintain design consistency whilst we are working on something.

Some developers (most likely our future selves) who will one day stumble on DatasetLog, should go into CurationLog expecting to see the same design as a hint about their similarity and vice-versa, otherwise it's extra mind effort to figure why the difference.

kencho51: Hi @rija
I have implemented the factory pattern `datasetLogEntryFactory` for the function `createDatasetLogEntry`, and the `unit test` is passing.

kencho51:
Hi @pli888 @rija , 
I think this issue is ready for final review. 
1. The `deleteFileAttribute` feature has been implemented and the `behat test` for it is passing.
2. Deleting `Published` dataset file attribute will be logged in the `dataset_log` table, the `unit tests` for `datasetLogEntryFactory` and `createDatasetLogEntry` are passing and `behat test` for this feature is passing. 
3. Deleting non `Published` dataset file attribute will be logged in the `curation_log` table, the `unit test` for `curationLogEntryFactory` is passing.

rija:
Hi @kencho51,

Unit tests are meant to run against the test database.
You don't want to use main database to run unit tests because:

* Testing a unit of code need to be done in isolation of the deployed application to avoid side effects and tainted test outcome
* Unit tests are the basic developer's test harness and needs to run often and fast. So the active test data need to be just enough to run the tests to help that goal and no more, they are called test Fixtures
* Because they are tested in isolation, we need to have full control of their context, including the data they use

So, It's not the best approach to change ``gigadb_test.json`` to run the unit tests against the main database.

The management of test data for unit tests is done using [Yii Fixtures](https://www.yiiframework.com/doc/guide/1.1/en/test.fixture). They are located in ``protected/tests/fixtures``.

I see that in ``CurationLogTest.php`` and ``DatasetLogTest.php``, you have references to a dataset of id 8. Have a look at ``protected/tests/fixtures/dataset.php``  if any of these test dataset records can be used in your test. If not, feel free to add in that fixture file a new record with suitable column data.

pli888:
@kencho51 

>Unit tests are meant to run against the test database.

I understand why we have not been running against the test database. The reason is because we have been using the command below to develop and run CurationLogTest:
```
$ docker-compose run --rm test ./bin/phpunit --testsuite unit --bootstrap protected/tests/bootstrap.php --verbose --configuration protected/tests/phpunit.xml --no-coverage protected/tests/unit/CurationLogTest.php
```

This bypasses the step that creates the test database which is executed in `unit_functional`:
```
psql -U $GIGADB_USER -h $GIGADB_HOST < /var/www/sql/gigadb_unit_tests.sql
```

protected/config/db_test.json
```diff
@@ -1,6 +1,6 @@
 {
-    "database": "gigadb",
+    "database": "gigadb_test",
     "host"    : "database",
-    "user"    : "gigadb",
-    "password": "vagrant"
+    "user"    : "test",
+    "password": "test"
```
kencho51: @pli888 @rija Update database config to `gigadb_test`.

protected/tests/fixtures/dataset_log.php
```diff
@@ -28,6 +28,15 @@
...
 	),
+    array(
+        'id'=>4,
+        'dataset_id'=>1,
+        'message'=>"Pygoscelis_adeliae.scaf.fa.gz: file attribute deleted",
+        'created_at'=>"2020-10-27 16:36:23.106366",
+        'model'=>"FileAttributes",
+        'model_id'=>11020,
+        'url'=>"./bin/adminFile/update/id/664",
+    ),
```
kencho51, line 39: @pli888 @rija Add mockup data here.

protected/tests/fixtures/init.php
```diff
@@ -22,6 +22,7 @@
 	$this->loadFixture("file_format");
 	$this->loadFixture("file_type");
 
+	$this->loadFixture("dataset_log");
```
kencho51, line 25: @pli888 @rija Load in `dataset_log` here, its columns seem more suitable for testing `dataset_log table`.

protected/tests/unit/DatasetLogTest.php
```diff
@@ -3,33 +3,37 @@
...
-        $this->assertEquals($datasetId, $datasetlog->dataset_id);
-        $this->assertEquals("$fileName: file attribute deleted", $datasetlog->message);
-        $this->assertEquals($fileModel, $datasetlog->model);
-        $this->assertEquals($modelId, $datasetlog->model_id);
-        $this->assertEquals("./bin/adminFile/update/id/$fileId", $datasetlog->url);
+        $this->assertEquals($datasetlog->dataset_id, $this->log(3)->dataset_id);
+        $this->assertEquals($datasetlog->message, $this->log(3)->message);
+        $this->assertEquals($datasetlog->model, $this->log(3)->model);
+        $this->assertEquals($datasetlog->model_id, $this->log(3)->model_id);
+        $this->assertEquals($datasetlog->url, $this->log(3)->url);
```
kencho51: @pli888 @rija This test is `passing`.

protected/tests/unit/DatasetLogTest.php
```diff
@@ -3,33 +3,37 @@
...
+//    {
+//        $datasetId = 1;
+//        $fileName = "Pygoscelis_adeliae.scaf.fa.gz";
+//        $fileModel = "FileAttributes";
+//        $modelId = 11020;
+//        $fileId = 664;
+//        $datasetlog = new DatasetLog(); //Instantiate a new Object
+//        $result = $datasetlog->createDatasetLogEntry($datasetId, $fileName, $fileModel, $modelId, $fileId);
+//        $this->assertTrue(true===$result, "No new entry in dataset log table");
+//        $this->assertTrue($datasetlog->isNewRecord);
```
kencho51, line 37: @pli888 @rija This test thrown the following error 
```
CDbException: CDbCommand failed to execute the SQL statement: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "dataset_log_pkey"
DETAIL:  Key (id)=(1) already exists.
```
The mockup data should be `id=4`, so no duplication should be the case. I will try to figure it out.

pli888: @kencho51 `testDataSetLogEntryFactory()` is a bit confusing. What seems to be currently happening is that it is creating a new `DatasetLog` record using the `datasetLogEntryFactory()` function then you are comparing this new `DatasetLog` record's attributes with those from the 4th array in `dataset_log.php`:
```
array(
         'id'=>4,
         'dataset_id'=>1,
         'message'=>"Pygoscelis_adeliae.scaf.fa.gz: file attribute deleted",
         'created_at'=>"2020-10-27 16:36:23.106366",
         'model'=>"FileAttributes",
         'model_id'=>11020,
         'url'=>"./bin/adminFile/update/id/664",
     ),
```

I think in this case you don't need to have any fixtures for this test but need to check that the website application has saved the newly created `DatasetLog` record correctly.

I think the reason for `CDbException` error message is that after the DatasetLog fixtures have been created in `DatasetLog` table then the primary key sequence responsible for creating new ids in `DatasetLog` table becomes out of sync.

rija:
Hi @kencho51,

The implementation of the functionalities looks ok, but there is a problem with the acceptance tests in 
``dataset-log-file-attribute-add-delete.feature``

They run assuming the current state of your development database.
If you start your setup from scratch or someone else try to run the test (like me now), there will be failure.

There seem to be two issues here:

* You need to to have the tests to run from predictable state of data. You do so by adding a step loading a binary dump of a suitable set of data in a Background section.
* I can't see in the scenario any steps that actually make the action for which we expect an entry in history. I guess it passes on your current setup because you've been manually doing the actions as part of testing/development.

Anyway, I don't think there should be a separate feature file for dataset logging, as adding a dataset log entry is part of the functionality for deleting file attribute on a published dataset.
So in that functionality's feature file, ensure you distinguish scenarios for published and unpublished datasets, then for the former you can add a "Then" step along the lines of 
```
...
And I should see a dataset log entry containing "file attribute deleted"
```

features/admin-file-attributes.feature
```diff
@@ -25,7 +38,18 @@
...
+  Scenario: Sign in as admin, delete an attribute of a published dataset and check history tab
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/13973"
+    And I should see "last_modified"
+    When I press "Delete"
+    And I go to "dataset/100056"
+    Then I should see "Termitomyces sp. J132 fungus genome assembly data."
+    And I follow "History"
+    And I should see "History" tab with text "Termitomyces_assembly_v1.0.fa.gz: file attribute deleted"
+
```
kencho51, line 50: Hi @rija 

I found this way `I should see History tab with text` to make sure that the attribute has been actually deleted for `Published` dataset.  
I will try to implement this `And I should see a dataset log entry containing "file attribute deleted"`, do you have any idea that I can access the `dataset_log` table in `context` file?

rija: Hi @kencho51,
I might have answered that to you on Monday already, but to record an answer for the question here, here's the clarification: The history tab displays entries from the dataset_log table, so it's enough to check there is an entry in the history tab using browser navigation (and the step sentence you've come up with for that is fine - although the "text" ending could be dropped or replaced with "entry"). There is no need to access the database table programmatically in the step implementation. Also it defeat the purpose of acceptance testing. 
Think of it as "Black Box Testing", whereas Unit testing is "White Box testing" (you can look those up).
It's ok and often necessary to interact with low-level elements like database or filesystem in "Given ..." steps, because they set up the scene for playing the feature that need to be verified with browser navigation.

kencho51: Hi @rija 

I must have forgotten what you said 😞 
Right now I am working on to pass the `curation_log` test, here is my scenario:
```
  Background:
    Given Gigadb web site is loaded with production-like data

    @wip @issue-#457 @javascript
    Scenario: Go to a non published dataset found in production-like database
      Given I sign in as an admin
      And I go to "/adminFile/update/id/xxx"
      And I should see some file attributes
      When I press "Delete"
      And I should not see file attributes
      And I go to "dataset/YYY"
      And I follow "History"
      Then I should not see some file attributes
```
But I think the above scenario is not a proper way to test.
Because,  I could not identify the correct `xxx` and `YYY`. And even I found the `xxx`, this step `And I go to "dataset/YYY"` will not pass as error `The dataset does not exist` will be found.

So, can I have your suggestion on how to pass the `curation_log` test?

rija: Hi @kencho51,

Among a dataset's attributes in the database table there are ``id``, and ``identifier``.

Controller actions that alter the dataset expect to have the internal id ``id`` as parameter (that's your xxx) as that's the most appropriate one to uniquely identify a row in a database table.

In ``/adminFile/update/id/xxx``, ``/id/`` is how yii framework knows that we are looking for a dataset whose ``id`` attribute is ``xxx``

``identifier`` is for the DOI and is the public identifier for the dataset, that's why it is used in the view as the dataset url is meant to be shared publicly (that's your ``yyy``).

So when you pick a dataset as example, select both attributes from the table.
```
gigadb=# select id, identifier, title from dataset;
id  | identifier |  title
...
 189 | 100133     | Reference data for phage M13 dsDNA generated with the Oxford Nanopore MinION.
 204 | 100143     | Software and supporting material for "SMAP: a streamlined methylation pipelinefor bisulphite sequencing".
 210 | 100155     | Supporting data for "Sparse whole-genome sequencing identifies two loci for major depressive disorder".
...
```

Regarding your second point: datasets that are not published cannot be accessible publicly, so going to the public url ``/dataset/yyy`` would result in that error.

You've got couple of approaches:

You could add Given steps to change the upload status temporarily away from published, but then you would need to find a way to revert the change after the test has run.

Or you could have the Then steps to check this page:
(http://gigadb.gigasciencejournal.com:9170/datasetLog/admin/)
or more precisely manually navigate to the last page, and use that in a Then step.
It's not the most robust (the last page may change if we add/remove example datasets from production-like), but it's the easiest.

A better approach  is to use the test database instead of production-like then checking`` /datasetLog/admin/``'s first page would be enough, but you will need some Given steps to add some file attributes first to a dataset as I think the test database doesn't have any.

An improved alternative to that latter pre-condition would be to update the test data for the test database to have some file attributes in it so you don't have to add extra Given steps (I've pasted bash code to do that easily in @pli888's PR 532, you'd just have to update the corresponding CSV file beforehand)

kencho51: Hi @rija 

I finally passed the acceptance test on `curation log` by checking the last page of `datasetLog/admin`.
This dataset `100245` with `upload_status == Pending`, I manually add attribute to its file with `id == 95354` then delete it.

```
  @ok @issue-457 @javascript @NonPublished
  Scenario: Go to a non published dataset found in production-like database, create then delete a keyword attribute
    Given I sign in as an admin
    And I am on "/adminFile/update/id/95354"
    And I press "New Attribute"
    And I select "keyword" from "FileAttributes_attribute_id"
    And I fill in "FileAttributes_value" with "test Bauhinia"
    And I press "Add"
    And I should see "test Bauhinia"
    And I press "Delete"
    And I should not see "test Bauhinia"
    #Go to the last page of dataset log
    When I go to "datasetLog/admin/DatasetLog_page/74"
    Then I should not see "100245"
    And I should not see "FCHCGJYBBXX-HKBAUpcgEAACRAAPEI-201_L2_2.fq.gz: file attribute added"
    And I should not see "FCHCGJYBBXX-HKBAUpcgEAACRAAPEI-201_L2_2.fq.gz: file attribute deleted"
```

kencho51:
Hi @pli888  and @rija ,

This PR is ready for final review now.

In this PR, feature `delete file attribute` is implemented and the process will be `logged`.  
1. A `Delete` button is added in the `/adminFile/Update/id` page. 
2. Deleting `Published` dataset will be logged in `dataset_log` table, its `unit test` and `acceptance test` are passing. 
3. Deleting  `Non Published` dataset will logged in `curation_log` table, its `unit test` and `acceptance test` are passing.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,86 @@
...
+  @ok @issue-457 @javascript @Published
+  Scenario: Sign in as admin, no delete button should be seen after delete action has been triggered
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/13973"
+    And I should see "last_modified"
+    And I should see "2013-7-15"
```
pli888, line 56: `"last_modified"` and `"2013-7-15"` are strings in a table row. I guess these 2 steps could be improved by replacing them with a check for a table containing a row with these strings, .i.e.:
```
And I should see a table
| Attribute Name | Value | Unit |
| last_modified | 2013-7-15 | |
```



```

**Comment by pli888** [comment]
The `files/html/help.html` file should not be a part of this pull request.

File: protected/controllers/AdminFileController.php [hunk 23]
```diff
@@ -464,41 +464,51 @@ public function actionEditAttr()
...
-        } else
+
+        if (!isset($_GET['ajax'])) {
+            $this->redirect(isset($_POST['returnUrl']) ? $_POST['returnUrl'] : array('admin'));
+        } else {
             throw new CHttpException(400, 'Invalid request. Please do not repeat this request again.');
-      
+        }
+
     }
```
pli888, line 512: Are you sure the new added curly braces are required here?

protected/models/CurationLog.php
```diff
@@ -81,29 +84,48 @@ public function attributeLabels()
...
+     * Retrieves attributes and store them in curation_log table when triggered
+     * @param $id
+     * @return bool
+     */
+    public static function createCurationLogEntry(int $id): bool
+    {
+        $curationlog = self::makeNewInstanceForCurationLogBy($id, "System");
+        $curationlog->action = "Status changed to stuff";
+        return $curationlog->save();
     }
```
pli888, line 129: Is this `createCurationLogEntry(int $id)` function required? The `$curationlog->action = "Status changed to stuff";` line suggests it is not required.

protected/controllers/AdminFileController.php
kencho51: I think they are required, otherwise there will have syntax error.

protected/models/CurationLog.php
kencho51: `createCurationLogEntry(int $id)` is needed for `DeleteFileAttributeAction`, which is responsible to delete file attribute.

kencho51: The function has been updated to `createCurationLogEntry(int $id, string $fileName)`, then it could save delete message in the `action column` in `curation_log` table.

features/admin-file-attributes.feature
kencho51: `I should see a file attribute table` is added to the context file, so it can called from feature file to check the rows in file attribute table.

features/admin-file-attributes.feature
pli888: It might be better to wait a bit longer for [PR #548](https://github.com/gigascience/gigadb-website/pull/548) to be merged then add some file attributes into the relevant CSV file test data. This will then allow the above scenario to be more concise since you wouldn't need to have the steps to add a file attribute for the test to delete.

kencho51: OK.

rija (changes requested):
Hi @kencho51

The implementation looks fine to me now. Just couple things:

#### 1. In ``ops/packaging/Dockerfile``, you should probably replace in line 97,
```
    pecl install xdebug; \
```
with:
```
    pecl install xdebug-2.9.8; \
```

to prevent failure when building containers  (which ones does a lot when jumping between different feature branches)

#### 2. Background  in 'admin-file-attributes.feature'


I noticed there is no Background step related to the admin user. It's better to make explicit that an admin user is needed given every scenarios log in as admin. Also it could happen that the database dump is missing an admin user by accident, in which case it's quicker to have the test fails early, than wait until the first two scenarios finish running before seeing a failure.

```
	Given Gigadb web site is loaded with production-like data
	And an admin user exists
```

features/admin-file-attributes.feature
pli888: PR #548 has now been merged into `develop` branch. If you integrate `develop` with your feature branch then you will be able to add the required test data into the relevant CSV file(s).

kencho51: This branch has been rebased with the `develop` branch, and attribute value `test Bauhinia` is added to the `data/production_like/file_attributes.csv`. So the following steps could be removed.
```
And I press "New Attribute"
And I select "keyword" from "FileAttributes_attribute_id"
And I fill in "FileAttributes_value" with "test Bauhinia"
And I press "Add"
```

pli888:
@kencho51 Can you check if you are able to delete the following file attributes for these datasets on a fresh `dev` installation:

* http://gigadb.gigasciencejournal.com:9170/adminFile/update/id/5102
* http://gigadb.gigasciencejournal.com:9170/adminFile/update/id/664
* http://gigadb.gigasciencejournal.com:9170/adminFile/update/id/5103
* http://gigadb.gigasciencejournal.com:9170/adminFile/update/id/5107

I get an `Error 500 trying to get property of non-object` message when I try to delete the above file attributes.

pli888:
@kencho51 can you check if you are seeing an `error 500 Undefined variable: attribute_id` at http://gigadb.gigasciencejournal.com:9170/adminFile/view/id/663? The log says the problem is this:
```
2021/03/26 02:36:20 [error] [php] Undefined variable: attribute_id (/var/www/protected/views/adminFile/view.php:54)
```

ops/packaging/Dockerfile
```diff
@@ -94,7 +94,7 @@ ARG INSTALL_XDEBUG=false
 RUN if [ ${INSTALL_XDEBUG} = true ]; then \
   # Install the xdebug extension
   if [ $(php -r "echo PHP_MAJOR_VERSION;") = "5" ]; then \
-    pecl install xdebug-2.5.5; \
+    pecl install xdebug-2.9.8; \
```
pli888, line 97: @kencho51 I'm not sure why version `2.9.8` is specified here as the same version number is also below?

kencho51: @pli888 I just want to make sure only xdebug-`2.9.8` will be installed when building the containers. Or should I remove the `else` block?

pli888: @kencho51 The top of `Dockerfile` says an image with PHP version 7 will be used when building containers with `Dockerfile`. Based on the conditional statements then `xdebug-2.9.8` is installed. You can check this by dropping into the `test` container and checking the version of xdebug installed:
```
$ docker-compose run --rm test bash
Creating deployment_test_run ... done
root@bf59cce4c0cb:/var/www# php -v
PHP 7.1.30 (cli) (built: Jun 11 2019 04:18:15) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.1.30, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.9.8, Copyright (c) 2002-2020, by Derick Rethans

root@bf59cce4c0cb:/var/www# pear list
Installed packages, channel pecl.php.net:
=========================================
Package   Version State
apcu      5.1.8   stable
libsodium 2.0.23  stable
mailparse 3.1.1   stable
xdebug    2.9.8   stable
```

kencho51:
>   an `error 500 Undefined variable: attribute_id` at http://gigadb.gigasciencejournal.com:9170/adminFile/view/id/663? The log says the problem is this:
> 
> ```
> 2021/03/26 02:36:20 [error] [php] Undefined variable: attribute_id (/var/www/protected/views/adminFile/view.php:54)
> ```

It is because at first the `attribute_id` was queried from the `FileSample`, which does not contain some files, like `663` and `80845` in `devdb` database, so it would return empty database object.
It is fixed now by getting it from `FileAttribute` table with `file_id=$modle->id`, and if there is no `file attribute` return, `Not Set` will be indicated in the `File Attribute Id` row.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,99 @@
...
+    And I should not see "last_modified"
+    And I should not see "2013-7-15"
+    And I should not see a button "Delete"
+    And I press "Save"
+    Then I go to "/adminFile/view/id/13973"
+    And I should see field "File Attribute Id" without value "1 "
+    And I go to "dataset/100056"
+    And I should see "Termitomyces sp. J132 fungus genome assembly data."
+    And I follow "History"
+    And I should see "History" tab with text "Termitomyces_assembly_v1.0.fa.gz: file attribute deleted"
```
pli888, line 63: @kencho51 I think its better to break this scenario into smaller scenarios because it appears there are observed outcomes in `Given` steps and in When` steps. For example, `And I should see field "File Attribute Id" with value "1 "` sounds like an observed outcome but it is in a `Given` step. Outcomes to be observed should be in `Then` steps. Take a look at scenarios in other features and the behat [documentation](https://docs.behat.org/en/latest/user_guide/writing_scenarios.html).

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,99 @@
...
+    And I press "Delete"
+    And I should not see "test photo"
+    And I press "Save"
+    And I go to "/adminFile/view/id/95354"
+    And I should see field "File Attribute Id" without value "5442 "
+    #Go to the last page of dataset log
+    And I go to "datasetLog/admin/DatasetLog_page/74"
+    And I should not see "100245"
+    And I should not see "FCHCGJYBBXX-HKBAUpcgEAACRAAPEI-201_L2_2.fq.gz: file attribute added"
+    And I should not see "FCHCGJYBBXX-HKBAUpcgEAACRAAPEI-201_L2_2.fq.gz: file attribute deleted"
```
pli888, line 92: @kencho51 I think this scenario could be broken down into smaller scenarios if you want to test the observed outcomes in `Given` and `When` steps.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I follow "History"
+    Then I should see "History" tab with text "File Termitomyces_assembly_v1.0.fa.gz updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.cds.fa updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.cds.fa updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.gff updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.gff updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.pep.fa updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.pep.fa updated"
+
+  @ok @Published
```
pli888, line 31: @kencho51 If you look at tags in other features, they are always lower case. Let's be consistent so change this to `@published`.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I should see a button input "Delete"
+    And I should see a button input "Save"
+
+  @ok @javascript @Published
+  Scenario:  Go to a published dataset found in production-like database, should see a last modified attribute
+    Given I sign in as an admin
+    And  I go to "/adminFile/update/id/13973"
+    And I should see a file attribute table
+    | Attribute Name | Value     | Unit |
+    | last_modified  | 2013-7-15 |      |
```
pli888, line 49: @kencho51 Let's move this observed outcome after the `When` step.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    When I go to "/adminFile/view/id/13973"
+    Then I should see field "File Attribute Id" with value "1 "
+
+  @ok @javascript @Published
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute and save
+    Given I sign in as an admin
+    And  I go to "/adminFile/update/id/13973"
+    And I should see a file attribute table
+      | Attribute Name | Value     | Unit |
+      | last_modified  | 2013-7-15 |      |
```
pli888, line 59: @kencho51 You've already tested this step in the scenario above. I would delete this.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    When I press "Delete"
+    Then I should not see "last_modified"
+    And I should not see "2013-7-15"
+    And I should not see a button "Delete"
+    And I press "Save"
+    And I go to "/adminFile/view/id/13973"
+    And I should see field "File Attribute Id" with empty value
```
pli888, line 66: @kencho51 I would move the last 3 steps into a new scenario that checks the `Save` commit of deleting the file attribute, otherwise, you have observed outcomes which are followed by the press "Save" key action.

Also, the website automatically goes to "/adminFile/view/id/13973" after pressing "Save"? You could just check that you are on "/adminFile/view/id/13973" if you wait a couple of seconds as part of this scenario instead of explicitly going to the page.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    Then I should see "History" tab with text "File Termitomyces_assembly_v1.0.fa.gz updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.cds.fa updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.cds.fa updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.gff updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.gff updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.pep.fa updated"
+    And I should see "History" tab with text "File Termitomyces_gene_v1.0.pep.fa updated"
+
+  @ok @published
+  Scenario: Sign in as admin and visit admin file update page and see New Attribute, Edit, Delete, Save buttons
```
pli888, line 32: Your `Scenario` descriptions could be shortened and therefore faster to read. For example, you don't need to say `Sign in as Admin` because it is already in the `Given` line immediately below. Therefore the `Scenario` could be `See Attribute, Edit, Delete, Save buttons on admin file update page`.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I should see a button input "Save"
+
+  @ok @javascript @published
+  Scenario:  Go to a published dataset found in production-like database, should see the File Attribute Id of the last modified attribute
+    Given I sign in as an admin
+    When I go to "/adminFile/view/id/13973"
+    Then I should see field "File Attribute Id" with value "1 "
+
+  @ok @javascript @published
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute
```
pli888, line 50: This could be shortened by removing `found in production-like database` as we already know this in the `Background` section above.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/13973"
+    When I press "Delete"
+    Then I should not see "last_modified"
+    And I should not see "2013-7-15"
+    And I should not see a button "Delete"
+
+  @ok @javascript @published
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute and save
```
pli888, line 59: This could be `Delete file attribute on file admin update page`.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I should see field "File Attribute Id" with empty value
+
+  @ok @javascript @published
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute and check for history tab
+    Given I sign in as an admin
+    And I go to "/adminFile/update/id/13973"
+    And I should see a file attribute table
+      | Attribute Name | Value     | Unit |
+      | last_modified  | 2013-7-15 |      |
+    And I should see a button input "Delete"
```
pli888, line 74: Do we need this step and the step above since it is already tested in `Scenario: Sign in as admin and visit admin file update page and see New Attribute, Edit, Delete, Save buttons`?

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute and save
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/13973"
+    When I press "Delete"
+    And I press "Save"
+    Then I am on "/adminFile/view/id/13973"
+    And I should see field "File Attribute Id" with empty value
+
+  @ok @javascript @published
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute and check for history tab
```
pli888, line 68: The `Scenario` could be shortened to `File attribute deletion is recorded in History tab` which is because of an update to the `dataset_log` table.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I should see a button input "Delete"
+    When I press "Delete"
+    And I press "Save"
+    Then I go to "dataset/100056"
+    And I should see "Termitomyces sp. J132 fungus genome assembly data."
+    And I follow "History"
+    And I should see "History" tab with text "Termitomyces_assembly_v1.0.fa.gz: file attribute deleted"
+
+  @ok @javascript @nonPublished
+  Scenario: Go to a non published dataset found in production-like database, should see a keyword attribute and a camera parameters attribute and their File Attribute Id
```
pli888, line 83: Scenario could be `Display attributes in non-published dataset`. I would think about separating into 2 scenarios, one scenario to check display of attributes on file update page and second scenario to check display for file attributes on file view page.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/95354"
+    And I should see a file attribute table
+    | Attribute Name    | Value         | Unit |
+    | keyword           | test Bauhinia |      |
+    | camera parameters | test photo    |      |
+    When I go to "/adminFile/view/id/95354"
+    Then I should see field "File Attribute Id" with value "5441 5442 "
```
pli888, line 91: Displaying the file attribute values to the curators on the admin file view page will be more useful than displaying the file attribute identifiers?

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+
+  @ok @javascript @nonPublished
+  Scenario: Go to a non published dataset found in production-like database, delete a keyword attribute
```
pli888, line 94: Could shorten scenario to `Delete keyword attribute from non published dataset`?

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+    Then I am on "/adminFile/view/id/95354"
+    And I should see field "File Attribute Id" with empty value
+
+  @ok @javascript @nonPublished
+  Scenario: Go to a non published dataset found in production-like database,  delete a file attribute and check the last page of dataset log
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/95354"
+    When I press "Delete"
+    And I press "Save"
+    Then I go to "datasetLog/admin/DatasetLog_page/74"
```
pli888, line 136: Deleting a file attribute in a non-published dataset create a new entry in the `curation_log` table. Is there a `CurationLog_page` that you could go to so that the Behat test can check that the `FCHCGJYBBXX-HKBAUpcgEAACRAAPEI-201_L2_2.fq.gz: file attribute deleted` action is displayed?

protected/views/adminFile/view.php
```diff
@@ -8,17 +8,32 @@
...
+$attribute_id = "";
+
+ if(!empty($attributes))
+ {
+     foreach ($attributes as $attribute)
+     {
+         $attribute_id .= $attribute->id . " ";
+     }
+ }else{
+     $attribute_id = "Not Set";
```
pli888, line 34: If a `File` object does not have a `Sample` or `Attribute` associated with it then I would not display these rows at all in the admin file view page.

protected/controllers/adminFile/DeleteFileAttributeAction.php
```diff
@@ -0,0 +1,33 @@
...
+    public function run()
+    {
+        if (!Yii::app()->request->isPostRequest)
+            throw new CHttpException(404, "The requested page does not exist.");
+
+        if (isset($_POST['id'])) {
+            $attribute = FileAttributes::model()->findByPk($_POST['id']);
+
+            if ($attribute) {
+                $out = $attribute->file->dataset_id;
```
pli888, line 17: Instead of `$out`, I would name this variable `$dataset_id`.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,142 @@
...
+  Scenario: Go to a published dataset found in production-like database, delete a last modified attribute and check for history tab
+    Given I sign in as an admin
+    And I go to "/adminFile/update/id/13973"
+    And I should see a file attribute table
+      | Attribute Name | Value     | Unit |
+      | last_modified  | 2013-7-15 |      |
+    And I should see a button input "Delete"
+    When I press "Delete"
+    And I press "Save"
+    Then I go to "dataset/100056"
```
pli888, line 77: Should it be `"/dataset/100056"`?

features/admin-file-attributes.feature
kencho51: This step is to check the history tab containing the delete message or not after a file attribute is deleted. So, I think it is needed.

features/admin-file-attributes.feature
kencho51: There is no such page `CurationLog_page`. In fact, I was thinking, now we could check whether a file attribute has been deleted or not by checking the `File Attribute Id` on `admin file view` page, so checking the message in history tab could be removed?

protected/views/adminFile/view.php
kencho51: Updated the table to show `File Attribute Value` row only if it is not empty, otherwise, the row will not be displayed.

pli888:
file_view

In the file view, I think `File Attribute Value` can be shortened to `File Attribute`. In all the other rows, we kind of assume that they are all values without the need to state this.

.gitignore
```diff
@@ -74,3 +74,4 @@ docs/cache
 
 !style-guide/vendor
 !style-guide/index.php
+
```
pli888, line 77: Sometimes in your pull requests, there are new lines appearing at the end of files like this one. In your PHPStorm, in Preferences > Editor > General, is the `Ensure every saved file ends with a line break` option checked?

features/admin-file-attributes.feature
pli888: I don't mean the Scenario. I mean these 2 steps:
```
And I should see a file attribute table
       | Attribute Name | Value     | Unit |
       | last_modified  | 2013-7-15 |      |
     And I should see a button input "Delete"
```

They are the same in the above `Scenario: See Attribute, Edit, Delete, and Save buttons on admin file update page`.

features/admin-file-attributes.feature
pli888: What you suggest is already tested in the above `Scenario: Delete all attributes and save, File Attribute Value on admin file view page should be empty`?

kencho51: Ok.

features/admin-file-attributes.feature
kencho51: These two steps have been removed.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,132 @@
...
+  Scenario: Delete a last modified attribute on admin file update page
+    Given I sign in as an admin
+    And I am on "/adminFile/update/id/13973"
+    When I press "Delete"
+    Then I should not see "last_modified"
+    And I should not see "2013-7-15"
+    And I should not see a button "Delete"
+
+  @ok @javascript @published
+  Scenario: File Attribute value is empty empty after deleting an attribute and saving
```
pli888, line 60: `empty` is written twice here.

features/admin-file-attributes.feature
```diff
@@ -0,0 +1,132 @@
...
+    And I should see a file attribute table
+    | Attribute Name    | Value         | Unit |
+    | keyword           | test Bauhinia |      |
+    | camera parameters | test photo    |      |
+
+  @ok @javascript @nonPublished
+  Scenario: See File Attribute Value on admin file view page
+    Given I sign in as an admin
+    When I go to "/adminFile/view/id/95354"
+    Then I should see field "File Attribute Value" with "test Bauhinia test photo "
```
pli888, line 92: `test Bauhinia test photo` - these are 2 file attribute values joined together? If so, I think it's better to split them into 2 separate file attributes otherwise it looks like it is a single file attribute with a value `test Bauhinia test photo`.

kencho51: Now, the file attribute value will be separated by `,` like this `test Bauhinia, test photo`

kencho51:
Hi @pli888 ,

I tried to replace the `CDetailView` widget with html table, and the view looks like this:
image

Will try to pass its behat test.

pli888:
>Is it OK?

Instead of the new CSS and table rendering code, I was thinking that you could push separate arrays of key value pairs into `$attributes` which are then rendered by the `CDetailView` widget:
```
$attributes = array(
      'id',
      'dataset_id',
      'name',
      'location',
      'extension',
      'size',
      'description',
      'date_stamp',
      'format_id',
      'type_id',
      array(
          'name'=>'Sample',
          'value'=> $name)
  );

  if(!empty($fileAttributes))
  {
      foreach ($fileAttributes as $fileAttribute)
      {
          array_push($attributes, array('name' => 'File Attribute', 'value' => $fileAttribute->value ));
      }
  }
 
$this->widget('zii.widgets.CDetailView', array(
  'data' => $model,
  'attributes' => $attributes
));
```

kencho51:
Hi @rija and @pli888 ,

Please have a look of this PR, you may find the detailed description at [here](https://github.com/gigascience/gigadb-website/pull/503#issue-479521155).

ScottBGI:
One minor piece of feedback noticed from the demo is you need to update the BGI and CNGB logos. You might want to replace GigaScience logo with the GigaScience Press one too.

kencho51:
Hi Scott,

> One minor piece of feedback noticed from the demo is you need to update the BGI and CNGB logos. You might want to replace GigaScience logo with the GigaScience Press one too.

Those logo would only be seen when you have logged into the `admin` page, while the `admin` page is in grey layout, the new logos  would be barely visible as their background colour looks alike.

rija (approved):

Issue #293, opened 2019-03-20
bug fixes to github version


only1chunts:
there are a number of things that have been fixed in the production version recently that have not made it into the master branch on github for various reasons:

- [x] http://gigadb-staging-jesse.gigatools.net/site/help still has the old submission guidelines

- [x] http://gigadb-staging-jesse.gigatools.net/site/term the "General" heading is missing and the other headings are odd font format

- [x] http://gigadb-staging-jesse.gigatools.net/site/team Our team page has had some updates of pictures and text

- [x] the login page still has the ORCID option to login with, we removed that because we cant get a contact email from them if they use that.

- [x] The map browser of samples is not working http://gigadb-staging-jesse.gigatools.net/site/mapbrowse

- [x] The new feature of a User being able to claim authorship by associating their user account to an author name is working except for the email notification, database@ should receive a notification if an user tries to link an authors name to their account. The website tells me an email has been sent, but I dont see it in the mailbox.

- 6 FAQ's missing:
- [x] Why do you request so many sample attributes?
- [x] Why is the directory structure missing from the file table view on my dataset page?
- [x] What should I do if I accidentally identify an individual from anonymized human (meta)data within a dataset?
- [x] What curation do you carry out?
- [x] What procedures are in place to ensure data integrity?
- [x] What data storage procedures do you follow?

- [x] due to the missing style sheet in the admin pages (#289) its very difficult to tell if there anythings missing there at the moment!

pli888:
@jessesiu how much work has been completed for this issue?

Rahul-coder01:
Why the website is offline..

only1chunts:
Dear @Rahul-coder01 ,  [GigaDB.org](www.gigadb.org) is the live website location and it is online. If you are referring to the temporary url listed in the above ticket this was only a temporary development staging version used for testing this particular update/code change. Please feel free to email me (chris@gigasciencejournal.com) if you require any information about GigaDB.org. Regards Chris.

pli888:
>due to the missing style sheet in the admin pages (#289) its very difficult to tell if there anythings missing there at the moment!

@only1chunts This is fixed now - see [staging server](http://gigadb.gigasciencejournal.com:9170/site/admin).

pli888:
@only1chunts 

[...]

PR #504, opened 2020-09-11
Using swiftmailer for sending email notifications

pli888:
# Pull request for issue: gigascience/gigadb-website#293

Sending email notifications to curators about authorship claims on datasets made by users. Emails are sent using a new `MailService` component as suggested by @rija in gigascience/gigadb-website#496.

## Changes to the component classes

There is a new `MailService` class in the `protected/components` directory containing 2 functions, `sendEmail` and `sendHTMLEmail` for sending emails in plain text and HTML formats. Both functions use `yii2-swiftmailer` for sending emails from GigaDB. HTML emails are composed by embedding the message content in a HTML table provided by `mail/template.php` which is in turn embedded into the `<body>` tag in `mail/layouts/html.php`.

The `sendNotificationEmail($user, $author, $dataset)` function in `UserCommandController` class is responsible for sending emails to notify curators when a user has claimed authorship of a dataset. This function has been updated to call the `sendHTMLEmail` function in `MailService`.

## Changes to the tests

A new `MailServiceTest` class tests the email sending functionality. For testing, swiftmailer emails are configured to be saved as eml files using `test.php.dist`. `MailServiceTest` contains a `testItShouldSendEmailUsingYii2Swiftmailer` function to test that the Yii2 swiftmailer package can be called by the GigaDB Yii1 application. The eml email message that is created is checked for the correct sender and recipient email addresses, subject and message body. `MailServiceTest` also has a `testItShouldSendEmailUsingMailService` function that checks that the `sendEmail` and `sendHTMLEmail` functions in the `MailService` component class are able to send emails which it does via the Yii2 swiftmailer package.

I also looked into the use of Mailtrap.io for testing email functionality.

`composer.json.dist` has been updated to install the `yii2-swiftmailer` package in the `vendor` directory. In addition, extra configuration is required to ensure that the GigaDB Yii application is able to find classes in the `swiftmailer` package. 

Configuration to use the `MailService` class as a new component can be found in `local.php.dist`. Configuration to use Yii2 swiftmailer package is in `web.php.dist`. `generate_config.sh` has been updated to generate `local.php` and the `test.php` and `web.php` files in `protected/config/yii2` directory.

## Example of email sent to curator

email


rija (commented):
Hi @pli888,

* In ``UserCommandController.php``,

It's probably just an omission, but hard-coded email should be replaced in:
```
Yii::app()->mailService->sendHTMLEmail("database@gigasciencejournal.com",
```
by the ``$recipient`` variable already used elsewhere in the function.

* In ``MailServiceTest.php``,

It was good idea to test wiring of yii2-swiftmailer first, and then MailService.php functions in separate tests like you did. The method for picking up the last message from ``eml_dir`` is quite elegant. I want to reuse it in my branch to replace the time-based approach.

The test for sending HTML mail should problably contain specific assertions to test the image parts of the email, otherwise from a test perspective, the two methods are doing the same thing and we don't know whether the email templating worked for images.

pli888:
@rija 

>It's probably unintended, but hard-coded email should be replaced in:
>`Yii::app()->mailService->sendHTMLEmail("database@gigasciencejournal.com",`
>by the $recipient variable already used elsewhere in the function.

The hard-coded sender email address has been replaced with `$recipient` now.

>The test for sending HTML mail should problably contain specific assertions to test the image parts of the email, otherwise from a test perspective, the two methods are doing the same thing and we don't know whether the email templating worked for images.

The `testItShouldSendHTMLEmailUsingMailService` function in `MailServiceTest` has been improved. Using [php-mime-mail-parser](https://github.com/php-mime-mail-parser/php-mime-mail-parser), the function will query the attachment filenames making sure they match the name of the image files, this ensuring that a HTML email has been sent.

Using [php-mime-mail-parser](https://github.com/php-mime-mail-parser/php-mime-mail-parser) required updating `composer.json.dist` so that the package is installed in the `require-dev` section.  This package itself requires the mailparse extension (http://php.net/manual/en/book.mailparse.php) installed so this has been done `Dockerfile` and is installed for the `test` container. I've also had to declare that `mailparse` is installed in the `application` container using `INSTALL_MAILPARSE=true` otherwise `composer` complains that `mailparse` is missing when `docker-compose run --rm webapp` is executed.

rija (approved):
@pli888 

Thanks. It looks good to me and tests are all passing for me.

Issue #515, opened 2020-10-09
No longer being indexed in Google search


only1chunts:
## User Story
> As a human
> I want to find information held in GigaDB when I Google search for things
> so that I get all the relevant information

## Acceptance Criteria
> Given I do a google search for the title of a dataset `Supporting data for "TransHLA: A Hybrid Transformer Model for HLA-Presented Epitope Detection"`
>When I view the search results
> Then the GigaDB dataset page https://gigadb.org/dataset/102633 is one of the top hits

**Describe the bug**
It appears we're no longer being indexed by google search as none of our datasets are appearing in search results even when I use a dataset title to search with?!

**To Reproduce**
1. Go to google.com
2. Search for "Genome of the small hive beetle"
3. Scroll down results you will see the GigaScience paper with the title starting "Genome of the small hive beetle" is the top hit, but there is NO result for the GigaDB dataset (http://dx.doi.org/10.5524/100511)

**Expected behavior**
The GigaDB dataset should appear somewhere in the google search results

**Additional context**
I have no evidence, but I'm certain we used to appear in Google hits when searching for things that are in our database.

only1chunts:
Jesse removed the metatags, this is what he wrote in his email to me:
`I think it should be the reason that prevents Google crawlers from indexing our dataset pages.

Just removed the <meta> noindex tags in the dataset page, and checked the log file, we get lots of traffic from search engine web crawlers including the google now.

66.249.79.26 - - [11/Nov/2020:13:41:41 +0000] "GET /dataset/view/id/100334/File_page/12/Sample_page/8/File_sort/type_id HTTP/1.1" 200 13482 "-" "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"
`
That was 5 days ago. I still see no results in a normal google search for anything that I would expect to see GigaDB dataset results for, but now it does appear that all our datasets are being indexed in the google-datasets search:
https://datasetsearch.research.google.com/search?query=gigadb
NB - I'm not sure if they were before or not.

only1chunts:
@jessesiu,
CNGB are thinking the server issues might be being caused by 1000's of bots and crawlers taking up all the bandwidth. 
Presumably, the bots can only get the FTP server addresses from the webpages so perhaps if we can add a meta tag for nofollow to each dataset page it would prevent the bots from trying to index the FTP server?

From this website it looks like the header code we would need is:

<meta name="robots" content="index, nofollow"/>

[...]

Issue #85, opened 2016-11-10
Add the metadata schema on dataset page to allow Hypothesis to parse citation DOI


only1chunts:
## User Story

>As a researcher
>I want the Hypothesis annotation tool to recognise a GigaDB dataset page url as an alias of its DOI
>So that I can correctly cite it irrespective of which website it appears on or I annotate from

## Acceptance Criteria

>Given I am not logged in to Gigadb web site
>When I go to the primary dataset url "/dataset/100002"
>And I view the page source
>Then I can see a meta-tag for a citation DOI predicate with value "10.5524/100002"

>Given I am not logged in to Gigadb web site
>When I go to the url "/dataset/view/id/100002/File_sort/date_stamp"
>And I view the page source
>Then I can see a meta-tag for a citation DOI predicate with value "10.5524/100002"

## Additional Info

This could be done in conjunction with #73 as its a related concept.

Following an email discussion with Jon Udell at Hypothes.is we should include an additional flag in the HTML code of dataset pages to inform machines about the connection between the webpage and the DOI.

Hi Chris,
Here is the short answer: I think pages like http://gigadb.org/dataset/100132 should include statements like:
<meta name="citation_doi" content="10.5524/100132">
How we use that metadata to alias documents is a mechanism that has changed, and will likely change again. But informing Hypothesis about the DOI of each GigaDB page is a good way to futureproof your namespace.
Regards,
Jon

For example see the source code for this article:
view-source:http://www.jneurosci.org/content/34/17/6112
among all that extra metadata included in it, is this line:

<meta name="citation_doi" content="10.1523/JNEUROSCI.5212-13.2014" />

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Code is complete
* [ ] Automated tests related to the changes are implemented and passing
* [ ] All automated test suites are passing locally
* [ ] Code is refactored to best practices and coding standards
[...]

PR #516, opened 2020-10-14
Add hypothesis metadata schema #85

kencho51:
# Pull request for issue: #85 

Add Hypothesis `Highwire Press Tags` to the /`dataset/view.php`.

## Changes to the tests
Create acceptance test for the presence of `meta` tag with `DOI` value.
 
## Progress
- [x] Add Hypothesis `meta data tag` to the `view class`
- [x] Add `fullDOI` variable to the `Controller class`
- [x] Allow search engines to crawl and index GigaDB pages for #515 
- [x] Pass the `Behat test`


rija (changes requested):
Hi @kencho51
 
the scenario step
```
    Then I should see a "DOI" related links to "10.5524/100002"
```
doesn't quite capture the desired outcome here.
Use the scenario step mentioned in #85. That step doesn't have an implementation yet,  you'll have to implement it so it check in the source  that there is a meta-tag with the corresponding name and content

features/dataset-metadata-citation-doi.feature
```diff
@@ -0,0 +1,14 @@
...
+  I want the Hypothesis annotation tool to recognise a GigaDB dataset page url as an alias of its DOI
+  So that I can correctly cite it irrespective of which website it appears on or I annotate from
+
+  @wip @javascript @issue-85
+  Scenario: Go to dataset page and there is a meta-tag with DOI
+    Given I am not logged in to Gigadb web site
+    When I go to "/dataset/100002"
+    Then I should see "Genomic data from Adelie penguin (Pygoscelis adeliae)"
+    And There should be a meta-tag element has attribute "name" with value "citation_doi"
+    And There should be a meta-tag element has attribute "content" with value "10.5072/100002"
```
kencho51, line 13: Hi @rija This acceptance test has to use doi `10.5072/10002` to pass, the doi `10.5524/100002` you mentioned is not working.  Is `10.5072/10002` the correct doi? :rofl:

rija: @kencho51 
yes, ``10.5072`` is fine.

rija (changes requested):
Hi @kencho51,

The acceptance test still passes if I replace the meta tag addition with the incorrect lines below
```
<meta name="citation_doi">
<meta content="<?= $mainSection->getReleaseDetails()['full_doi']; ?>">
```
you will need to to test the values of the "name" and "content" attributes  together and in one step, not separately.

rija (changes requested):
Hi @kencho51

Nice, I see there is now one step only for checking the meta-tag.
However the implementation is still problematic (the example from my previous comment will still pass the test).

Looking at your code:

```
metaNode = $this->minkContext->getSession()->getPage()->find('css', "meta[$arg1='$arg2']");
        if( $metaNode->hasAttribute($arg1) && $metaNode->hasAttribute($arg3)) {
            PHPUnit_Framework_Assert::assertEquals($arg2, $metaNode->getAttribute($arg1));
            PHPUnit_Framework_Assert::assertEquals($arg4, $metaNode->getAttribute($arg3));
        }
```

If the meta-tag element doesn't have both attributes , the assertions in the ``if`` block won't get executed, and the function will return true and pass the test.
A quick fix to the above involves adding an else block that capture and fail the incorrect meta-tag case:

```
      $metaNode = $this->minkContext->getSession()->getPage()->find('css', "meta[$arg1='$arg2']");
        if( $metaNode->hasAttribute($arg1) && $metaNode->hasAttribute($arg3)) {
            PHPUnit_Framework_Assert::assertEquals($arg2, $metaNode->getAttribute($arg1));
            PHPUnit_Framework_Assert::assertEquals($arg4, $metaNode->getAttribute($arg3));
        }
        else {
            PHPUnit_Framework_Assert::fail();
        }
```

However, there's a better way. CSS expressions are not good for complex queries like this. XPath is easier here: 
``
//meta[@name='$name' and @content='$content']
``

Also, the phrasing for the scenario step is not specific enough and doesn't follow the usual "I should see..." pattern. 

Taking both above observations together and you get the more concise:

```
    /**
     * @Then I should see a meta-tag with name :name and content :content
     */
    public function iShouldSeeMetaTagWithNameAndWithContent($name, $content) {
        $metaNode = $this->minkContext->getSession()->getPage()->find(
            'xpath', 
            "//meta[@name='$name' and @content='$content']"
        );
        PHPUnit_Framework_Assert::assertNotNull($metaNode);
    }
```

features/bootstrap/DatasetViewContext.php
```diff
@@ -417,14 +417,13 @@ public function iHaveSetPagesizeToOn($arg1, $arg2)
...
     {
-        $metaNode = $this->minkContext->getSession()->getPage()->find('css', "meta[$arg1='$arg2']");
-        if( $metaNode->hasAttribute($arg1) && $metaNode->hasAttribute($arg3)) {
-            PHPUnit_Framework_Assert::assertEquals($arg2, $metaNode->getAttribute($arg1));
-            PHPUnit_Framework_Assert::assertEquals($arg4, $metaNode->getAttribute($arg3));
-        }
+        $metaNode = $this->minkContext->getSession()->getPage()->find('xpath', "//meta[@$arg1='$arg2' and @$arg3='$arg4']");
+        PHPUnit_Framework_Assert::assertNotNull($metaNode);
+        PHPUnit_Framework_Assert::assertEquals($arg2, $metaNode->getAttribute($arg1));
+        PHPUnit_Framework_Assert::assertEquals($arg4, $metaNode->getAttribute($arg3));
```
kencho51: Hi @rija 
I have updated the function using `xpath` which could hold two attributes at the same time. In fact, I tried the same way and put it into `('css', "meta[$arg1='$arg2' $arg3='$arg4']")`, but it could not work😢 

In this function, I have added 2 more parameters, `$arg1` and `$arg3`,  which may make it more reusable for future use. And I have add 2 more assertions to further confirm that the returned value equals to what we expected.

Is it redundant to add the extra parameters and assertions?

rija: Hi @kencho51,

When you work on a functionality you should always focus on making it working according to specifications first with the simplest code that is readable by the future you and others.
Then you can assess the state of the code and see if there's anything worth improving.

Remember the scenario steps have to be understood by anyone in the business. Acceptance tests is in effect a business contract between the product owner and the development team (hence the name),
so when writing the steps, never sacrifice conciseness and readability for re-usability. 

If you are looking for potential for re-use, always start thinking of a private function that can be used by multiple steps (you've already done that in a previous task), if opportunity is not there, leave it be.

So as it stands now,

```
And There should be a meta-tag has attribute "name" with value "citation_doi" and attribute "content" with value "10.5072/100002"
```

is not a good sentence for a business contract between the product owners and the development team.

Beside, look at the definition of a meta-tag in the HMTL specs, there is not a large number of attributes that ``<meta>`` elements can takes, so the effort of trying to make the step generic with  the attributes key as variables is not worth losing the readability and precision.

We can always change our mind later if and only if the situation calls for it.

To answer your question about the assertions, yes they are redundant because the fact that the xpath return a non-null object is enough proof since we specified an xpath for a <meta> with the given name and content. If the meta-tag is not there or didn't have the specified name AND content, the query would return a null object. Therefore it's enough to assert the query return something that's not null.

rija:
Hi @kencho51 

The scenario step for the new meta-tag and its implementation look good now.

two observations:

1. is there a reason you changed test DOI from 100002 to 100006 as the title being matched in the first ``Then`` step still pertain to the former
1. ``Then I should see "Genomic data from Adelie penguin (<em>Pygoscelis adeliae</em>)"`` has a problem:

The implementation of "I should see..." searches for matching text in the **rendered** page content, which means there won't be any HTML tags in the searched content, as they have already been **rendered**.

This is different from the new step you've implemented that must search for a match in the **source** of the page, before it is being rendered. 

Therefore the ``<em>...</em>`` HTML tags you've added to the first ``Then`` step will cause that step to fail. 
Humans don't see HTML tags when they look at a page in a web browser, and the "Then I should see..." steps are from the perspective of a human reading the page.

kencho51:
Hi @rija 

>1. is there a reason you changed test DOI from 100002 to 100006 as the title being matched in the first Then step still pertain to the former

It is because in my local gigaDB, the dataset table only has `100006` and its corresponding paper title is `Genomic data from Adelie penguin (<em>Pygoscelis adeliae</em>).`
image

>2. Then I should see "Genomic data from Adelie penguin (<em>Pygoscelis adeliae</em>)" has a problem

If the `<em></em>` was not included in the `I should see` step, it would throw the following error: 
```
The text "Genomic data from Adelie penguin (Pygoscelis adeliae). " was not found anywhere in the text of the current page. (Behat\Mink\Exception\ResponseTextException). 
```

features/dataset-metadata-citation-doi.feature
```diff
@@ -4,10 +4,10 @@ Feature: Add the metadata schema on dataset page to allow Hypothesis to parse ci
...
-  @ok @javascript @issue-85
+  @wip @javascript @issue-85
   Scenario: Go to dataset page and there is a meta-tag with DOI
     Given I am not logged in to Gigadb web site
-    When I go to "/dataset/100006"
-    Then I should see "Genomic data from Adelie penguin (<em>Pygoscelis adeliae</em>)"
-    And There should be a meta-tag with name "citation_doi" and content "10.5072/100006"
+    When I go to "/dataset/100094"
+    Then I should see "Data and software to accompany the paper: Applying compressed sensing to genome-wide association studies."
+    And There should be a meta-tag with name "citation_doi" and content "10.5072/100094"
```
kencho51, line 12: Hi @rija 
I found this `DOI` in the `dataset table` has title with no html tag, so better use this one for this scenario?

rija: Hi @kencho51,

It's better to run tests on a known state of database because the state of your local db may change over time as you've noticed. You may have nocticed that some of the acceptance tests have the following step has background, just for that purpose:
```
Background:
	Given Gigadb web site is loaded with "gigadb_testdata.pgdmp" data
```

That way, the tests can expect a stable and predictable state for the database

rija (approved):
Thanks, looks good now.

protected/views/dataset/view.php
```diff
@@ -36,6 +36,7 @@
                             <h4 class="left-border-title left-border-title-lg"><?= $mainSection->getHeadline()['title']; ?></h4>
                             <p class="dataset-release-date-text">Dataset type:  <?= $mainSection->getHeadline()['types'];?> <br> Data released on <?= $mainSection->getHeadline()['release_date'] ?></p>
                             <div class="color-background color-background-block dataset-color-background-block">
+                                <meta name="citation_doi" content="<?php echo $mainSection->getReleaseDetails()['full_doi']; ?>">
```
pli888, line 39: This `meta` tag appears somewhere within `<body>` in the HTML page. Since the `citation_doi` information is metadata within a `<meta>` tag, I think it is better located within the [`<head>`](https://www.w3schools.com/tags/tag_head.asp) section.

If we look at examples from other journals, e.g. this [one](https://www.cell.com/ajhg/fulltext/S0002-9297(17)30074-5) they also have the `meta` tag for `citation_doi` located in `<head>`.

protected/views/layouts/new_datasetpage.php
```diff
@@ -6,6 +6,8 @@
     <meta name="language" content="en" />
     <meta name="robots" content="noindex">
     <meta name="googlebot" content="noindex">
```
kencho51, line 8: Hi @pli888 and @rija 

Change the `noindex` to `index` here should fix the problem mentioned in issue #515. 
Reference can be found at [Google Search Central](https://developers.google.com/search/docs/advanced/crawling/special-tags).

protected/views/layouts/new_datasetpage.php
```diff
@@ -6,6 +6,8 @@
...
+    <meta name="citation_doi" content="<?php echo $mainSection->getReleaseDetails()['full_doi']; ?>">
+
```
kencho51, line 10: Hi @pli888 

I tried to added `full_doi` to here, which will make it appear inside `<head>`, but `$mainSection` becomes an undefined variable when calling it.

rija: Hi @kencho51, @pli888 

You can use the same approach as for the title for the HTML Page which specifified in \<head\>, therefore in the layout, but is specific to the dataset. (Look for ``pageTitle`` at the top of ``protected/views/dataset/view.php`` )

protected/views/layouts/new_datasetpage.php
```diff
@@ -6,7 +6,7 @@
     <meta name="language" content="en" />
     <meta name="robots" content="noindex">
     <meta name="googlebot" content="noindex">
-    <meta name="citation_doi" content="<?php echo $mainSection->getReleaseDetails()['full_doi']; ?>">
+    <meta name="citation_doi" content="<?php echo CHtml::encode($this->fullDOI); ?>">
```
kencho51, line 9: Hi @rija 

I followed `pageTitle` in `view.php` to return `full_doi` in meta tag in `new_datasetpage.php`. However, I got this error:
```
Error 500
Property "DatasetController.fullDOI" is not defined.
```

kencho51: Hi @rija 

I tried the following replacement in `/view.php`:
```
$this->pageTitle = $mainSection->getReleaseDetails()['full_doi'];
```

And call `$this->pageTitle` inside meta tag in `/new_datasetpage.php`
```
<meta name="citation_doi" content="<?php echo CHtml::encode($this->pageTitle); ?>" >
```
The `full doi` can be shown in the web page properly.
image

So, I think the problem is `$this->fullDOI` has not been properly defined. Do you know where should I look to about how to properly define a variable in view class?

rija: Hi @kencho51,

Regarding figuring out how ``pageTitle`` works.

Looking at the top of the ``DatasetController.php`` file, you have: 


```
class DatasetController extends Controller
```

If you suspect a class is doing something but you see no evidence of it in the code, check if it extends another class, in this case ``Controller.php``.

If the code in that class still doesn't explain things, look at its header:

```
class Controller extends CController
```

and drill down to its parent class (a.k.a its superclass)


If you're wondering how the thing you see there works, have a look at this documentation:


https://www.yiiframework.com/doc/guide/1.1/en/basics.component#defining-and-using-a-component-property

That documentation also tells you what you need to do to get your code working

kencho51: Hi @rija 

I added `fullDOI` function to `CController.php` and it is working properly now. And the `behat test` is also passing.

kencho51:
Hi @pli888 and @rija 

This PR is ready for final review now.

In this PR, 
- A DOI `meta tag` is added to the `<head>` in the HTML page, and `behat test` is passing.
- `fullDOI` function are created in `CController.php`
- Calling `fullDOI` from `view.php` and `new_datasetpage.php`

protected/views/layouts/new_datasetpage.php
kencho51: Hi @rija 
After looking at the `CController.php`, I tried to add `getFullDOI` and `setFullDOI` methods with reference to the `pageTitle`, this has been working and the `meta tag` was found inside `<head>` in HTML page.
But the `CController.php` is inside the `vendor` directory, so changing it is not a proper way?

pli888: @kencho51 I'm not entirely sure what you need to do here but if you want to retrieve or change the value of `pageTitle` then this can be done by calling the `setPageTitle()` and `getPageTitle()` functions in [`Controller`](https://www.yiiframework.com/doc/api/1.1/CController) from within the `DatasetController` class. As @rija says, this is because the `DatasetController` class extends `Controller` class which itself extends `CController` Therefore, `CController`shares its functions with`DatasetController` so this class can use them.

pli888: @kencho51 Looking at how to create the `<meta>` tag for `name="citation_doi"`, dataset web pages use the `new_column2` layout as defined in `actionView($id)` function in `DatasetController`. The `new_column2` layout itself uses `new_datasetpage` for rendering the `<head>` section and is provided with an array called `metaData` containing data it needs to use. You can add more items into the `metaData` array in `DatasetController` and then this can be accessed in `new_datasetpage.php`.

kencho51:
Hi @pli888 and @rija 
This PR is ready for final review now.

In this PR, 
1. A variable `fullDOI` is added to the `metaData` array in `Controller.php`. 
2. Call `fullDOI` from `new_datasetpage.php` `<head>` section, the DOI `<meta>` is shown properly in HTML page.
3. `Acceptance test` is passing.

protected/views/layouts/new_datasetpage.php
pli888: >Change the noindex to index here should fix the problem mentioned in issue #515.

@kencho51 This still needs to be done.

features/bootstrap/DatasetViewContext.php
```diff
@@ -416,7 +416,12 @@ public function iHaveSetPagesizeToOn($arg1, $arg2)
         }
     }
 
-
-
-
+    /**
+     * @Then There should be a meta-tag with name :arg1 and content :arg2
+     */
+    public function thereShouldBeMetaTagWithNameAndWithContent($arg1, $arg2)
```
pli888, line 422: Maybe this can be shortened, e.g. `@Then the page meta :arg1 should be :arg2`?

kencho51: Change it to `There is a meta tag :arg1  with DOI :arg2`.

pli888 (changes requested):
The `css/current.css` file should not part of this pull request and needs to be deleted. It should not be committed as it is created new every time the `webapp` service is started using `webapp_setup.sh`.

features/bootstrap/DatasetViewContext.php
```diff
@@ -416,7 +416,12 @@ public function iHaveSetPagesizeToOn($arg1, $arg2)
         }
     }
 
-
-
-
+    /**
+     * @Then There is a meta tag :arg1 with value :arg2
+     */
+    public function thereIsAMetaTagWithDOI($arg1, $arg2)
```
pli888, line 422: The name of this function looks strange and should be changed to correspond with the `@then` feature step. What if the content of the `<meta>` tag doesn't contain a DOI? For example, your `@ok @issue-85 @issue-515` behat test contains this line:
```
And There is a meta tag "robots" with value "index"
```

features/bootstrap/DatasetViewContext.php
```diff
@@ -416,7 +416,12 @@ public function iHaveSetPagesizeToOn($arg1, $arg2)
         }
     }
 
-
-
-
+    /**
+     * @Then There is a meta tag :arg1 with value :arg2
```
pli888, line 420: `@Then There is` should `@Then there is`.

features/dataset-metadata.feature
```diff
@@ -0,0 +1,35 @@
+@issue-85 @dataset-metadata-citation-doi @issue-515
+Feature: Add the metadata schema on dataset page to allow Hypothesis to parse citation DOI
+  As a researcher
+  I want the Hypothesis annotation tool to recognise a GigaDB dataset page url as an alias of its DOI
+  I want all search engines, including google to recognise all GigaDB pages url
+  So that I can correctly cite it irrespective of which website it appears on or I annotate from
+  So that all search engines can crawl and index GigaDB pages
```
pli888, line 7: This feature looks strange with two `I want` and two `So that` lines. I think [features](https://behat.org/en/latest/user_guide/features_scenarios.html#user-guide-features-scenarios-features) should only contain three indented lines.

I'm not sure what is best solution. In issue #85, @rija suggests having ` features/dataset-metadata-citation-doi.feature` file for the Hypothesis acceptance. Maybe you could have a ` features/dataset-metadata-search-engine.feature` file too as I can see more tests being added to such a feature file in the future.

kencho51: A new feature file `features/dataset-metadata-search-engine.feature` is created for search engine features.

protected/views/layouts/new_datasetpage.php
```diff
@@ -4,8 +4,10 @@
 <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <meta name="language" content="en" />
-    <meta name="robots" content="noindex">
-    <meta name="googlebot" content="noindex">
+    <meta name="robots" content="index">
+    <meta name="googlebot" content="index">
```
pli888, line 8: This will need updating based on [comment](https://github.com/gigascience/gigadb-website/issues/515#issuecomment-740445381) by @only1chunts. Also need to repeat for layout classes below.

only1chunts: is the googlebot tag actually needed? I think google bots adhere to the "robots" meta tags so unless we want google to treat us differently to other bots (which we dont) then we shouldn't need to explicitly name googlebot.
Please confirm this is correct as I'm no expert in these matters! but I think we should only need:
<meta name="robots" content="index, nofollow"/>

kencho51: Hi @only1chunts , from [here](https://developers.google.com/search/docs/advanced/crawling/special-tags) it says:
```
The <meta name="robots"> tag applies to all search engines, 
while the <meta name="googlebot"> tag is specific to Google.
```
`robots` tag has included Google already, so I think `googlebot` could be removed.

pli888 (approved):
~~I think this PR all looks good now.~~

Actually looking at PR #547 made me think about this PR, specifically this function in `DatasetViewContext` class:
```
/**
     * @Then there is a meta tag :arg1 with values :arg2 and :arg3
     */
    public function thereIsMetaTagWithValues($arg1, $arg2, $arg3)
```

This `thereIsMetaTagWithValues` function takes two `@content` arguments. However, any scenario step that uses this function can also use the `thereIsAMetaTagWithValue($arg1, $arg2)` function since the value of `content` is just a string.

features/bootstrap/DatasetViewContext.php
```diff
@@ -416,7 +416,21 @@ public function iHaveSetPagesizeToOn($arg1, $arg2)
...
-
-
+    /**
+     * @Then there is a meta tag :arg1 with values :arg2 and :arg3
+     */
+    public function thereIsMetaTagWithValues($arg1, $arg2, $arg3)
+    {
+        $metaNode = $this->minkContext->getSession()->getPage()->find('xpath', "//meta[@name='$arg1' and @content='$arg2, $arg3']");
+        PHPUnit_Framework_Assert::assertNotNull($metaNode);
+    }
```
pli888, line 435: Maybe this `public function thereIsMetaTagWithValues($arg1, $arg2, $arg3)` function is not required? Scenario steps that use this function, e.g. `Then There is a meta tag "robots" with values "index" and "nofollow"` can use the function above `public function thereIsAMetaTagWithValue($arg1, $arg2)` and be re-written as `Then There is a meta tag "robots" with value "index, nofollow"`.

kencho51: This step has been replaced and the behat test passed.

features/dataset-metadata-citation-doi.feature
```diff
@@ -0,0 +1,15 @@
...
+
+  Background:
+    Given Gigadb web site is loaded with "gigadb_testdata.pgdmp" data
+
+  @ok @issue-85
+  Scenario: Go to dataset page, Hypothesis tag is found
+    Given I am not logged in to Gigadb web site
+    When I go to "/dataset/100002"
+    Then I should see "Genomic data from Adelie penguin (Pygoscelis adeliae)."
+    And There is a meta tag "citation_doi" with value "10.5072/100002"
```
pli888, line 15: `There` should `there` on this line. Also correct the same typo in `features/dataset-metadata-search-engine.feature` and then I think this PR can be merged.

kencho51: The `feature` files updated.

Issue #498, opened 2020-08-25
Understanding styling and layout in GigaDB


pli888:
There are two layouts being used in GigaDB:

| Current green layout | Previous grey layout |
|----------------------|----------------------|
| ![current_gigadb](https://user-images.githubusercontent.com/2101491/91116128-3eccb800-e6be-11ea-9248-6ccac93b9186.png) | ![old_gigadb](https://user-images.githubusercontent.com/2101491/91116136-442a0280-e6be-11ea-9f78-23033a24a6e4.png) |

These layouts are being generated by CSS files (some via less files) and Yii view classes:

## CSS files

| CSS file | Version | Referenced In | Function |
|---|---|---|---|
| jquery-ui.css | 1.12.1 | new_main.php | Current green layout |
| bootstrap.min.css | 3.3.7 | new_main.php | Current green layout |
| font-awesome.min.css | 4.7.0 | new_main.php | Current green layout |
| open_sans.css | 13 | new_main.php | Current green layout |
| pt_sans.css | 8 | new_main.php | Current green layout |
| lato.css | 11 | new_main.php | Current green layout |
| datatables.css | Not known | new_dataset.php | Current green layout |
| common.css | Not known | views/layouts/new_datasetpage.php, views/layouts/new_faq.php, views/layouts/new_main.php | Current green layout |
| site.css* | Not known | views/layouts/datasetpage.php, views/layouts/main.php | Previous grey layout |
| chosen.min.css | 1.3.0 | views/report/index.php | For displaying Google Analytics data |
| jquery.tag-editor.css | Not known | datasetSubmission/create1.php, adminDataset/_form.php, datasetSubmission/_form1.php | For editing tags in WL submission wizard |
| pager.css | 1.0? | SiteLinkPager.php |	For pagination in dataset pages |

*This file is created from site.less using lesscompilercommand in webapp_setup.sh script which is executed by `docker-compose.yml` and `docker-compose.staging.yml`.

## Less files

| Less file | Function |
|---|---|
| bootstrap/*.less | Less files for Bootstrap (3?) |
| about.less | About page? |
| admindataset.less | Dataset pages? |
| datasetupdate.less | Dataset update pages? |
| globals.less | General variables, e.g. font type, button colours |
| pagesetting.less | Not known |
| site.less | General site CSS rules |
| thumbnailSlider.less | CSS rules for carousel in old grey layout |
| updatelog.less | CSS rules for dataset log pages? |
[...]

PR #520, opened 2020-10-29
Less files for creating the current GigaDB look and layout

pli888:
# Pull request for issue: #498

* A [pattern library](https://gigadb-staging.gigatools.net/style-guide/) showing a list of components used by the GigaDB website. The pattern library has been implemented using code from https://github.com/bjankord/Style-Guide-Boilerplate which has been placed into the `style-guide` directory and is available at http://gigadb.gigasciencejournal.com:9170/style-guide/index.php on a local deployment.
* A set of less files which can be compiled into a current.css file to provide the styling and layout for the current green GigaDB website. These less files are not the finished product but are "good enough" to allow development of functionality requiring new CSS rules such as the WL submission wizard.

## Changes to the style

A number of colours are seldom used in GigaDB. These have been deleted to reduce the 40 plus colours that are referenced in `common.css`. CSS rules involving the use of these colours now use another very similar shade of colour.

## Changes to the tests

Two new bash scripts can be found in `tests` directory, `css_check` and `javascript_check` for running error checks on Javascript files and CSS files. These 2 scripts are executed in `all_and_coverage` via `docker-compose run --rm test`.

CSS errors in `common.css` are detected using [CSSLint](https://github.com/CSSLint/csslint). Checks for Javascript errors are performed by [JSHint](https://jshint.com/docs/) and is only done for `style-guide/js/sg-scripts.js` - other javascript files such as those from Bootstrap are assumed to be error-free. If any CSS or Javascript errors are detected then they are outputted into `tmp/css_errors.txt` and `tmp/js_errors.txt`.  

A set of new `less` files can be found in `less/current` directory. These files have been created by separating CSS rules in `common.css` based on SMACSS principles. The `current.less` file is responsible for importing `current` less files when compiling into a new `current.css`. This new step has been added into `LessCompilerCommand.php` which is executed as part of `docker-compose run --rm webapp`.

The `style-guide` directory has been added as a new volume into the `web` service in `docker-compose.yml` so that it is available from the `nginx` web server container.

Checking CSS and javascript files for errors involves using CSSLint and JSHint, respectively. Steps for these tools to be installed have been added into `Dockerfile`.


rija (approved):
This looks great. Implementation looks fine to me.

some minor  observations/questions:

* In the style guide, the doc for "fixed-btn-left" appears funny (title's blown out) on my screen with all three browsers, everything else is fine though
* What is the purpose of ``nodesource_setup.sh`` ? 

In the future, we probably should standardise the way we use NodeJS. I prefer the approach we took for the "csv-to-migrations" container service (which is the approach I use to run javascript tests on my branch). But let's not make these changes here, as usability wise, they make sense only with the inside-out test-runners that come with my branch.

pli888:
>In the style guide, the doc for "fixed-btn-left" appears funny (title's blown out) on my screen with all three browsers, everything else is fine though

From what I understand, that `fixed-btn-left` button appears way over to the left of the page because of `margin-left: -630px;` in its CSS rule. Its font size is 60px in the CSS rule so this is why the title looks big. I opted to leave the `fixed-btn-left` button as it is for the time being since that is how its CSS rule is specified and used at the moment in GigaDB.

>What is the purpose of nodesource_setup.sh?

IIRC, its for installing a more recent version of nodes which is required by CSSLint in `Dockerfile`:
```
curl -sL https://deb.nodesource.com/setup_10.x -o nodesource_setup.sh && \
    bash nodesource_setup.sh && \
    apt-get install -y nodejs && \
    rm nodesource_setup.sh && \
    npm install -g csslint jshint \
```

Issue #527, opened 2020-11-09
Application docker image fails to build


pli888:
@kencho51 found that the command `$ docker-compose run --rm webapp` fails to start the `webapp` docker service defined in `docker-compose.yml`. If the `application` image defined in `docker-compose.prof.yml` needs to be built then it will fail to complete its build and produces the error message below:

![image](https://user-images.githubusercontent.com/2101491/98491515-0e81c800-2270-11eb-827b-83135849975a.png)

The problem seems to be caused by `git` not being available when building the image.

PR #528, opened 2020-11-09
Fix git not found error in Profiler installation step

pli888:
# Pull request for issue: #527

* Fix git not found error when installing Profiler in `Dockerfile`. This results in `docker-compose run --rm webapp` failing and the webapp docker service defined in docker-compose.yml not being able to start since `application` image cannot complete its build.

`git` is installed as part of the Tideways Profiler installation in `Dockerfile`.


kencho51 (approved):
After installing the `git`, the `webapp` service can be rebuilt in my branch now.

PR #532, opened 2020-11-21
Use database migrations to provide data for acceptance tests

pli888:
# Pull request for issue: #533

* Use Yii database migration scripts for setting up database to provide data for running acceptance tests

## Changes to the tests

Perhaps the most significant change is the one made to the `gigadbWebSiteIsLoadedWithData`  function in `features/bootstrap/GigadbWebsiteContext.php`. It has been updated so that it executes Yii migration scripts to create database schema and upload data into the tables now:
```
$schema_output = shell_exec("./protected/yiic migrate --migrationPath=application.migrations.schema --interactive=0 2>&1");
if(!strpos($schema_output, "Migrated up successfully.")) {
    echo "$schema_output";
    exit("Problem with running database schema migration scripts\n");
}
$data_output = shell_exec("./protected/yiic migrate --migrationPath=application.migrations.data.${arg1} --interactive=0 2>&1");
if(!strpos($data_output, "Migrated up successfully.")) {
    echo "$data_output";
    exit("Problem with running database data migration scripts\n");
}
```

After each `shell_exec`, the output is checked for any errors which will then be displayed in STDOUT.

The function `gigadbWebSiteIsLoadedWithProductionLikeData` in `GigadbWebsiteContext.php` and any `.feature` files using `production_like.pgdmp` and `gigadb_testdata.pgdmp` to refer to the corresponding `pgdmp` database backup file have been updated.  The `.pgdmp` file extension has been removed so they refer to directories in `data/` and `protected/migrations` now.

CSV files representing the data in `production_like.pgdmp` and `gigadb_testdata.pgdmp` have been added into the `data` directory.

These CSV files have been used to create the Yii migration scripts in `protected/migrations/production_like` and `protected/migrations/gigadb_testdata` using `docker-compose up csv-to-migrations` command. I have decided to include these Yii migration scripts as I don't expect them to change so much and they do not have to be made on the fly when tests are being setup for execution in GitLab CI/CD pipeline.

It should be noted that the `production_like` data consists of 21 datasets only. In `production_like.pgdmp`, there are 319 datasets in total. The problem is that if 319 datasets are uploaded into a test database using Yii migration scripts then we run into memory problems:
```
Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 266240 bytes) in /var/www/protected/migrations/data/production_like/m200529_050310_insert_data_file_tab.php on line 283324
```

## Changes to the documentation

`data/TESTING.md` has been updated so that it now describes how the CSV files and Yii migration scripts have been generated.

### DIFF
--- a/ops/packaging/Dockerfile
+++ b/ops/packaging/Dockerfile
@@ -147,6 +147,7 @@ RUN if [ ${INSTALL_TIDEWAYS_XHPROF} = true ]; then \
     cd /usr/local/src/ && \
     tar xzvf /usr/local/src/xhgui.tar.gz && \
     cd xhgui-0.8.1 && \
+    apt-get install -y git && \
     composer install && \
     echo "auto_prepend_file=/usr/local/src/xhgui-0.8.1/external/header.php" > /usr/local/etc/php/conf.d/xhgui.ini \
 ;fi


rija (commented):
Hi @pli888,

Good thinking leveraging the database migrations for this purpose.

However even though those test data do not change often, the migrations are still duplication of the CSVs files and the original motivation was for CSVs being the source to be versioned from which other representations are derived on deployments ; also  that approach creates potential  inconsistencies in the ``protected/migrations/data`` directory where some data are gitignored while other are versioned.

What I like about your approach though, is that you do away with the ``*.pgdmp`` altogether.
However, it assumes the binary dump were only used for acceptance testing.

But searching through the codebase, I see it's used to bootstrap the functional tests as well.
Although I've rewritten the test runners in my branch, I haven't removed that dependency and there might be other situations were we want to quickly load (probably programmatically) a known canned set of data from a dump without going through generating the migrations.

I took a different approach in my branch with the same starting point (the CSVs for the test data, except I called it ``data/test`` instead of ``data/gigadb_testdata`` but either naming is fine):

I have the following  shell script at ``ops/scripts/setup_testdb.sh`` to run from project's root:
```
#!/usr/bin/env bash

# create database if not existing
docker-compose run --rm test bash -c "psql -h database -U gigadb -c 'create database gigadb_test'" || true

# generate migrations 
docker run -v `pwd`:/var/www node:14.9.0-buster bash -c "node /var/www/ops/scripts/csv_yii_migration.js test"

# and run them
docker-compose run --rm  application ./protected/yiic migrate to 300000_000000 --connectionID=testdb --migrationPath=application.migrations.admin --interactive=0
docker-compose run --rm  application ./protected/yiic migrate mark 000000_000000 --connectionID=testdb --interactive=0
docker-compose run --rm  application ./protected/yiic migrate --connectionID=testdb --migrationPath=application.migrations.schema --interactive=0
docker-compose run --rm  application ./protected/yiic migrate --connectionID=testdb --migrationPath=application.migrations.data.test --interactive=0

# export a binary dump
docker-compose run --rm test bash -c "pg_dump --no-owner -U gigadb -h database -p 5432 -F custom -d gigadb_test -f /var/www/sql/gigadb_testdata.pgdmp"
```

I have also almost identical script that does the same thing for resetting the default dev database with the test data (probably could refactor into one parameterisable script as improvement)

The aim of these scripts is to easily populate the default and test database as well as generate the binary dump as a side effect.

Downside is it's an additional step to perform at checkout/deployment, but I guess it could be part of one startup script. Also I haven't done it, but we could generate the production-like database dump in a similar way.

Notice the ``--connectionID`` argument to Yii migration script. 
This is how to tell Yii on which database to run the migration against (default to 'db' if not present).
My ``protected/config/console.php``'s ``components`` section  looks like so at the moment:

```
       'db'=>array(
            'class'=>'system.db.CDbConnection',
            'connectionString'=>"pgsql:dbname={$dbConfig['database']};host={$dbConfig['host']}",
            'username'=>$dbConfig['user'],
            'password'=>$dbConfig['password'],
            'charset'=>'utf8',
            'persistent'=>true,
            'enableParamLogging'=>true,
            'schemaCachingDuration'=>30
        ),
        'testdb'=>array(
            'class'=>'system.db.CDbConnection',
            'connectionString'=>"pgsql:dbname=gigadb_test;host={$dbConfig['host']}",
            'username'=>$dbConfig['user'],
            'password'=>$dbConfig['password'],
            'charset'=>'utf8',
            'persistent'=>true,
            'enableParamLogging'=>true,
            'schemaCachingDuration'=>30
        ),     
```

Issue #537, opened 2020-12-04
504 Gateway Timeout errors on dataset webpages


pli888:
If a dataset has associated publications, then it's webpage might take a long time to load. This is because the web page is waiting for citation information about publications before it is displayed.

For example, a GET request to `http://dx.doi.org/10.1186/2047-217X-1-14` is made by http://gigadb.org/dataset/100039. If this dataset page is loaded on a local deployment using http://gigadb.gigasciencejournal.com:9170/dataset/100039 then it can result in an error message:
```
Server error: `GET http://dx.doi.org/10.1186/2047-217X-1-14` resulted in a `504 Gateway Timeout` response
```

Because of this problem, acceptance tests involving dataset pages such as `@author-names-display` can fail as expectations do not end up being displayed if the page is not loaded in time. For example, see [here](https://gitlab.com/gigascience/forks/kencho51-gigadb-website/-/jobs/890122509).

The GET request to http://dx.doi.org/10.1186/2047-217X-1-14 is made in getPublications() in StoredDatasetConnections class.

Perhaps this can be made as a AJAX call to fix this problem?

only1chunts:
not sure if it helps, but there might also be an option to pull the information directly from the CrossRef API using the DOI details.
https://www.crossref.org/education/retrieve-metadata/
that might reduce the reliance of the DOI resolver AND the GigaScience website both needing to be working

rija:
Hi @pli888

I run into same issue on my branch a while ago.
I thought i fixed it, but I've just tagged the impacted acceptance tests and skip them as it was a distraction from my main work.

Before that, I thought the issue was that we are not using the resolver correctly, because:

*  the DX endpoint is deprecated 
* and we're passing a Request header that seems no longer supported ('style' => 'apa')
* we are not using HTTPS to make the query 

but fixing the above didn't stop the timeout to happen.

We cannot use AJAX on the frontend as that information we're fetching is core data needed by model classes and the DOI resolver is considered a remote data source and therefore its client implemented in the model layer's ``Stored*`` PHP sub-layer.


The correct approach is to insulate that sub-layer from issues in the underlying data source - actually we should do that for all remote web services and database servers we query  -, to prevent "cascading failures"  by installing a [Circuit Breaker](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern).

[...]

PR #538, opened 2020-12-10
Fix gateway timeout problem when retrieving publication information

pli888:
# Pull request for issue: #537

* Update endpoint and use HTTPS for retrieving publication information

Acceptance tests involving the checking of content in dataset pages were failing because the pages were taking too long to load. This resulted in 504 Gateway Timeout errors that were caused when trying to retrieve citation information using a DOI with the `dx.doi.org` endpoint. Updating to the preferred `https://doi.org` endpoint appears to fix the timeout problem when running tests on a local deployment and on CI/CD.

We can investigate this problem in more detail if it arises in the future.

## Changes to the components classes

The `getPublications` function in `StoredDatasetConnections.php` has been updated so that it now uses the `https://doi.org/` endpoint for retrieving citation information. Previously, it used `http://dx.doi.org/` which is still [supported](https://www.doi.org/factsheets/DOIProxy.html) but `https://doi.org/` is now preferred. 

## Changes to the tests

The `StoredDatasetConnectionsTest` unit test has been updated to use the `https://doi.org/` endpoint in its tests.

### DIFF
--- a/protected/components/StoredDatasetConnections.php
+++ b/protected/components/StoredDatasetConnections.php
@@ -96,7 +96,7 @@ class StoredDatasetConnections extends DatasetComponents implements DatasetConne
 		foreach ($rows as $result) {
 			$response = null;
 			try {
-				$response = $this->_web->request('GET', 'http://dx.doi.org/'. $result['identifier'], [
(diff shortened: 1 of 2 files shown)


rija (approved):
Looks good to me.

Issue #513, opened 2020-10-06
Add the metadata schema on dataset page to allow other web sites to make link previews of our web site


only1chunts:
## User Story

```
As an operator of a partner website
I want to extract preview metadata from links to GigaDB datasets
So I can present preview information to my visitors interested in those links
```

### example from  metatags.io

```

      <!-- Primary Meta Tags -->
      <title>Meta Tags — Preview, Edit and Generate</title>
      <meta name="title" content="Meta Tags — Preview, Edit and Generate">
      <meta name="description" content="With Meta Tags you can edit and experiment with your content then preview how your webpage will look on Google, Facebook, Twitter and more!">

      <!-- Open Graph / Facebook -->
      <meta property="og:type" content="website">
      <meta property="og:url" content="https://metatags.io/">
      <meta property="og:title" content="Meta Tags — Preview, Edit and Generate">
      <meta property="og:description" content="With Meta Tags you can edit and experiment with your content then preview how your webpage will look on Google, Facebook, Twitter and more!">
      <meta property="og:image" content="https://metatags.io/assets/meta-tags-16a33a6a8531e519cc0936fbba0ad904e52d35f34a46c97a2c9f6f7dd7d336f2.png">

      <!-- Twitter -->
      <meta property="twitter:card" content="summary_large_image">
      <meta property="twitter:url" content="https://metatags.io/">
      <meta property="twitter:title" content="Meta Tags — Preview, Edit and Generate">
      <meta property="twitter:description" content="With Meta Tags you can edit and experiment with your content then preview how your webpage will look on Google, Facebook, Twitter and more!">
      <meta property="twitter:image" content="https://metatags.io/assets/meta-tags-16a33a6a8531e519cc0936fbba0ad904e52d35f34a46c97a2c9f6f7dd7d336f2.png">
```

### example from cell.com

```
<meta property="og:title" content="Mutations in TMEM260 Cause a Pediatric Neurodevelopmental, Cardiac, and Renal Syndrome" />
    <meta property="og:type" content="Article" />
    <meta property="og:url" content="https://www.cell.com/ajhg/abstract/S0002-9297(17)30074-5" />
    <meta property="og:site_name" content="The American Journal of Human Genetics" />
    <meta property="og:image" content="//els-jbs-prod-cdn.jbs.elsevierhealth.com/cms/attachment/2089506234/2075402522/gr2.jpg" />
[...]

Issue #543, opened 2020-12-15
WL testing - lots of carriage returns in description


only1chunts:
**Describe the bug**
Maybe not a bug, just a point of discussion.

The description added in the submission wizard should be processed to replace all carriage returns with html encoded line breaks. 
The reason is that we want to avoid carriage returns in the database because when anything is exported it makes a mess and they are difficult to parse.

This should be considered open to discussion, as the issue with line breaks in the database is one that probably shouldn't be an issue with the correct tools! It stems from the curators need to export certain database tables to OpenRefine for refining the contents, any values containing a line break make that export very messy.
I cannot think of a scenario where we would need to export the "Dataset" table to OpenRefine so it may not be so important to try to keep consistency over all the tables about the line breaks.

**To Reproduce**
See the screenshot of the Admin view of the dataset I just submitted via the submission wizard:
![image](https://user-images.githubusercontent.com/6037145/102236480-79679400-3eeb-11eb-8c8e-1b06fd987862.png)

**Expected behavior**
When Saved the description should be converted to a linear single line format for storage in the database.

**Additional context**
I think it's impractical to allow users to add Tables to the dataset description so that functionality can be removed from the editor tool.

PR #547, opened 2020-12-22
Add the metadata schema to allow other website link preview #513

kencho51:
# Pull request for issue: #513

1. Store `title` and `description` to `metaData`  array in DatasetController.php

1. Add meta tags to `new_dataset.php`

## Progress

- [x] Store title and description to metaData  array 
- [x] Primary meta-tags
- [x] Open Graph meta-tags
- [x] Twitter meta-tags
- [x] HTML meta-tags
- [x] Pass behat test


features/dataset-metadata-link-preview.feature
```diff
@@ -0,0 +1,18 @@
...
...surrounding islands. Being very sensitive to climate change, and due to changes in their behavior based on minor shifts in climate, they are often used as a barometer of the Antarctic.With its status as one of the adorable and cuddly flightless birds of Antarctica, they serve as an example for conservation, and as a result they are now categorised at low risk for endangerment. The sequence of the penguin can be of use in understanding the genetic underpinnings of its evolutionary traits and adaptation to its extreme environment; its unique system of feathers; its prowess as a diver; and its sensitivity to climate change. We hope that this genome data will further our understanding of one of the most remarkable creatures to waddle the planet Earth.We sequenced the genome of an adult male from Inexpressible Island, Ross Sea, Antartica (provided by David Lambert) to a depth of approximately 60X with short reads from a series of libraries with various insert sizes (200bp- 20kb). The assembled scaffolds of high quality sequences total 1.23 Gb, with the contig and scaffold N50 values of 19 kb and 5 mb respectively. We identified 15,270 protein-coding genes with a mean length of 21.3 kb."
```
kencho51, line 16: Hi @rija ,

Can you have a quick look at this step for me?
This step is failing and the message is `Failed asserting that null is not null.`, and the debug output does return `empty`. But, the previous step to check the `title` with the same implementation is passing.

kencho51: I found the problem, the `description` of  dataset `100002` is too long and separated in 3 lines, so the `metaNode` could not find its content.
I then use dataset `100004` which has shorter description with 1 line only, the tests pass now.

only1chunts: Hi Ken, 
There is an issue with line breaks in the descriptions. We try to not include them, instead using "<br>" so that the descriptions render correctly in the website, but we sometimes miss some. There is also a ticket #543 that mentions this issue particularly in the WL work Peter has been doing. If there are better ways to handle the description field with all its formatting then we should explore them?!

kencho51:
Hi @rija and @pli888 ,

This PR is ready for final review.

In this PR, 
1. 4 variables are created for `$metaData` array.
```php
'title', 'description', 'full_doi' , 'image_url'
```

2. Store `full doi`, `dataset title`, `description`, `image url` into `$metaData` array.

3. Implement meta tags for `HTML`, `Twitter/OGP` and `OGP`. 

4. Pass the `behat test` by including the `attribute` variable, which will make the test more flexible, so it can test for both `name` and `property` attributes of meta tag.

5. Specify `xdebug` version in `Dockerfile` to solve the `docker build` error.

features/bootstrap/DatasetViewContext.php
```diff
@@ -416,7 +416,13 @@ public function iHaveSetPagesizeToOn($arg1, $arg2)
         }
     }
 
-
-
+    /**
+     * @Then there is a :arg1 meta tag :arg2 with value :arg3
+     */
+    public function thereIsAMetaTagWithValue($arg1, $arg2, $arg3)
```
pli888, line 422: PR #516 has functions with similar names: `thereIsAMetaTagWithValue($arg1, $arg2)` and `thereIsMetaTagWithValues($arg1, $arg2, $arg3)` which will conflict with the above function. I think its worth having a think about how to name these functions that do similar things.

features/dataset-metadata-link-preview.feature
```diff
@@ -0,0 +1,33 @@
...
+    And there is a "name" meta tag "description" with value "The aim of a genome-wide association study (GWAS) is to isolate DNA markers for variants affecting phenotypes of interest. Linear regression is employed for this purpose, and in recent years a signal-processing paradigm known as compressed sensing (CS) has coalesced around a particular class of regression techniques. CS is not a method in its own right, but rather a body of theory regarding signal recovery when the number of predictor variables (i.e., genotyped markers) exceeds the sample size. The paper shows the applicability of compressed sensing (CS) theory to genome-wide association studies (GWAS), where the purpose is to ﬁnd trait-associated tagging markers (genetic variants). Analysis scripts are contained in the compressed CS file. Mock data and scripts are found in the compressed GD file. The example scripts found in the CS repository require the GD files to be unpacked in a separate folder. Please look at accompanying readme pdfs for both repositories and annotations in the example scripts before using."
```
pli888, line 15: Have a think about using [placeholders and tables](https://behat.org/en/latest/user_guide/writing_scenarios.html) for writing these test scenarios. It should make the scenario appear neater.

features/bootstrap/DatasetViewContext.php
```diff
@@ -431,12 +431,13 @@ public function thereShouldBeAMetaTagWithValue($arg1, $arg2, $arg3)
...
 
         $metaNode = $this->minkContext->getSession()->getPage()->find('xpath', "//meta[@$arg1='$arg2']");
         $actualContent = $metaNode->getAttribute('content');
-//        file_put_contents('test_expect.txt', print_r($expectContent, true));
+        $actualContent = preg_replace('/\s\s/', ' ', implode(" ", preg_split('/\r\n|\r|\n/', $actualContent)));
 
-        PHPUnit_Framework_Assert::assertEquals(implode(" ", preg_split('/\r\n|\r|\n/', $expectContent)), implode(" ", preg_split('/\r\n|\r|\n/', $actualContent)));
+        PHPUnit_Framework_Assert::assertEquals($expectContent, $actualContent, 'The content is different!');
 
     }
```
kencho51: Hi @rija ,

I used `PyStringNode` to allow input of multiple lines, then clean the `expectContent` and `actualContent` by removing the `carriage return` and `extra space`, finally `assertEquals` both of them. 
The test is passing.
What do you think about this implementation? I feel that there may be too many manipulation/operation 😅

rija (changes requested):
Hi @kencho51

It looks good, I have two observations:

## (1) Regarding the multi-lines issues: 

The use of pyStrings (that's what I called triple-quoted strings on Monday), is a good idea. However the implementation is a bit convoluted (especially if you knew ``getRaw()`` internally uses ``implode()`` on an array of strings). The code below will work too and is more succinct:

```
public function thereShouldBeAMetaTagWithMultiplines($arg1, $arg2, \Behat\Gherkin\Node\PyStringNode $expectedValue )
    {
        $actualNode = $this->minkContext->getSession()->getPage()->find('xpath', "//meta[@$arg1='$arg2']");
        [$expectContent, $actualContent] = str_replace(["\r","\n","\r\n","\t","\v","\0"," "], "", [$expectedValue->getRaw(), $actualNode->getAttribute('content')]);
        PHPUnit_Framework_Assert::assertEquals($expectContent, $actualContent, 'The content is different!');

    }
```

It's also more robust/indulgent as it takes care of additional idiosyncratic termination markers like vertical tabs,  NULL bytes and of the issue of varying number of spaces.

The succinctness relies on ``str_replace()`` having the ability to trim characters in an array of strings, not just in a string.

>**Notes**:

 * When the use case can be implemented with both ``str_replace()`` and ``preg_replace()``, it's better to use the former for performance reason.
* use ``$expectedValue`` and ``$actualNode``, not ``$arg3`` and ``$metaNode``, it makes the code so much more readable.


## (2) The ``@issue-513`` tag doesn't need to be on every scenarios AND on the feature:

Placing a tag above a ``Feature`` block means that you want the tag to apply to all scenarios, so it's redundant to also add the tag above the ``Scenario`` blocks.

You will sometimes see in the feature files specific scenarios having an issue related tag. This is because it's usually a bug and the scenario was added specifically to fix that issue.

rija (approved):
Thanks, looks good

Issue #533, opened 2020-11-23
Issues with using pgdmp files as data for running tests


pli888:
It is problematic to add new test data into the test harness for the GigaDB application because of the use of PostgreSQL `pgdmp` files. In order for new data to be added into pgdmp files, one has to load data from `gigadb_testdata.pgdmp` or `production_like.pgdmp` files into a database. New data are then inserted into the relevant tables and then a new `pgdmp` file is created for use by the Behat tests. Writing new acceptance tests will likely require this process to be repeated which makes it tedious to perform.

It would be more convenient if test data can be provided in the form of CSV files which are transformed into Yii database migration scripts that are executed prior to each test being executed in a process that is already in place for deploying a database for development work - see #469. With CSV files, we can easily see what data is contained in database table rows. Any changes to the test data can be made directly to the CSV files which are loaded into a test database as Yii migration files for tests to run against.

PR #548, opened 2021-01-08
Creating pgdmp files from CSV files to provide data for running tests

pli888:
# Pull request for issue: #533

This is a new pull request based on rija's comments in #532 for the following functionalities:

* A set of CSV files for creating `gigadb_test.pgdmp`
* A set of CSV files for creating `production_like.pgdmp`
*  Shell scripts for creating above `pgdmp` files

## Changes to the configuration

`ops/configuration/yii-conf/console.php.dist` has been updated with properties to allow access to the `gigadb_testdata` and `production_like` database when running Yii database migrations.

CSV files containing the data in `production_like.pgdmp` and `gigadb_testdata.pgdmp` have been added into the `data` directory. Based on Rija's suggestion, two shell scripts `setup_testdb_gigadb_testdata.sh` and `setup_testdb_production_like.sh` have been added to automate the creation of the `pgdmp` files. 

`setup_testdb_gigadb_testdata.sh` converts the `gigadb_testdata` CSV files into Yii migration scripts which are then used to upload data into the `gigadb_testdata` database. However, data for the`production_like` database are loaded directly from CSV files in `setup_testdb_production_like.sh` using `psql \copy` commands. This is because loading all 319 datasets into the `production_like` database via Yii migration scripts leads to memory errors:
```
    > insert into file_type ... done (time: 0.001s)
    > insert into file_type ... done (time: 0.001s)
*** applied m200529_050300_insert_data_file_type_tab (time: 0.098s)

*** applying m200529_050310_insert_data_file_tab
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 266240 bytes) in /var/www/protected/migrations/data/production_like/m200529_050310_insert_data_file_tab.php on line 291790
```

In addition, the `setup_testdb_production_like.sh` script retrieves the GigaDB database password from `.secrets` so the user does not have to enter it on the command line when prompted by all the multiple `psql` commands.


rija:
Hi @pli888,


The scripts are doing well what they are intended for.
Regarding the workaround for the "out of memory" error, I think it's fine to keep it as the ``develop`` branch
is still setup with the same version of Postgresql as your feature branch.

It's more effective for me to check the problem on more recent version of Postgresql once your work has been merged into ``develop`` and then rebased onto by my feature branch.

I noticed that the ``setup_testdb_gigadb_testdata.sh`` script asks for the database password for its last command.
As this script is going to be run on Gitlab pipelines too, it's probably best it is made password-less like the other script.

rija (approved):
all good.

pli888:
@rija 

The shell scripts `setup_testdb_gigadb_testdata.sh` and `setup_testdb_production_like.sh` are now used in the `run_all_tests` stage in `.gitlab-ci.yml` so that the CI/CD pipeline can create new `gigadb_testdata.pgdmp` and `production_like.pgdmp` files which are then used to provide data in acceptance tests.

The use of Yii migration scripts to provide production_like data in the Postgresql database have been removed from the `with_new_cert_deploy` and `deploy_app` stages. On thinking about GigaDB deployment to a staging server, data is already provided for the Postgresql database on the ocker server using the Ansible `postgres-postinstall` role so not really required in the CI/CD pipeline.

The other major change is that there is now a `ops/scripts/setup_devdb.sh` script which is used for creating the Postgresql database for development work so there is no need to execute multiple Yii migration scripts anymore. There is also a `ops/scripts/setup_prod_likedb.sh` script if more production-like data is required for development work. Using `setup_devdb.sh` script is more consistent with `setup_testdb_gigadb_testdata.sh` and `setup_testdb_production_like.sh` scripts.

Documentation in `README.md` and `docs/DATABASE.md` have been updated to inform users that database setup is done by `setup_devdb.sh`.

There is  a change to `ops/deployment/docker-compose.staging.yml` in which the definition of the `csv-to-migrations` container has been removed. This is not required as it's already present in `docker-compose.yml`.

rija:
Hi @pli888,

this looks good. Some comments inline below

>The use of Yii migration scripts to provide production_like data in the Postgresql database have been removed from the with_new_cert_deploy and deploy_app stages. On thinking about GigaDB deployment to a staging server, data is already provided for the Postgresql database on the ocker server using the Ansible postgres-postinstall role so not really required in the CI/CD pipeline. 

Yes, that makes sense for the initial schema and the data.  
However, schema always evolves and there will always be new migrations to run.
And these new migrations need to be applied in an audited and automated fashion as well.
In short, we  still need to have Gitlab pipelines to apply migrations to all environments, including staging and live.
That said, I think that's beyond the scope of this Pull Request. The story on high availability of database server from the project plan is probably the right place to implement that. 


>The other major change is that there is now a ops/scripts/setup_devdb.sh script which is used for creating the Postgresql database for development work so there is no need to execute multiple Yii migration scripts anymore. There is also a ops/scripts/setup_prod_likedb.sh script if more production-like data is required for development work. Using setup_devdb.sh script is more consistent with setup_testdb_gigadb_testdata.sh and setup_testdb_production_like.sh scripts.

Yes, that's a good idea. Could we actually rename the latter two scripts to express that they output a pgdmp file. e.g:

* make\_pgdmp\_gigadb\_testadata.sh
* make\_pgdmp\_production\_like.sh

I find confusing that they are all called "setup\_*" despite the fact the "\*testdb\*" ones have a different final purpose to the others


>There is a change to ops/deployment/docker-compose.staging.yml in which the definition of the csv-to-migrations container has been removed. This is not required as it's already present in docker-compose.yml.

The ``docker-compose.staging.yml`` is not an override of ``docker-compose.yml`` because for most its services, it must work with images built by the pipeline and doesn't use docker volumes, so if a service is needed on staging, it needs to be defined in the former too, not just the latter.

However, given the first paragraph, that change is still appropriate. We might revisit that when we will do that database high availability work.

PR #554, opened 2021-02-16
Fix 504 Gateway timeout error on dataset pages

pli888:
# Pull request for issue: #537 

A fix for 504 Gateway timeout error on dataset pages when retrieving citation information.

## Changes to the component classes

* `StoredDatasetConnections.php` has been updated to include a clause to catch all 500-level errors which includes the 504 Gateway timeout error sometimes returned by dataset pages such as parrot when citation information cannot be retrieved. If the `strpos` function detects a "504" string in the exception response then this is logged in `application.log`:
```
2021/02/11 15:32:52 [error] [application] Gateway timeout problem with https://doi.org/10.1186/2047-217X-1-17
```

A gateway timeout error will result in the dataset page displaying just the PubMed Ids without the citation information associated with the scientific article identified by the PubMed identifier.

## Changes to the tests

* `DatasetViewTest.php` contains a new functional test `testItShouldDisplayCitations()` to test that citation information is present for the parrot dataset page at `http://gigadb.dev/dataset/100039`.
* `RSSFeedTest.php` has been updated to include the identifier for the parrot dataset page.
* The `testItShouldShowLatestNews()` function in `protected/tests/functional/SiteTest.php` has been updated to include the date (2012-09-11) for the parrot dataset page. 

## Changes to the test data

* The `gigadb_testdata.pgdmp` file has been updated to include the Puerto Rican parrot data. Any dataset requiring retrieval of citation data could have been used by `testItShouldDisplayCitations()` but it is usually this parrot dataset that results in a 504 error in failed acceptance tests.

* A location directive for `/test_gateway_error` to return a 504 error has been added into `ops/configuration/nginx-conf/sites/gigadb.dev.http.conf`.
* A `DEV_GATEWAY_TIMEOUT_ERROR_URL` variable has been added into `ops/configuration/variables/secrets-sample`. The value for this variable should be `http://web/test_gateway_error` in order for the page to return 504 error in the GigaDB web application.


rija (commented):
Hi @pli888,

The implementation in ``StoredDatasetConnections.php`` is nicely done.

One remark regarding the logging: 

```
$exceptionResponse = $se->getResponse()->getBody()->getContents();
if(strpos($exceptionResponse, "504")) {
		Yii::log( '504 Gateway timeout problem with https://doi.org/'. $result['identifier'], "error");
}
Yii::log($exceptionResponse, "error");
```

the two ``Yii::log`` calls are a bit redundant, and we probably wants a friendly error message for other 5xx errors too.
Furthermore, I think the full exception trace is only useful when we debug during development, and in a deployed instance it increases storage usage of logs. 
Maybe we could replace the whole block above with:

```
Yii::log( "{$se->getResponse()->getStatusCode()} {$se->getResponse()->getReasonPhrase()} with https://doi.org/". $result['identifier'], "error");
Yii::log($se->getTrace(), "debug");
```

Regarding the change to the Nginx configuration to add the 504 returning endpoint:
I was rather envisaging it to be used in a new test  so we have a scenario that exercises the exception code every time the test runs.
The changes you made to existing functional tests are all right but we don't know from running them (and the tests are not making assertions about it) wether the call resulted in a 504 nor whether the exception code was called, which actually make me think that's more a job for a unit test rather than functional (something to go in ``StoredDatasetConnectionsTest.php``), and the nginx config change wouldn't be needed in that case.

In any case, I'm not sure how ``DEV_GATEWAY_TIMEOUT_ERROR_URL`` variable functions, I didn't see it referenced anywhere but the secrets.

On a side note, I had to update the ``ops/packaging/Dockerfile``  to specify the `2.9.8`` version for ``xdebug``

pli888:
@rija 

>Maybe we could replace the whole block above with:
>
>Yii::log( "{$se->getResponse()->getStatusCode()} {$se->getResponse()->getReasonPhrase()} with https://doi.org/". $result['identifier'], "error");
Yii::log($se->getTrace(), "debug");

The catch block for `GuzzleHttp\Exception\ServerException` has been updated with the above code now.

>The changes you made to existing functional tests are all right but we don't know from running them (and the tests are not making assertions about it) wether the call resulted in a 504 nor whether the exception code was called, which actually make me think that's more a job for a unit test rather than functional (something to go in StoredDatasetConnectionsTest.php), and the nginx config change wouldn't be needed in that case.

There is now a new unit test `testStoredReturnsPublicationsWithGatewayTimeoutError()` in `protected/tests/unit/StoredDatasetConnectionsTest`. This test simulates two 504 gateway timeout errors which are caught by the `ServerException` catch block when `getPublications()` in `StoredDatasetConnections` is called. The value for the citations key value pair is `null` when there is a 504 error returned by `https://doi.org` and it is this `null` citations value that the `testStoredReturnsPublicationsWithGatewayTimeoutError()` checks for.

>In any case, I'm not sure how DEV_GATEWAY_TIMEOUT_ERROR_URL variable functions, I didn't see it referenced anywhere but the secrets.

The nginx location directive and the DEV_GATEWAY_TIMEOUT_ERROR_URL variable is not required anymore and has been removed.

rija (approved):
Hi @pli888 

All good for me.

(just had to replace ``xdebug`` with ``xdebug-2.9.8`` in the main Dockerfile as a pre-requisite)

Issue #602, opened 2021-04-21
Test container image not building due to node.js installation problem


pli888:
When building the `test` container image in a local `dev` deployment, the following error message appears:
```
Get:16 http://deb.debian.org jessie/non-free amd64 Packages [101 kB]
Fetched 21.3 MB in 41s (517 kB/s)
Reading package lists...
W: GPG error: https://deb.nodesource.com jessie InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 1655A0AB68576280

## Run `sudo apt-get install -y nodejs` to install Node.js 10.x and npm
## You may also need development tools to build native addons:
     sudo apt-get install gcc g++ make
## To install the Yarn package manager, run:
     curl -sL https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --dearmor | sudo tee /usr/share/keyrings/yarnkey.gpg >/dev/null
     echo "deb [signed-by=/usr/share/keyrings/yarnkey.gpg] https://dl.yarnpkg.com/debian stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
     sudo apt-get update && sudo apt-get install yarn
```

This problem came to light when @kencho51's  GitLab CI/CD pipeline was breaking at the `build_webapp` stage with the same error when building a new `application` container which is required if there isn't already one in the `Container Registry`:
```
Get:16 http://deb.debian.org jessie/non-free amd64 Packages [101 kB]
Fetched 21.3 MB in 35s (597 kB/s)
Reading package lists...
W: GPG error: https://deb.nodesource.com jessie InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 1655A0AB68576280
```

The fix is to use this command `curl --silent https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \` to explicitly download the `nodesource.gpg.key` before installing node.js in `Dockerfile`:
```
RUN if [ ${INSTALL_CSSLINT_JSHINT} = true ]; then \
    # To fix public key not found error when running nodesource_setup.sh
    curl --silent https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - && \
    curl -sL https://deb.nodesource.com/setup_10.x -o nodesource_setup.sh && \
    bash nodesource_setup.sh && \
    apt-get install -y nodejs && \
    rm nodesource_setup.sh && \
    npm install -g csslint jshint \
;fi
```

[...]

PR #606, opened 2021-04-22
Fix test image build

pli888:
# Pull request for issue: #602

A fix for building `test` image. This was broken due to a missing package key 
for node.js which is required to authenticate the package in `Dockerfile`.

A step to download and add the nodejs package key has been added into 
`Dockerfile`. This is required for installing nodejs in the `test` image.


kencho51 (approved):
I tested it in one of my feature branch #503 , the CI/CD pipeline is passing now.

rija (approved):
yes, I had the exact same fix on the fuw-cicd branch already.

PR #614, opened 2021-05-05
Fix intermittent failure of DatasetViewTest.php functional test

rija:
Recently StoredDatasetConnections was updated to catch server exceptions that occur often
due to the doi.org API timeout out on requests, resulting in a Service Unavailable error ????

However there is another kind of failure that I have witnessed happening intermittently on my local dev environment
and on CI: The Guzzle client sends request to the doi.org API, but the latter queues the request in a waiting loop.
The "connect_timeout" parameter to Guzzle client actually determines how long it should maintain itself in the awaiting connection stage.
Intermittently (actually 50% of the time on my machine), that threshold will be triggered, causing an error that could not be caught by the Guzzle ServerException because connection wasn't established yet therefore there was no server response.

Additionally, whichever cause makes the API call to fail, it will result in the citation for the associated link to not be retrieved.
The test data has 3 links, and any of them could fail. On my development machine, the frequency of failure of at least two of the link
varies form intermittent to every other run, and on my CI pipeline, it's every time (does that mean that API call on CI is failing every time for some reasons?)

Anyway, the fact that the DatasetViewTest::testItShoulddisplayCitations() test tries to test the presence of the data contained in the API response is problematic, as it is not always there.

The fix is a two throng approach:

1. Add an other catch block to capture Guzzle ConnectException in StoredDatasetConnections.php
2. In the DatasetViewTest::testItShoulddisplayCitations(), assert the existence of the PubMed reference only, not the citation text

#### Changes to protected/components/StoredDatasetConnections.php

Added an additional catch block to capture PHP error thrown when  ``connect_timeout`` is triggered, meaning the remote API hasn't accept the request by that time.
The consequence is the response object doesn't exist and ``GuzzleHttp\Exception\ServerException`` is not thrown.
That's why we need to catch the ``GuzzleHttp\Exception\ConnectException`` that would be thrown when ``connect_timeout`` is triggered.

#### Change to protected/tests/functional/DatasetViewTest.php

The test ``testItShouldDisplayCitations()`` asserts for content that's dependent on the request to the remote API to provide a response, which won't happen if the connection is not even made, resulting in an empty citation text which causes the functional test to fail in those circumstances. Instead we test for something on the page that still pertain to the block of citation links, but will still appear even if the response is empty. 
While not ideal, it still fulfill the original intent for the try/catch work (prevent website users to see a server error when call to remote API fails, and preserve on he dataset view page the visibility of the rest of its content)

### DIFF
--- a/protected/components/StoredDatasetConnections.php
+++ b/protected/components/StoredDatasetConnections.php
@@ -100,7 +100,7 @@ class StoredDatasetConnections extends DatasetComponents implements DatasetConne
 									    'headers' => [
 									        'Accept' => 'text/x-bibliography',
 									    ],
-									    'connect_timeout' => 30
+									    'connect_timeout' => 5
 									]);
 			}
 			catch(GuzzleException $e) {
--- a/protected/components/StoredDatasetConnections.php
+++ b/protected/components/StoredDatasetConnections.php
@@ -98,11 +98,18 @@ class StoredDatasetConnections extends DatasetComponents implements DatasetConne
 			try {
 				$response = $this->_web->request('GET', 'https://doi.org/'. $result['identifier'], [
 									    'headers' => [
+                                            'style' => 'apa',
 									        'Accept' => 'text/x-bibliography',
 									    ],
 									    'connect_timeout' => 5
 									]);
 			}
+            catch(RequestException $e) {
+                Yii::log( Psr7\str($e->getRequest()) , "error");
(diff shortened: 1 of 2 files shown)


kencho51 (approved):
It looks good to me, and the functional test is passing in my branch.

pli888 (approved):
I confirm that the functional DatasetViewTest passed locally and on CI/CD pipeline.

Issue #647, opened 2021-05-17
File Upload Wizard


rija:
Review:
workflow 
PHP backend
Yii2 advanced template
background Workers: move files

Issue #643, opened 2021-05-17
Changes to test infrastructure


rija:
Codeception
setup/tear down of tests for test data setup

This task is part of Story #567

Issue #629, opened 2021-05-17
Create and run search and replace script on local database


rija:
Beware,not all ftp links are hosted by us.
But when it is, it's always the same hostname

Exmaple of query we can use:
```
update media_instance set location=replace(location,'v.','')
where location like '%/0/v.%'
```

This task is part of Story #617

only1chunts:
An additional related point that I've just realised....
The DATASET table contains a variable called ftp_site which also holds a link to the FTP site which needs to be updated.
The screenshot shows the location of this link on the public dataset page for your information:
![Screenshot_11](https://user-images.githubusercontent.com/6037145/118836926-3f485700-b8bc-11eb-9f08-6ea529329f41.png)

At some point in the future it maybe more efficient to make that link a website-level thing just using the DOI instead of adding the full FTP path to the database for every dataset!

rija:
Hi @only1chunts 
that ftp_site link is in the code, not in the database. I don't think it's worth our time  and the risk to make that change.
It's more important to get the links to individual dataset files working which is feasible because it's an update to the database content.
Feel free to make a new PBI to describe how that part should be like in the future.

only1chunts:
Hi Rija,
At the moment that link is in the database not the website code. I'm
suggesting it should be in the website code at some point in the future,
but right now the links in the database need to be updated with the cngb
file server location.
The link is stored in the dataset table.
Cheers
Chris

rija:
Hi @only1chunts,

I guess you are referring to the value of the ``ftp_site`` column in the ``file`` table, then?
I''ll update the acceptance scenarios on the main story #617 to reflect that.
Thanks.

only1chunts:
That's correct. Thanks.

only1chunts:
closing as work is completed

Issue #617, opened 2021-05-06
Use HTTP protocol for downloading FTP link to fix issues with Chrome and Safari


rija:
## User story

>As a website user
>I want to download from GigaDB website the files associated with a dataset 
>So that I can do my work

## Acceptance criteria

>Given I use Chrome web browser to navigate to GigaDB web site  
>When I navigate to the dataset pages's files tab
>And I click on a download link
>Then the file is downloaded

>Given I use Firefox web browser to navigate to GigaDB web site  
>When I navigate to the dataset pages's files tab
>And I click on a download link
>Then the file is downloaded

>Given I use Safari web browser to navigate to GigaDB web site  
>When I navigate to the dataset pages's files tab
>And I click on a download link
>Then the file is downloaded

## Additional Info

* Make production database backup accessible to the whole IT team
* Load production database backup on a local postgresql server
* Create and run search and replace script on local database
* Create and run acceptance tests
* Iterate on batches of datasets
* Restore replaced batches on production

**Note:**
> Behat tests currently use Chrome as backend for test automation, so no problem for implementing the first acceptance test criterium.
> It has been very hard to automate Firefox as backend to Behat as Firefox WebDriver API had long happening issues.
> Safari has a WebDriver API. I haven't tried it yet and I've haven't seen anyone documenting success automating it with Behat yet

**Useful resources:**

* Safari WebDriver tutorial: https://developer.apple.com/documentation/webkit/testing_with_webdriver_in_safari

## Product Backlog Item Ready Checklist

* [x] Business value is clearly articulated
* [x] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [x] Dependencies are identified and no external dependencies would block this item from being completed
* [x] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [x] This item is estimated and small enough to comfortably be completed in one sprint
* [x] Acceptance criteria are clear and testable
* [x] Performance criteria, if any, are defined and testable
* [x] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Code is complete
* [ ] Automated tests related to the changes are implemented and passing
* [ ] All automated test suites are passing locally
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] A Pull Request has been created and review requested
* [ ] Pull Request is reviewed and approved
* [ ] The item has been merged to the develop branch
[...]

PR #662, opened 2021-06-01
Skeleton structure for GigaDB CLI tool and stub classes and methods for Files url updater

rija:
Hi @pli888, @kencho51

This pull request is for skeleton project structure for the tool to update the dataset files url on production database.
It doesn't do anything yet, there is no business logic, but the classes and method stubs are there and the automated tests apparatus (based on Codeception) is set up and working 

we would need this merged to ``develop`` at such a stage so that:

1. It shows the structure and design I plan for this kind of tool. Easier to feedback and correct now that there's nothing yet
2. If any of you want to pick up one the task associated with Sprint story #617, you can branch out of  ``develop`` and build upon that code without spending time in setting things up, while making it a bit easier to merge the various tasks afterwards
3. When there's a need to build a non-trivial GigaDB-related tool with test automation (e.g: backup tool), you can branch out of  ``develop`` and copy this code into a new tool (if not needing exotic dependencies like a legacy postgres, the commands should also run without Docker installed)
4. It reduces time to review upcoming related pull requests as by then the design and structure would have been understood already
5. It will help reviewing aspects of the ``fuw-cicd`` branch as the code design and tests are pulled from the CLI tools on that branch (notably for #643 and #647)
6. It is totally independent from the main gigadb webapp, so no impact to gigadb legacy codebase and operation anyway 

The ``gigadb/app/tools/files-url-updater/README.md`` lists the commands that can be run, they should all be executing without error.

### a. The files url updater can be considered a command associated with GigaDB:

* It has one unique job, which it performs when it is been passed a list of dataset ids. Otherwise it does nothing
* It doesn't require a graphical user interface
* It needs to log what it does and success and failure
* it needs a dry-run mode

### b. We have a detailed acceptance test scenarios to implement, that need to be co-located with the script  and the application needs access to a remote target database server

* This cannot just be a simple bash script in ops/scripts.
* But it's not a complete service like FUW, it's still attached to GigaDB.
* It's not a worker either because we use it interactively, so it doesn't go into ``gigadb/app/worker``.
* The target database server won't be the default one used by the main Gigadb codebase
* Yii2 basic template will setup Codeception for unit, functional and acceptance tests
* so I can create a distinct Yii2 basic application in ``gigadb/app/tools`` as ``files-url-updater``
* There is no need to use the Yii2 advanced template

### c. The application needs access to an old  version of  PostgreSQL database 

So the client library needs to be old (I think maximum 9.3)

We can run the command in a docker container defined by a Dockerfile based on Debian Stretch (that's what's used on ``fuw-cicd`` branch) and install the client library for an appropriately old version of PosgreSQL from PostgreSQL Debian package repository.

We cannot base the Dockerfile on Debian Jessie (which is used by the containers on ``develop`` branch)
as I couldn't find old PostgreSQL client libraries packages for it, and it's good thing to start moving away from the Debian Jessie as it 's too old a version.

The PHP image for the container I want to host the tool in is the official one, but they don't have a Stretch version for PHP 7.4, so we will use Debian Buster, its successor.

# Pull request for issue: #629

n/a

n/a

n/a

n/a

n/a

## Changes to the style

n/a

## Changes to the tests

n/a

n/a


pli888 (commented):
This pull request is good for us to understand the project structure for a standalone GigaDB-related tool in case @kencho51 or I need to implement such a tool in the future. However, it will be populated with business logic for the files url updater tool in a future pull request. This means that we will need to delete this business logic code after we copy this code before starting to create a new tool. Maybe some comments can be added into the code so that we know exactly what code needs adding into classes like the Controller, etc.

Maybe in the future, a [Gii](https://www.yiiframework.com/extension/yiisoft/yii2-gii/doc/guide/2.2/en/topics-creating-your-own-templates) template code generator can be created to auto-generate this skeleton project structure?

To really understand how this project structure works to implement a standalone GigaDB-related tool, I would need to use it. I can try to use this skeleton project structure to implement the Tencent back up tool.

rija:
@pli888 insightful observations. The generator could either be a Gii command or a Composer command (that's what I used to generate that project). The scope of Gii might a bit limited as it doesn't know how to interact with Composer, and having a project with the right dependencies could be very useful.

kencho51:
I think this PR looks good to me. 
But would it be a concern that the code base will become so big if we keep adding tools to GigaDB? 
Would it be possible to have a separate repo for the GigaDB tools?

kencho51 (dismissed):
I think this PR looks good to me.
But would it be a concern that the code base will become so big if we keep adding tools to GigaDB?
Would it be possible to have a separate repo for the GigaDB tools?

kencho51 (commented):
I think this PR looks good to me.
But would it be a concern that the code base will become so big if we keep adding tools to GigaDB?
Would it be possible to have a separate repo for the GigaDB tools?

pli888:
Closing this PR because the final working version of the Files URL Updater tool 
was merged into the develop branch by PR #707.

Issue #742, opened 2021-07-29
[S781] Verify and update Content Security Policy configuration on AWS deployed web site (iframe)


rija:
* Ensure when accessing staging.gigadb.org, iframe content still work
* Ensure nginx configuration for live environment production has content policy settings (``add_header``)

rija:
Finding dataset pages to test:

```
gigadb=> select el.external_link_type_id, elt.name, max(identifier) from external_link el, external_link_type elt, dataset d  where el.external_link_type_id = elt.id and d.id = el.dataset_id and el.external_link_type_id in (2,3,4,5,6,7,8) group by el.external_link_type_id, elt.name;
 external_link_type_id |          name          |  max   
-----------------------+------------------------+--------
                     7 | Github links           | 200143
                     6 | Code Ocean             | 200051
                     2 | Genome browser         | 200140
                     3 | Protocols.io           | 101053
                     4 | JBrowse                | 200079
                     8 | USCS Tumour Map Viewer | 200070
                     5 | 3D Models              | 200070
```

only1chunts:
FYI - "Genome browser" and "Github links" currently dont display as iFrames. 
#266 is about our desires for GitHub links.
Genome browser is a catch-all for non-JBrowse browsers, and I'm still not sure what we want to do with those! for the purposes of this testing ticket you can ignore both "GitHub link" and "Genome browser" types.

rija:
Thanks @only1chunts for the info. I'll still record their status as them not displaying like on live may be symptom of an underlying more general problem, either in code or in the freshness of the test data I use, and even so, we could still decide not to do anything if that make better sense.

Started testing and here is the initial status (had to changes some of the example DOI as several were timeing out on my AWS deployment, which is a problem, but not in scope for this ticket): 

| external_link_type_id |  name  |  example DOI | status on my Fork deployment | status on staging.gigadb.org |   
| --- | --- | --- | --- | --- |
| 2 | Genome browser  			| 101001 | OK | ? |
| 3 | Protocols.io           			| 101053 | No [1] | ? | 
| 4 | JBrowse                 			| 100644 | No [2] | ? |
| 5 | 3D Models             			| 100294 | No [3] | ? |
| 6 | Code Ocean          			| 100909 | No [4] | ? |
| 7 | Github links           			| 100935 | No [5] | ? |
| 8 | USCS Tumour Map Viewer | 100781 |  No [6] | ? |
[...]

PR #850, opened 2021-11-30
Content Security Policy

rija:
# Configuration of Content Security Policy for reducing surface attack and safe use of third party scripts and links

PR for task #742 

## Content Security Policy

Content Security Policy (CSP) is a web standard from the W3C aimed at providing web site owners a mechanism to tell web browsers the rules
for allowing and denying  execution of resources when a user navigates to the owner's web site.

The goal is to mitigate certain types of attacks, especially XSS.

>more info: 
>
>* https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
>* https://owasp.org/www-community/controls/Content_Security_Policy

Although all resources both hosted by the web sites and external resources referred to by the HTML and javascript code are concerned.
With regards to GigaDB, we are particularly concerned with external links and the integration with hypothes.is as each of those involves executing several external resources (javascript, CSS, ...)

The aforementioned rules are called policies and are specified in response headers from the web servers using a specific configuration syntax

>See cheatsheet here: https://scotthelme.co.uk/csp-cheat-sheet/

In our case the headers are defined in ``ops/configuration/nginx-conf/sites/nginx.target_deployment.https.conf.dist``, in the ``Security headers`` section of the ``server`` block.

CSP is not trivial to setup and they're are two types of pitfalls that can lead to either in failure in reducing attack surface and/or breakage of web site functionalities:

* policy misconfiguration
* too permissive policies

>Note: this blog post is a good illustration of those pitfalls: https://www.troyhunt.com/how-to-break-your-site-with-content/

To help with the policy definition process, the CSP specification has two modes, a reporting-only mode and an enforcement.
The former will have web browsers to only report violations, while the latter mode will have the web browsers to actually block the resources in violation of policies.
It is recommended to start with report-only mode, then switch to enforcement when we fixed the violations form our own resources or authorised resources.
The two mode can be used at the same time, for example when we want to introduce a new policy for a new type of external link, we define that policy in a report-only header, while all previously defined policies stay in their existing policy enforcement header.

An attribute of CSP header is the possibly to specify the url of a service where to send policy violation report. It can be a self provisioned service or an external service.
For GigaDB, I have created an account (it's free and trustworthy) at ReportURI (https://report-uri.com) which offer detailed, nicely navigable reports (amongst other features)

The above paragraph is also added to ``docs/SECURITY.md``

What to do with the content generated report ? Analyse (using the CSP report and web browser dev tools' inline debugger) to detect patterns and possible grouping and create product backlog items (PBI) for later sprint scheduling as part of Technical debt repayment.

## Changes

The changes in this PR are:

* The addition of  two response headers in the ``ops/configuration/nginx-conf/sites/nginx.target_deployment.https.conf.dist`` nginx template: ``Content-Policy`` and ``Content-Policy-Report-only``
* Addition of a CSP section in ``docs/SECURITY.md``

Another related change but not in codebase is the addition of ReportURI credentials to GitLab ``cnhk-infra`` project's variables:

| Key | Masked? |
| --- |  --- |
| reporturi_email | no |
| reporturi_password | yes |

With those changes the issues highlighted in #742's table now work: see https://github.com/gigascience/gigadb-website/issues/742#issuecomment-983708401

## Known issues out of scope of this PR

* Issues with Github external links: they are not appearing, not because of CSP, it's a backend concern (server side code or data) (see #855)
* iframe in tabs that are not of type "Protocols.io","JBrowse","3D Models", "Code Ocean" won't load due to discrepancy between CNGB code and GitHub code (see #856)


pli888 (approved):
Looks good. I understand that we need to remember to update the CSP policies whenever we use content from a new external site in the future.

kencho51 (approved):
This PR makes me understanding more about the CSP and its importance in securing a site. I am happy to approve it.

Issue #864, opened 2021-12-02
Error page showing CDbException error when creating user account


rija:
**Describe the bug**

**To Reproduce**
1. Go to http://staging.gigadb.org/login
2. Click on "Create account"
4. Instead of the user account creation form, there is server error stack trace

**Expected behavior**
A user creation form should appear at http://staging.gigadb.org/user/create

**Screenshots**

**Additional context**

* Possibly same problem in #859 
* Need to create a new acceptance test to pick this problem up
* Why doesn't the current acceptance test detect this problem?

rija:
Done

PR #880, opened 2021-12-06
A server error is thrown when loading the new user creation form

rija:
# Pull request for issue: #864

Fixing the form for new user creation (#864)

n/a

## Changes to the model and controller classes

* ``protected/models/User.php``
* ``protected/controller/UserController.php``

The code handling the form was not handling Yii model scenarios causing issue with the password field.
Correct handling requires the controller to specify upon instanciaiton of a new model, a scenario that matches the set of attributes that can be altered on that model for the given scenario (the validation method call should also take that scenario as input that was already the case) . The model needs a change too to indicate in which scenario it is safe to alter 
the password attribute.

Additional change to the controller involves wrapping a buggy email sending feature inside a try/catch to have the error logged instead of crashing the form submission and leaking stack trace to the user.
 
>Note: Fixing the problem with email sending is not in the scope of this PR

A further change to the controller is to add a test to detect whether the code is running on a ``dev`` and ``CI`` environment, in which case we set the captcha session variable to whatever is passed in the form so automated test can pass captcha verification while avoiding to mess up with the backend validation code.

>Note: Fixing the big problem with the captcha is not in the scope of this PR.

Finally, a minor issue (revealed when writing functional test) was fixed to ensure the generated image content is in PNG format.

* ``protected/views/user/_form.php``

The root cause of the problem raised in #864 lies in this file. The feature for linking author record and user record only make sense for a user that exists, so it shouldn't be active on  the new user creation form. I've fixed the code so the worklflow is not invoked for new user creation

## Changes to the style

n/a

## Changes to the tests

Installed Codeception for the main codebase, so that we can start reducing dependency on legacy testing libraries that prevent us from migrating PHP versions (in order to improve peformances and mitigate the security vulnerabilities uncovered by container scanning and SAST tools):

* Installed Codeception (see ``ops/configuration/php-conf/composer.json.dist`` and ``composer.lock``)
* Configured Codeception for acceptance tests (see ``tests/acceptance.suite.yml``)
* Configured Codeception for functional tests (see ``tests/functional.suite.yml``)
* Implemented the steps of the feature file scenario in ``tests/_support/AcceptanceTester.php``
* Added Codeception acceptance tests run command to ``tests/acceptance_runner``

Created and implemented two acceptance test scenarios to test the new user creation process:

* Added a feature file for new user creation: ``tests/acceptance/NewUser.feature``
* Added required generic steps in ``tests/_support/AcceptanceTester.php``
* Added required steps specific to ``asa:Curator`` user story in ``tests/_support/CuratorSteps.php``

Change to functional tests:

Replaced coverage-incomplete functional test for captcha (removed ``protected/tests/functional/CaptchaImageTest.php``)
with new Codeception functional tests (in ``tests/functional/CaptchaCest.php``) that cover more and relevant behaviours of our captchas usage:

* the tests cover both captcha usage, the one on contact form and the one on user form.
* we test that the generated captcha is different between successive call
* no need to test for the image size specific to each form, as captcha image presentation
should be the same across usage. We are not harmonising sizes in this task, it will be done later when we fix the big captcha bug
* we test the captcha is image/png mime which revealed a minor bug in User form controller (discrepancy between file format and generated format) which we fix here (see controller changes above)

That functional test refactoring is made more necessary now that acceptance tests scenario for user creation form is bypassing captcha using server variables and environment detection as we need to make sure captcha mechanism has test coverage.

The new functional tests called for the creation of a helper function in ``tests/_support/Helper/Functional.php`` to help with comparing two images.

n/a


pli888 (commented):
When running the [tests](https://gitlab.com/gigascience/forks/pli888-gigadb-website/-/jobs/1870249288) on CI/CD pipeline, I get the following error:
```
/usr/local/bin/docker-php-entrypoint: 9: exec: ./vendor/codeception/codeception/codecept: not found
```

rija:
> When running the [tests](https://gitlab.com/gigascience/forks/pli888-gigadb-website/-/jobs/1870249288) on CI/CD pipeline, I get the following error:
> 
> ```
> /usr/local/bin/docker-php-entrypoint: 9: exec: ./vendor/codeception/codeception/codecept: not found
> ```

Yes, I had that and it went away after clearing the cache of the pipeline (as the composer's vendor directory is saved in gitlab caches)

pli888 (approved):
Tests are passing on my CI/CD pipeline after clearing runner cache and code looks good. I approve this PR.

kencho51 (approved):
Hi @rija Unit test, functional test and acceptance are passing in my dev computer and also my CI/CD pipeline, so I approve this PR.
Will spend more time on understanding `codeception` later.

Issue #877, opened 2021-12-03
static content test failed: "dataset page"


only1chunts:
generally the static content of the dataset page is OK with 2 exceptions:
 -  The metrics icons appear to be pre-ceeded by a little hyphen, see red arrows in screenshot below:
![Screenshot_41](https://user-images.githubusercontent.com/6037145/144075287-4926b8a1-036c-477c-865e-bd94d511f5bd.png)

 - similarly, the linked project images are missing from the staging server, so not showing on dataset pages.

_Originally posted by @only1chunts in https://github.com/gigascience/gigadb-website/issues/743#issuecomment-982739373_

* Write acceptance test to check single dataset page contains an image

rija:
Closing as fixed long time ago

PR #885, opened 2021-12-08
Remove extra hyphen mark between the semantic icons

kencho51:
# Pull request for issue: #877

1. The hyphen between the semantic icons has been removed

image

## Changes to the tests

1. All three semantic links has been updated and pass the acceptance test.


kencho51:
Hi @rija @pli888 , I want this PR merged first as it has passed all the acceptance tests. 
Will come back to this using `migration script` after all sprint 5 bugs gone.

rija (commented):
Hi @kencho51, 

if you wish to divide the work on issue #877 in several parts, it's better to follow logical partitioning
otherwise the PRs will be harder to understand and it will create confusion.
The issue #877 mentions clearly two distinct problems (in hindsight, the reporter should have created two separate issues), so divide your work based on that .

This PR seems to be about the first problem (images for semantic links), so you will want to remove from it any changes and tests that are not about fixing and checking just that (and there's no point adding project images assets if they cannot be referenced yet).

Then create a new bug issue on GitHub for the second problem (project images) and make a comment on #877 about the new issue and that #877 is only about 
the images for semantic links (and change the issue title appropriately). That new bug can be tackled later (and I'll be happy to pair with you on it if so you wish)

rija (commented):
@kencho51
Thanks for removing the project images.

However the hyphen fix is not made in the right place:
changes need to made in/with FormattedDatasetMainSection.php's ``getCitationsLinks()`` method, the call of which you removed from the view.

rija (changes requested):
Hi @kencho51

Below are three preliminary observations:

(1) I notice the url for Google Scholar and Europe are "http", but we should  use "https" everywhere.
So the scenario steps need to be changed accordingly and so are  the config file for those links:  ``ops/configuration/yii-conf/main.php.dist``

(2) This unit tests is failing for me: ``FormattedDatasetMainSectionTest::testFormattedReturnsCitationsLinks``. 

We don't want to duplicate the acceptance tests here, so the actual HTML snippet is not important in the context of the unit test (but we still need the unit tests to ensure the logic is exercised and does the flow we want, so I suggest the following patch as a fix:
```
diff --git a/protected/tests/unit/FormattedDatasetMainSectionTest.php b/protected/tests/unit/FormattedDatasetMainSectionTest.php
index 6ec7f0d31..3db0dcc8c 100644
--- a/protected/tests/unit/FormattedDatasetMainSectionTest.php
+++ b/protected/tests/unit/FormattedDatasetMainSectionTest.php
@@ -224,16 +224,8 @@ class FormattedDatasetMainSectionTest extends CTestCase
         );
 
         $expected = array(
-            "scholar_query" => '<span class="citation-popup" data-content="View citations on Google Scholar">
-                            <a href="http://scholar.google.com/scholar?q=10.5072/100243" target="_blank">
-                                <img class="dataset-des-images" src="/images/google_scholar.png" alt="View citations on Google Scholar"/>
-                            </a>
-                        </span>',
-            "ePMC_query" => '<span class="citation-popup" data-content="View citations on Europe PubMed Central">
-                            <a href="http://europepmc.org/search?scope=fulltext&query=(REF:\'10.5072/100243\')" target="_blank">
-                                <img class="dataset-des-images" src="/images/ePMC.jpg" alt="View citations on Europe PubMed Central"/>
-                            </a>
-                        </span>',
+            "scholar_query" => '<span>1</span>',
+            "ePMC_query" => '<span>2</span>',
         );
 
         //we mock the CachedDatasetMainSection
@@ -250,7 +242,7 @@ class FormattedDatasetMainSectionTest extends CTestCase
         $daoUnderTest = new FormattedDatasetMainSection(
                             $cachedDatasetMainSection
                         );
-        $this->assertEquals($expected, $daoUnderTest->getCitationsLinks() ) ;
+        $this->assertCount(count($expected), $daoUnderTest->getCitationsLinks());
     }
 
     /**

```

(3)  ``StaticContentNavigation.feature`` is  not the best place for this scenario. This is a feature on the dataset page, so it should be in a ``DatasetView.feature`` using the same feature story as in ``dataset-view.feature``. To avoid future trivial merge problem, you can checkout that file from one of my branch of open PR  and remove everything but the feature before adding your scenarios: 
https://github.com/gigascience/gigadb-website/pull/906

protected/views/dataset/view.php
```diff
@@ -19,20 +19,18 @@
...
-                <?= CHtml::image($url, $model->image->tag ,
-                    array(
-                        'class'=>'media-object',
-                        'title'=>$model->image->tag.' License: '.$model->image->license.' Source: '.$model->image->source.' Photographer: '.$model->image->photographer
-                    )
-                ); ?>
-            </a>
-            <?php } ?>
-
+                            <?php if($model->image) {
```
pli888, line 22: These changes in `protected/views/dataset/view.php` are not related to the icons that link out to citation resources? It seems this part of the code is for displaying the dataset image.

pli888 (commented):
The code in this PR looks good to me now but the tests are not passing on my CI/CD pipeline probably because of the recent website performance issues. These are now fixed in PR #912 so after it has been merged, we can run the tests for this PR after it is rebased with `develop` branch.

rija (changes requested):
Hi @kencho51,

Thanks for the changes, look good. Here are a couple of suggestions of further changes:


(1) it's ok and expected for special characters like single quotes to be escaped in URLS, that's actually what the W3C standard definition says we SHOULD [1] do

So you can change both:

* the scenario  step to use a url like ``https://europepmc.org/search?scope=fulltext&query=(REF:%2710.5072/100094%27)``
* the ``main.php.dist`` config file's ``urls>ePMC_query`` key to have as value: https://europepmc.org/search?scope=fulltext&query=(REF:%27@id%27)

which will remove the need for a custom step that perform conversion process that's not necessary

(2) It's probably best to have a separate scenario for each of the service as they are different aspects of the user story and it would help ensure we've covered every angles (eventually) for each service while maintaining the scenarios simple to read and to validate


[1] https://url.spec.whatwg.org/#url-rendering-i18n

protected/views/dataset/view.php
kencho51: The content of the code has not been refactored, except the alignment of the code, which would make it more readable. But you are right, maybe better do it in #895 .

rija:
Hi @kencho51 

Thanks for implementing the changes. They work fine.

However, there is a unit test failing for me:
```
1) StoredDatasetMainSectionTest::testStoredReturnsCitationsLinks
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'services' => Array (...)
     'urls' => Array (
-        'scholar_query' => 'http://scholar.google.com/scholar?q=10.5072/100243'
-        'ePMC_query' => 'http://europepmc.org/search?scope=fulltext&query=(REF:'10.5072/100243')'
+        'scholar_query' => 'https://scholar.google.com/scholar?q=10.5072/100243'
+        'ePMC_query' => 'https://europepmc.org/search?scope=fulltext&query=(REF:%2710.5072/100243%27)'
         'dimension_query' => 'https://app.dimensions.ai/dis...100243'
     )
     'images' => Array (...)
 )

/var/www/protected/tests/unit/StoredDatasetMainSectionTest.php:156
```

rija (approved):
Hi @kencho51,
All the tests are passing now, both on my local environment and on CI.
Could see the fix working on my staging environment too.
I'm happy to approve this PR.

pli888:
All the tests are passing on my pipeline - I approve this PR.

Issue #872, opened 2021-12-03
static content test failed: help page tables


only1chunts:
**Describe the bug**
The content of the tables in the help pages appears to be OK, but the dynamic sizing of the tables within the page is not working, i.e. on a narrow window (like I'm using for testing side by side) the current live site has a slider bar to scroll left/right in the tables, that is missing from the staging version

**To reproduce**
1. Go to http://gigadb.org/site/guide
2. reduce with of browser window
3. see the tables not resize or provide slider bars.

**Screenshots**
![guide-table-issue](https://user-images.githubusercontent.com/6037145/143900633-8771603f-68e2-452a-a3ff-e3306def9670.png)

**Additional information **
_Originally posted by @only1chunts in https://github.com/gigascience/gigadb-website/issues/743#issuecomment-981770698_

PR #894, opened 2021-12-17
Add scroll bar  to the tables in guide page

kencho51:
# Pull request for issue: #872

A scroll bar is added to each table in all the guide pages, for example(screenshots taken by codeception):

table_submission

table_attribute

table_details

Each table is wrapped in a separate `<div></div>` and with unique id.  
The scroll bar is available by adding the `class="table-scrollbar"`, which has attribute `overflow: auto`. The value `auto` will make the scroll bar visible when necessary.

## Changes to the tests

The acceptance tests are passing by asserting that  the table element contains the `class="table-scrollbar"`, this means the presence of the scroll bar.  The method `I make a screenshot of a table :table` could take a screenshot of the table and provides a visualised proof. 

Moving of the scroll bar should also be tested, but I found it difficult to locate the scroll bar and instruct codeception to move it, so this remains undone.


kencho51:
Hi @rija,

I am trying to pass the acceptance test on moving the scroll bar of each table. But I don't know how to locate the scroll bar exactly that I could move it, can you have a look on this for me?

The step `I scroll the table "table_submission" horizontally by "2000" "50"` could only locate the table, but not the scroll bar,
test-scrollbar2

 I have also tried `clickWithLeftButton()` which did not work neither.

less/current/layout/section.less
```diff
@@ -11,4 +11,8 @@ section.page-title-section {
 
 .subsection {
   margin-bottom: 30px;
+}
+.table-scrollbar {
```
rija, line 15: There is nothing table specific about this selector.
Also we may want to use it on other things rather than just tables,
so it's better rename the selector to just ``.scrollbar``

tests/acceptance/StaticContentNagvigation.feature
```diff
@@ -0,0 +1,52 @@
+Feature: A user visit gigadb website
```
rija, line 1: There is typo in the name of the file, it should be ``StaticContentNavigation.feature`` instead of ``StaticContentNagvigation.feature``

rija (changes requested):

rija (approved):
Thanks for changes.
And scroll bars function on my environment using Safari, Chrome and Firefox, after I've emptied browser cache.
I'm happy to approve this PR.

>note: don't forget to synchronise with develop as there are conflicts

pli888 (approved):
@kencho51 The code in this PR looks fine and I can see the scroll bars for the tables working across Safari, Firefox and Chrome browsers. I approve this PR - just requires rebasing with the latest code in `develop` branch to fix code conflicts.

Issue #895, opened 2021-12-20
static content test failed: project image not showing


kencho51:
Originally posted by @only1chunts in #743.

### Problem
It was reported in #877 that the linked project images are missing from the staging server, so not showing on dataset pages.

Fix for the project image not showing will be referred to this ticket.

PR #896, opened 2021-12-20
Fix project image not showing

kencho51:
# Pull request for issue: #960 (originally in #895)

All projects images have been downloaded from `http://gigadb.org/images/project/` directory and put into aws s3 bucket `https://assets.gigadb-cdn.net/images/projects/`.  All the project images are being served from CDN servers provided by cloudflare. Then updated the `image_location` in `project` table in `data/dev` for local testing, while the `m220126_062259_update_image_column_location_project_tab.php` script would do the same for staging deployment. 

So when this PR is merged, as the project images has been stored in the codebase, the dataset page will render it directly. 

##### Before the fix
image

##### After the fix in staging
image

## Changes to the tests

1. Project images with links pass the acceptance test.

Here is the way to run the migration script and pass the test locally:

```
% ./up.sh
#Make migration work in local gigadb website
% docker-compose run --rm  application ./protected/yiic migrate --migrationPath=application.migrations.fix_import
% docker-compose run --rm test ./vendor/codeception/codeception/codecept run -g issue-895 acceptance tests/acceptance/DatasetView.feature
```


protected/views/dataset/view.php
```diff
@@ -234,9 +234,12 @@ class="btn btn-green btn-block claim-button"
                     <h5><strong><?=Yii::t('app' , 'Projects:')?></strong></h5>
                     <p>
                         <? foreach ($projects as $project){
-                            echo $project['format'];
+                            $image_url = $project['url'];
+                            $image_name = $project['name'];
+                            $image_location = preg_replace("/^http:\/\/gigadb.org/m", "", $project['image_location']);
+                            echo "<a href='$image_url' target='_blank'><img alt='Go to $image_name website' src='$image_location'/></a>";
```
rija, line 240: Hi @kencho51, 
making the replacement in the view template is not ideal because:

* that's more processing on the view template which will slow down the loading of that public page
* the data in the database will still be incorrect, so we will need to duplicate the fix if we want to use the same data somewhere else than on that page

I did mention database migration, but that's not the great either because this is more about content problem rather than schema, and we want to fix the data problem only once.

So it turns out that this is similar problem to the file url to replace, and adapting file-url-updater to fix project images is the better approach there.

rija:
Hi @kencho51,

after revisiting the problem we are trying to solve here, I reckon using database migrations is the most reasonable approach in our current circumstance:

* A SQL instruction is enough to solve the problem
* Unlike file ftp urls, we don't need to try to correct the data in the CNGB database, we just need to ensure that a new deployment to AWS using existing data has correct location for the project images
* Project images are fairly static and unlike dataset images, are a small number and we don't expect to have to do this regularly
* Making it a command line tool, means:
	*  addditional code, which need more unit/functional tests to write
	* take longer to implement
*   @pli888's work on PR #899  has been merged and is a good model to emulate as both migratiosn belong to the same category (fixing imports) even though one changes schema and the other changes data


Here's the SQL to put in the ``safefUp``:

```
update project set image_location=regexp_replace(image_location,'http://gigadb.org(.*)','\1') returning *;
``` 

And here's the one to put in the ``safeDown``:

```
update project set image_location=regexp_replace(image_location,'^/(.*)','http://gigadb.org/\1') returning *;
```

The above syntax should be familiar to you as it resembles the PHP code of your implmentation. However, do note that the SQL approach fixes the problem at the origin which has following benefits:

* whichever endpoints need access to project image (e.g: APIs for third parties)  they will get the fix too
* we are note making the most popular page on gigadb.org slower with extra presentation layer processing
* no additional code

rija:
Hi @kencho51 

there is a failing test on this branch that is making think, there is something we've done with the migration that's is not quite correct:

```
2) filter tables on admin page: Projects
 Test  tests/acceptance/AdminTablesFilter.feature:Projects
 Step  I should see "the avian phylogenomic project" 
 Fail  Failed asserting that  on page /adminProject/admin
--> Home|
About|
Contact|
Terms of use
Help Joe's GigaDB Page Administration LogOut
Manage Projects
Create New Project
Displaying 1-1 of 1 result.
Url
Name
Image Location
    http://avian.genomics.cn/en/index.html the_avian_phylogenomic_project https://assets.gigadb-cdn.net/images/projects/the_avian_phy
[Content too long to display. See complete response in '/var/www/tests/_output/' directory]
--> contains "The Avian Phylogenomic Project".

Scenario Steps:

 20. $I->see("The Avian Phylogenomic Project") at tests/_support/AcceptanceTester.php:54
 19. $I->wait("1") at tests/_support/AcceptanceTester.php:219
 18. $I->pressKey("(//input)[2]","") at tests/_support/AcceptanceTester.php:210
 17. $I->fillField({"name":"Project[name]"},"Avian") at tests/_support/AcceptanceTester.php:152
 16. $I->amOnPage("/adminProject/admin") at tests/_support/AcceptanceTester.php:38
 15. $I->waitForText("Home",10) at tests/_support/CuratorSteps.php:39
```

We need to use the under_scored version of the the project name in order to build the image url, but we should alter the name. The name of a project is "The Avian Phylogenomic Project" and should remain so , not ``the_avian_phylogenomic_project``

rija:
@kencho51 

you should be able to use this:

```
update project 
set url = concat('https://assets.gigadb-cdn.net/images/projects/', lower(replace(replace(replace(name, ' ','_'),' - ','_'),',','')),  image_location) 
where image_location != '' and image_location not like 'https://assets.gigadb-cdn.net/images/projects/%';
```
after removing the queries with ``set name = lower(name)`` and ``set name = replace( name...``

rija (approved):
Hi @kencho51,

After I run

```
docker-compose run --rm application ./protected/yiic migrate --migrationPath=application.migrations.fix_import --interactive=0
```

I could see the functionality working on my local deployment.

All the tests are now passing locally and on CI.

After I have provisioned a new staging environment and deployed this PR's branch's code on it, I was able to see the functionality working for me.

Code changes look fine to me.

I'm happy to approve this PR.

pli888 (approved):
The code looks good to me and I can get the functionality working on my local dev (after running `fix_import` database migration) and on my personal staging server. All tests are passing on my CI pipeline.

I guess we should remind ourselves to update the `project.csv` files in the `data` directory when we host gigadb.org on the cloud so `image_location` contain the CDN links to the images.

Issue #849, opened 2021-11-30
CException errors on 2 admin pages (staging.gigadb.org)


only1chunts:
**Describe the bug**
From testing #747 checklist item "do all admin pages allow appropriate updates to the database?" NO
Two admin pages give CException errors; "Dataset_sample" and "Samples"

### Dataset_Sample page
**To Reproduce**
1. Go to 'https://staging.gigadb.org/adminDatasetSample/admin' https://staging.gigadb.org/adminSample/admin
2. See error
<html>
<body>
[...]

PR #897, opened 2021-12-21
849 samples admin crash fix

rija:
# Pull request for issue: #849

>It fixes the two admin pages that couldn't be navigated to because a server error is displayed when doing so:
> * /adminSample/admin
> * /adminDatasetSample/admin

The cause is that on those pages there is special presentation behaviour done on list of sample attributes identical to what is done on dataset view page's table.
However the code of that behaviour was on the backend model layer (presentation code should  not be on model layer as it complicates maintenance, change and understanding) and was moved to a presentation layer when the public dataset view was refactored. But the admin pages were not updated accordingly.

n/a

Updated ``protected/models/Sample.php`` with a method returning sample attribute as a list of name/value pair

## Changes to the presentation layer classes

the formatting class ``protected/components/FormattedDatasetSamples.php`` already has the code to handle the sample attributes descriptions correctly. We needed to change the visibility of the associated method so we can use them in the affected view templates.

This is not the proper approach overall to the problem: the software design for the admin page is old and inefficient and the proper approach is to refactor using modern modular SOLID and Onion based PHP like we did for the public dataset page.
But given this bug fix is to support urgent migration off the old infrastructure, we temporaliy borrow methods from the outer layer (``Formatted*`` classes) of the onion architecture, which is OK since it's still presentation layer.

Then we can update the two view templates below to fix the issue using the changes we made so far:
* ``protected/views/adminDatasetSample/admin.php``
* ``protected/views/adminSample/admin.php``

## Changes to the style

n/a

## Changes to the tests

We need to update a model class (``protected/models/Sample.php``), so we needed to work on unit tests.
So I've configured Codeception so it can be used for unit tests and added the tests I need:

* Copied ``protected/tests/unit_bootstrap.php`` as ``tests/unit/bootstrap.php`` and refers to it in ``tests/unit.suite.yml``
* Udpated ``tests/unit_runner`` to also run the Codeception based unit tests.
* Created a new unit test file tests/unit/SampleTest.php I need for unit testing my fix

Also created the feature files for each admin page:

* ``tests/acceptance/AdminSample.feature``
*  ``tests/acceptance/AdminDatasetSample.feature``

and updated the ``tests/_support/CuratorSteps.php`` to sign in as an admin.

In ``.gitlab-ci.yml``, bumped the version of Docker DinD to get around a build error.


pli888 (approved):
@rija I confirm the code in this PR fixes the problems in the two endpoints below which are present in the current `develop branch`:
1. /adminDatasetSample/admin causes CException ~> Property "Sample.displayAttr" is not defined.
2. /adminSample/admin causes CException ~> Property "Sample.fullAttrDesc" is not defined.

All tests are passing in my CI/CD pipeline and the code looks good to me. I approve this PR.

tests/acceptance/AdminDatasetSample.feature
```diff
@@ -0,0 +1,23 @@
+Feature: admin page for samples
+  as a curator
+  I want to see a table of all d dataset/samples associations
+  So that quickly navigate to the sample related data I am interested in
+
+  Background:
+    Given I have signed in as admin
+
+  Scenario: sample attributes short description an call to action for more
```
pli888, line 9: Needs *@ok* tag and so does the scenario below?

tests/acceptance/AdminSample.feature
```diff
@@ -0,0 +1,13 @@
+Feature: admin page for samples
+  as a curator
+  I want to see a table of all samples
+  So that quickly navigate to the sample related data I am interested in
+
+  Background:
+    Given I have signed in as admin
+
+  Scenario: sample attributes full description
```
pli888, line 9: Needs *@ok* tag?

rija:
@pli888 thanks again for spotting. My latest commit fixes that

kencho51 (approved):
Hi @rija 
All tests are passing in my local dev environment and on my CI/CD pipeline.
The fix is working on my dev gigadb website and staging website.
So, I would like approve this PR.

Issue #879, opened 2021-12-06
Security problem


rija:
See my email for details. Will update this ticket once the fix is live

rija:
Alghough fully functional, I missed some captcha file path related code that are no longer needed

PR #898, opened 2021-12-21
879 captcha file name problem

rija:
# Pull request for issue: #879

The main problem is the file name of the generated captcha image is defeating the purpose of a captcha image.
Additionally, there's duplication of code as the two areas (new user creation and contact form) that use captcha duplicate the same code in their respective controller.

n/a

* ``protected/models/User.php``
* ``protected/models/ContactForm.php``

those two model class were updated to replace the existing captcha validation logic that was based on existence of captcha image file. After realising that the validation code has become identical in both class, moved it in the new service component to avoid duplication (see below)

* ``protected/components/CaptchaService.php``
* ``ops/configuration/yii-conf/local.php.dist``

There is now a service component that will be used by the user creation form and the contact form for the generation and the image output of the captcha text (and by any form we want to add captcha to in the future), as well as for validation of the form submission.

Since it's a ``CApplicationComponent`` class, we need to register it in the Yii configuration file template, which allow the service to be instantiated only once and yet be available to the endire main gigadb codebase using ``Yii::app()->captcha`` calls

* protected/controllers/SiteController.php
* protected/controllers/UserController.php

Remove the problematic and now redundant captcha handling code.
Also in the user controller class, updated the deployment environment test condition so that unsuccessfull captcha validation can be exercised (this is done in acceptance test scenario, see below)

* protected/views/user/_form.php
* protected/views/site/contact.php

These view files for new user creation and contact form respectively have been updated to replace the old code that used controller code by call to the Captcha service we've created.

The service is needed twice, the first time to generate the text and save it in current session.
The second time, it is to output the image as  base64 raw image data [1], so there is no need to store image on file system.

[1] https://en.wikipedia.org/wiki/Data_URI_scheme

## Changes to the style

n/a

## Changes to the tests

* ``tests/functional/CaptchaCest.php``
* ``tests/acceptance/NewUser.feature``

The functional tests for testing captcha behaviour have been updated to take into account the use of raw image data in the ``src`` attribute of the ``img`` element.
I've added a scenario to the new user creation feature to test when a user enter an incorrect captcha, so that validation use cases can be fully exercised.

* ``ops/configuration/php-conf/composer.json.dist``
* ``composer.lock``

For the core mechanic of generating a random text string image that is human usable, we rely on a Composer library ``gregwar/captcha``:
 * is actively maintained
 * many stars and forks while  issues are answered and PRs processed
 * has test and documentation
 * is lightweight
 * is customisable
 * is simple to use

So, composer configuration files were updated accordingly to add support for that library. 

* ``ops/deployment/docker-compose.production-envs.yml``

Since we no longer have to handle physical captcha image, we can get rid of the Docker volume dedicated to that usage.

* ``ops/configuration/nginx-conf/sites/nginx.target_deployment.https.conf.dist``

We also need to add a new CSP directive for ``img-src`` resource to authorise web browser to load raw image data with the ``data:`` scheme.


tests/acceptance/NewUser.feature
```diff
@@ -43,3 +42,19 @@ Feature: NewUser
     And I fill in the field "User_verifyCode" with "shazam"
     And I press the button "Register"
     Then I should see "Welcome!"
+
+  Scenario: Providing erroneous captcha prevents submission and show error message
```
pli888, line 46: Does this scenario need an `@ok` tag in order for it to be run?

pli888 (approved):
@rija The code looks good to me and I confirm that the captcha image is embedded in base64 format within the `/site/contact` HTML page. Just wondering if the erroneous captcha acceptance test needs an `@ok` tag? All tests are passing on my CI/CD pipeline after I cleared my runner cache. I approve this PR.

tests/acceptance/NewUser.feature
rija: @pli888, thanks for spotting, yes it should have a \@ok tag. Will amend accordingly

kencho51 (approved):
Hi @rija ,

I confirm that the captcha image file is in base64 format in both /site/contact  and /user/create pages by looking at the page source in my dev gigadb website. The functional test and acceptance test are passing  in my dev environment as well. The code looks good to me, so I would like to approve this PR, although the CI/CD pipeline is failing because #912 is not merged yet.

Issue #857, opened 2021-12-01
Status "Private" datasets should not be visible on base URL


only1chunts:
**Describe the bug**
regardless of logged-in status I can view PRIVATE datasets without using the token URL.

**To Reproduce**
1. Go to 'https://staging.gigadb.org/dataset/200070'
2. See the Private dataset 200070!

**Expected behavior**
A private dataset (such as 200070) should not be visible on the base URL https://staging.gigadb.org/dataset/200070 or any derivative thereof (e.g. https://staging.gigadb.org/dataset/view/id/200070 or https://staging.gigadb.org/dataset/view/id/200070/File_page/1 etc...)
A private dataset (such as 200070) should only be visible via its token-URL, e.g. 
https://staging.gigadb.org/dataset/200070/token/ImP3Bbu7ytRSfYFh

## Additional information

* Maybe bug relating to status information in codebase

rija:
Fixed in PR #906

PR #906, opened 2021-12-30
857 private dataset visible

rija:
# Pull request for issue: #857

>A dataset whose upload status  is set to "Private" was still visible to the public

n/a

n/a

* ``protected/controllers/DatasetController.php``

There were several problems in the controler action method for showing dataset (``actionView()``),  more preciselly with the workflow based on the type of page (``invalid``, ``hidden`` or ``public``):

* A missing ``else`` clause after detection of the type of page was causing the normal rendering of the dataset whichever type of page
* Incorrect detection of mockup url (with ``token`` fragment)
* no call to the Yii Controller's ``render()`` method in the block dealing with private set mockup page

The logic was fixed and made more readable.

In the solution, there is a use of an anonymous function that solves the following dilemma:

 * it is best practice to keep the ``render()`` method in the Controller's action method for readability and simplicity
 * I needed to use that method twice with the same parameters (once for mockup page and once for public view) but I don't want to duplicate the same block of code

n/a

## Changes to the style

n/a

## Changes to the tests

* ``tests/acceptance/DatasetView.feature``
*  ``data/dev/dataset.csv``
* ``data/dev/image.csv``
* ``data/dev/gigadb_user.csv``
*  ``tests/_support/WebsiteUserSteps.php``
*  ``tests/acceptance.suite.yml``
* ``tests/acceptance_runner``

I have added two scenarios to the dataset view user story to cover private dataset behaviours: not visible unless accessed with the token based mockup url.

That necessitated augmenting the test data in ``data/dev`` directory so we have an example of private dataset.

To support the implementation of not signed in user step from the Codeception scenarios, I've created ``tests/_support/WebsiteUserSteps.php`` class to support actions by the ``website user`` role.

The ``tests/acceptance_runner`` needed to be updated to ensure the database is setup with ``dev`` test  data prior to run the Codeception acceptance tests.

n/a


data/dev/gigadb_user.csv
```diff
@@ -1,6 +1,7 @@
 id,email,password,first_name,last_name,affiliation,role,is_activated,newsletter,previous_newsletter_state,facebook_id,twitter_id,linkedin_id,google_id,username,orcid_id,preferred_link
 8,test+8@gigasciencejournal.com,5a4f75053077a32e681f81daa8792f95,Guojie,Zhang,BGI,"",f,t,t,,,,,test+8@gigasciencejournal.com,,EBI
 14,test+14@gigasciencejournal.com,5a4f75053077a32e681f81daa8792f95,Shifeng,Cheng,BGI,"",f,t,t,,,,,test+14@gigasciencejournal.com,,EBI
+38,admin1@gigasciencejournal.com,bcf839ae349d13cb0c3c4a9ec0cc243e,C,H,BGI,admin,t,t,t,,,,114715812788877222335,admin1@gigasciencejournal.com,,EBI
```
pli888, line 4: @rija This line seems to be causing the acceptance test for the `Users` scenario in `AdminTablesFilter.feature` to fail for me when executed locally and on CI/CD pipeline. This scenario fills in the search text field for the `Email` column with "admin" which detects this `admin1` user but the `And I should not see "@gigasciencejournal.com"` scenario step then cannot be passed.

pli888 (approved):
When I execute `./up.sh dev` on the code in this PR to install dev data, I can view dataset 200070 with its [token URL](http://gigadb.gigasciencejournal.com:9170/dataset/200070/token/ImP3Bbu7ytRSfYFh). This unpublished dataset cannot be viewed with using its public [URL](http://gigadb.gigasciencejournal.com:9170/dataset/200070) which is the correct behaviour.

If I switch to the current `develop` branch code then this unpublished dataset 200070 can be viewed using its [public URL](http://gigadb.gigasciencejournal.com:9170/dataset/200070), confirming the presence of this #857 bug.

Therefore, the code in this PR looks good to me and provides the required functionality. All tests are passing on my CI/CD pipeline. I approve this PR.

kencho51 (approved):
Hi @rija ,

I confirm that the private dataset `200070` is only available via the token link, and the acceptance test is passing as well.
If I switch to other branch, the private dataset `200070` becomes publicly available, so I further confirm this PR is working as expected.
Besides the CI/CD pipeline is failing because #912 is not merged yet, everything looks good to me, so I would like to approve this PR.

Issue #867, opened 2021-12-03
smoke test fail: download submission template or examples


only1chunts:
**Describe the bug**
able to download the templates for sample and files within submission wizard? NO. 
When I try to download the submission spreadsheet template, or either of the example forms i get an "Error 404." Unable to resolve the request "files/GigaDBUploadForm.xls".

**To Reproduce**
1. Go to https://staging.gigadb.org/datasetSubmission/upload
2. Click on "Download Template File" button
3. See error:
![image](https://user-images.githubusercontent.com/6037145/144614165-a2f81564-0f8a-42b1-9aaf-8fd2c0dd4a6e.png)

**Expected behavior**
The file selected is expected to be downloaded.

**Additional context**
The same error is seen for all 3 download buttons on that page:
"Download Template File"
"Download Example File 1"
"Download Example File 2"

## Additional information

* Retrieve missing files from CNGB server
* Place them in the code base
* Modify production container to get them into correct location

PR #907, opened 2022-01-03
Enable download template files from update page as user

kencho51:
# Pull request for issue: #867 and #938

1. Included the latest template files to the codebase.

The latest version `v22Dec2021` of upload templates are placed in `/files/templates` directory.

2. Fixed the broken link for downloading the template files.

The download is working on dev by updating the file path from `<a href="/files/file_name" class="...">XXX</a>` to `<a href="/files/templates/file_name" class="...">XXX</a>`

The latest version `v22Dec2021` of upload templates could be downloaded when press the buttons on page `/datasetSubmission/upload`.

3. The `Excel template file` link  in `/site/help#guidelines` is also updated as requested in #938 .

## Changes to the tests

1. Created `DownloadGigaDBTemplateFiles.feature` to test the download buttons are linked with correct file path.

2. To test the files are downloadable, a functional test `DownloadTemplateFilesCest` is created. After running the functional test `./tests/functional_runner`, the files will be downloaded to the file system, and `assertFileExists` function will check the existence of the files.


tests/_support/AcceptanceTester.php
```diff
@@ -227,4 +225,30 @@ public function iShouldSeeAnImageIsLinkedTo($image, $expectedUrl)
         $actualUrl = $this->grabAttributeFrom("//img[@src='$image']/parent::*", "href");
         $this->assertEquals($expectedUrl, $actualUrl);
     }
+
+    /**
+     * @When I sign in gigadb website as :user
+     */
+    public function iSignInGigaDBWebsiteAsUser($user)
```
pli888, line 232: This function is not required because you can use `iSignInAsAUser()` in `AuthorSteps.php`?

tests/acceptance/DownloadGigaDBTemplateFiles.feature
```diff
@@ -0,0 +1,22 @@
+Feature: An author can download file template
+  As an author
+  I want to download upload form template after login
+  So that I can upload data using the form
+
+  Background: Sign is as a user
+    Given I sign in gigadb website as "user"
```
pli888, line 7: This will need changing if the `iSignInAsAUser()` step is used from AuthorSteps.php.

tests/acceptance/DownloadGigaDBTemplateFiles.feature
```diff
@@ -0,0 +1,22 @@
+Feature: An author can download file template
+  As an author
+  I want to download upload form template after login
+  So that I can upload data using the form
```
pli888, line 4: I think something like this will be more accurate:
```
I want to download the dataset metadata upload form template after login
So that I can upload information about my dataset using the form
```

tests/acceptance/DownloadGigaDBTemplateFiles.feature
```diff
@@ -0,0 +1,22 @@
...
+
+  Background: Sign is as a user
```
pli888, line 6: Should be `Sign in as a user`.

tests/acceptance/DownloadGigaDBTemplateFiles.feature
```diff
@@ -0,0 +1,22 @@
...
+    Then I should see a link "Download Template File" to "/files/templates/GigaDBUploadForm-forWebsite-v22Dec2021.xlsx"
+    And I should see a link "Download Example File 1" to "/files/templates/GigaDBUpload-Example1-forWebsite-v22Dec2021.xlsx"
+    And I should see a link "Download Example File 2" to "/files/templates/GigaDBUpload-Example2-forWebsite-v22Dec2021.xlsx"
+
+  @ok
+  Scenario: Download template file as a user
+    When I am on "/datasetSubmission/upload"
+    Then I press "Download Template File" and the response should contain "200"
+    And I press "Download Example File 1" and the response should contain "200"
+    And I press "Download Example File 2" and the response should contain "200"
```
pli888, line 22: Each of these 3 steps read like they should be 2 steps:
```
And I press "Download Template File"
Then the response should contain "200"
```

This is what is done in [`dataset-view.feature`](https://github.com/gigascience/gigadb-website/blob/develop/features/dataset-view.feature) but I think I understand why you combined the 2 steps because it does not look like there is an easy way to test HTTP response code in a separate step in Codeception like you can do in Behat acceptance tests.

I guess you could test for actual file download in an acceptance test - see [here](https://stackoverflow.com/questions/32828524/codeception-how-do-i-test-file-download). Maybe @rija has another solution.

rija: Hi @kencho51,

@pli888 makes good point here,

I'd actually go further saying that in addition to that, you should make a separate scenario for each template file (I think I made a similar comment on another of your PR)
as they are different things that although are treated identically for now but that might change in the future.
We don't want to have to meddle with an existing and passing test scenario when a new template comes in or one of the template needs to be treated differently.

I think pressing a button and checking the files exists locally is a reasonable approach.

Testing for HTTP response is more of functional/integration tests level. Acceptance tests are from a user/product owner perspective and they don't care much about http response.

rija:
The newly raised #867 is another manifestation of the issue. It makes to implement its fix as part of this PR.

tests/acceptance/DownloadGigaDBTemplateFiles.feature
kencho51: Hi @rija , 

I tried to test for actual file download, first by specifying the download path in `acceptance.suite.yml` and using `file_exist()` for checking.
But I found out that, the download action cannot be triggered by `I press the button "Download Example File 1"` as no file is downloaded. I tried to explicitly `I press the button "http://gigadb.test/files/templates/GigaDBUpload-Example1-forWebsite-v22Dec2021.xlsx"`, but got `invalid CSS and XPath selector and Link or Button` error. 
Can you have a look for me?

pli888: I've not been able to get file download working acceptance tests today. Maybe the `selenium/standalone-chrome:3.141.59-oxygen` image in `docker-compose.yml` could do with  being updated to a newer version.

ops/deployment/docker-compose.production-envs.yml
```diff
@@ -25,6 +25,7 @@ services:
       - ${DATA_SAVE_PATH}/fuw/repo:/var/www/filedrop
       # public download path (TODO: temporary, to revisit in storage work stream)
       - /tmp:/var/www/datasetfiles
+      - templates:/var/www/files/templates
```
kencho51, line 28: Hi @pli888 and @rija ,

Need to specify this volume, otherwise will get this error on staging like:
```
Error 404
Unable to resolve the request "files/templates/GigaDBUploadForm-forWebsite-v22Dec2021.xlsx".
```

rija: Hi @kencho51, 

For production environments the protocol is to copy static files inside container image by using COPY in ``Production-Web-Dockerfile``.

You can see that being done already for the files in ``files/examples/``.

So you follow the same model for files in ``files/templates``.

The idea is that a container image is like a CD-ROM, a golden disc image that  contains a readonly snapshot of all our files, make it easier to rollback in case of problems, know exactly what set of file a  particular deployment consist of and for auditabilty.

So we copy all the files of the codebase into those immutable images: php code goes into ``Production-dockerfile``, and static files (because they are served by nginx only, not php-fpm) go into ``Production-Web-Dockerfile``

We only use volumes for when we know the deployed application needs to write something to the filesystem (like the Let's Encrypt certificates or some dynamic assets created by the Yii framework)

kencho51: Hi @rija,

The `- templates:/var/www/files/templates` has been removed. After rebuilding the containers in staging server through gitlab pipeline, the template files download works as expected.

rija: @kencho51 you sure about that? 
how could it work on staging if you don't copy them inside the production container image?

rija:
@kencho51 
Don't forget this PR. It's first in the queue.  What's left is update teh production dockerfile to ensure the files are downloadble on production environments

rija (approved):
Hi @kencho51

All tests are passing locally and on CI.
The new acceptance tests scenarios make sense.
The functionally works on my dev environment and on my personal staging deployment as well.
Therefore I'm happy to approve this PR.

pli888 (approved):
The new code looks good to me and is working as described on my local dev and personal staging server. The tests are passing on my CI pipeline too. I approve this PR.

Issue #886, opened 2021-12-08
CDbException when updating file metadata


only1chunts:
**Describe the bug**
This bug exists on current CNGB live production server too so its not something introduced in the code re-write, but it does need to be fixed.
When a file is updated via the admin pages it throws a CDbException error if the release date is not set.

**To Reproduce**
1. Go to any edit file page, e.g. https://staging.gigadb.org/adminFile/update/id/464662
2. delete the release date if present
3. Click on save

**Expected behavior**
There is a bigger issue that does need to be addressed (see ticket #887) but for initial release of the dev-staging codebase it will be sufficient to remove the requirement for the release date to be set when updating files via the admin page.

**Additional context**
<html>
<body>
[...]

PR #909, opened 2022-01-05
Server error occurs when updating file as an admin with no set release date

rija:
# Pull request for issue: #886

server error occurs when updating file metadata with an unset release date.
The root cause is that unsetting the date in the UI cause an empty string to be passed in the SQL to update the record, which is considered (rightfully) as an invalid date format by PostgreSQL.

n/a

* ``protected/models/File.php``

File is a subclass of CActiveRecord.
We've added a  ``beforeSave()`` hook that overrides Yii CActiveRecord's default ``beforeSave()`` in which 
we intercept changes to the release date field (``date_stamp``) and if it's an empty string
we set it to null.

n/a

n/a

## Changes to the style

n/a

## Changes to the tests

* ``tests/acceptance/AdminFileForm.feature``
*  ``tests/acceptance.suite.yml``
*  ``tests/acceptance/initdb.php``
*  ``tests/acceptance_runner``

Added acceptance tests scenarios to test that release date can be unset and can be changed.
Also added a boostrap script to the acceptance tests suite so that it can be run manually and independently as it can reset the database state independently.

n/a

### DIFF
--- a/protected/models/File.php
+++ b/protected/models/File.php
@@ -303,7 +303,23 @@ class File extends CActiveRecord
 
     /**
      * Before save file
+     *
+     * ensure the release date can be unset
+     *
+     * @see https://www.yiiframework.com/doc/api/1.1/CActiveRecord#beforeSave-detail
+     * @return boolean
      */
+    public function beforeSave()
+    {
+        if (parent::beforeSave()) {
+            $this->date_stamp = ('' === $this->attributes['date_stamp']) ? null :  $this->attributes['date_stamp'] ;
+            return true;
+        }
+        else
+            return false;
+
+    }
+
     public function setSizeValue()
     {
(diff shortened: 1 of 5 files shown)


pli888 (approved):
The code looks good to me and I confirm its functionality that allows the release date for a file to be `Not set` on admin file view pages such as `/adminFile/view/id/17` on a dev installation. Also, all the tests are passing on my CI pipeline. I approve this PR.

kencho51 (approved):
Hi @rija ,

The acceptance test is passing on my dev, and the fix can be confirmed by visiting one of the files, like, `/adminFile/view/id/5107` that the `Release date` value is `Not set`.  The CI/CD pipeline is passing as well. So, I would like to approve this PR.

Issue #918, opened 2022-01-14
Composer config has hard-coded "7.1" value as platform constraint and may cause errors


rija:
```
$ docker-compose run --rm test composer update yiisoft/yii yiisoft/yii2
yiisoft/yii2-composer contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
Do you trust "yiisoft/yii2-composer" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?] y
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires php ^7.2 but your php version (7.1.30; overridden via config.platform, actual: 7.2.34) does not satisfy that requirement.

```

PR #920, opened 2022-01-15
Fix fatal error when installing composer library due to hardcoded php version

rija:
# Pull request for issue: #918

* A fatal error occurs when installing a new composer library (see #918 for details)
* $PHP_VERSION and $NGINX_VERSION set in ``.gitlab-ci.yml`` or ``.env`` should be the only input for determining which version of PHP and Nginx respectively are used across all our systems
* Artifacts created after deployment  contains base images which led to large files size (1GB+) having to be uploading from finishing pipeline jobs (slowing the build pipeline and wasting space on the cloud and developer's machine)

* ``ops/configuration/php-conf/*.ini``

We've streamlined PHP configuration by eliminating all versions specific configurationce since now PHP 7.2 is the default for all parts of our codebase. Therefore we only keep a php configuration for dev, CI and production (that cover both staging and live) environments

* ``ops/deployment/docker-compose.yml``
* ``ops/packaging/Dockerfile``
* ``ops/packaging/Production-Dockerfile``
* ``ops/configuration/php-conf/composer.json.dist``

Replaced hard-coded reference to PHP version with $PHP_VERSION variable and ensure the correct INI config file is referenced

* ``ops/packaging/Web-Dockerfile``
* ``ops/packaging/Production-Web-Dockerfile``

Replaced hard-coded reference to Nginx version with $NGINX_VERSION

* ``ops/pipelines/gigadb-build-jobs.yml``
* ``ops/pipelines/gigadb-deploy-jobs.yml``

Streamline the prelaoding of base images to use $PHP_VERSION and $NGINX_VERSION for refering to which versions to use.

Also in the latter file, make sure the base images are not included in the artifacts.

* ``.gitlab-ci.yml``

Now that they are the canonical object for determining which versions to use, we ensure that the $PHP_VERSION and $NGINX_VERSION variables are set with  the correct versions of those software


ops/packaging/Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG PHP_VERSION=7.1
-FROM php:7.2-fpm-buster
+ARG PHP_VERSION=7.2.34
+FROM php:${PHP_VERSION}-fpm-buster
```
pli888, line 3: It kind of feels like the Dockerfiles in `ops/packaging` have the PHP version hardcoded. So for example if we upgrade the PHP version in `.env` to 7.3 then the value of `PHP_VERSION` in these Dockerfiles need to be updated too.

pli888 (approved):
The code looks good to me. I just have a query re: PHP versioning in Dockerfiles. All tests are passing on my CI pipeline. I approve this PR.

kencho51:
Hi @rija ,

The `NGINX_VERSION` and `PHP_VERSION` in  `ops/configuration/variables/env-sample` should also be updated as well?

rija:
Hi @kencho51, 

Thanks for spotting that, and it's now corrected.

kencho51 (approved):
Hi @rija ,

The code looks good to me now, so I would like to approve this PR.

ops/packaging/Dockerfile
rija: Hi @pli888 
if it's because of the ``ARG`` clause,  its value is a default value that in normal circumanstances would never be used 
since PHP_VERSION is populated wth value either set in ``.env`` or ``.gitlab-cy.yml``.
Therefore, there is no need to change the default value  for ``ARG`` whatsoever when we upgrade the version of PHP. 
If I could, I would prefer leave ``ARG`` empty, but the Dockerfile syntax force us to put a value in there.
Maybe I should set it something like ``ARG=null`` to avoid any confusion

Unless we think that when extraordinary circumstances where neither ``.env`` nor ``.gitlab-cy.yml`` set PHP_VERSIOM, the container image should still build with some version of PHP.

rija: > If I could, I would prefer leave ARG empty, but the Dockerfile syntax force us to put a value in there.

Ok, from the Dockerfile reference, it seems that we can use ``ARG`` without a default values

>Unless we think that when extraordinary circumstances where neither .env nor .gitlab-cy.yml sets PHP_VERSION, the container image should still build with some version of PHP.

No, if PHP_VERSION is not set from ``.env`` or ``.gitlab-cy.yml``, we don't want the container image to build at all, as this is abnormal circumstances (that surely willl result in other issues anyways), in which we don't weant to operate

So, I've removed the default value

Issue #865, opened 2021-12-03
Edit personal details smoke test failed


only1chunts:
**Describe the bug**
Unable to unsubscribe from the mailing list.
Note. the account I was using to test with was already subscribed to the mailing list so I don't know if it would be possible to subscribe via the edit personal details page?

**To Reproduce**
1. Go to https://staging.gigadb.org/user/view_profile
2. See "Add me to GigaDB's mailing list" checkbox is already checked and unable to uncheck it.

**Expected behavior**
on the edit profile page it is expected that a user can choose to un-subscribe from the mailing list.

**Screenshots**
![Screenshot_43](https://user-images.githubusercontent.com/6037145/144611816-afc8bbd7-9197-4246-9f1d-cf9cb97c7bf3.png)

** Additional info **
This is a bug found during tests listed in #747 

* Need to check why checkbox is greyed out
* Need to create an acceptance test to confirm proper behaviour

PR #923, opened 2022-01-18
Fix user subscribing to mailing list functionality

pli888:
# Pull request for issue: #865

* Enable user to subscribe / unsubscribe to mailing list.

The functionality of the mailing list checkbox is now under the control of an `onclick` event listener. When an author is viewing their profile page then clicking the checkbox will have no effect. However, if the `Edit` button is clicked then the `onclick` event listener is removed which allows the checkbox to be checked or unchecked.

Previously, users were stopped from checking/unchecking the mailing list checkbox with the `disabled` attribute. However, using the `disabled` attribute for the checkbox will now stop its corresponding change in the `mailing_list` column in the `gigadb_user` table from happening with our current codebase. Using `onclick='return false'` mimics the functionality of the `disabled` attribute of the checkbox as well as updating the `mailing_list` column.

## Changes to the tests

There are a number of new steps in `AcceptanceTester.php` which can be used for creating new acceptance tests:
* iShouldSeeAButton($buttonName)
* iShouldSeeCheckboxIsNotChecked($checkbox)
* iShouldSeeCheckboxIsChecked($checkbox)
* iCheckCheckbox($checkbox)
* iUncheckCheckbox($checkbox)

The first step is used to detect presence of the `Edit` button in the profile page. The latter 4 steps are used to detect and configure checkboxes. It should be noted that `iCheckTheField($fieldName)` is already in `AcceptanceTester.php` and can be used to check checkboxes. But I think `iCheckCheckbox($checkbox)` should replace `iCheckTheField($fieldName)` because the former is more descriptive and explicit in its function name as to what it does. If `iCheckTheField($fieldName)` is removed then DatasetUpload, ChangePassword and NewUser feature files need to be updated accordingly.

The changes in `AcceptanceTester.php` are used by the new `UserProfile.feature` which contains acceptance tests for the mailing list checkbox.


rija (commented):
Hi  @pli888, @only1chunts


I found the workflow for that functionality very confusing, which results in inconsistent implementation.

Testing locally on this branch, when I am on the user profile, the check box is not actionable as expected from my understanding of the bug.
So I click edit and on the update form and there I could interact with the checkbox.
When I click "Save" in the form and I am back on user profile, the check box state reflects my interaction in the update form , as i expected.
However if I didn't click "Save" and instead clicked "Cancel", when I'm taken back to the user profile, the checkbox suddenly become actionable on the user profile.

In any case, I think the check-box should not be at all visible on the user profile, because a check-box is a form element and there's no form on the user profile.
mailing-list membership of a user is an information about the user like their email or affiliation.

and the user profile should probably look like:

```
Email: user@gigadb.org
First Name: John
Last Name: Smith
Affiliation: BGI
Link out preference: EBI
MailingList Subscriber: Yes (or No)
```

The check-box should only be in the update form that appears when the user clicks "Edit"
If you think it's better to have a checkbox look to visually indicate membership on user profile, then don't use the HTML form input for check-box for that,
instead use a web icon like glyphicons [1] or fontawesome [2] and still present it in the same way (and aligned) as the other data about the user like this

```
MailingList Subscriber: [√]
```


[1] https://glyphicons.com/,  which is included with Bootstrap, so that's probably the one to pick
[2] https://fontawesome.com/

only1chunts:
I am OK with the user profile using either the text as Rija suggested or a visual checkbox glyphicon, whichever is the easier to implement.
Thanks

pli888:
>In any case, I think the check-box should not be at all visible on the user profile, because a check-box is a form element and there's no form on the user profile.
mailing-list membership of a user is an information about the user like their email or affiliation.

>and the user profile should probably look like:

>Email: user@gigadb.org
First Name: John
Last Name: Smith
Affiliation: BGI
Link out preference: EBI
MailingList Subscriber: Yes (or No)
The check-box should only be in the update form that appears when the user clicks "Edit"

This is done now.

rija (approved):
Hi @pli888,

The tests are all passing locally and on CI.
The functionality works as expected locally and on my staging environment.
The new acceptance scenarios are good and the code is fine.
I'm happy to approve this PR

kencho51 (approved):
Hi @pli888 ,

I confirm that the functionality is working as expected locally and in staging gigadb website.
And the acceptance test is passing locally and in gitlab pipeline.

Issue #915, opened 2022-01-11
Caching is broken


rija:
The cache mechanism has stopped working. It's a little while like that as we can feel it empirically on staging.gigadb.org.

Adding extra logging in StoredDatasetFiles::getDatasetFiles() as part of PR #912, 
we can see the evidence of it:
(if cache was working, there would have been only one log entry after two consecutive calls)

```
[11-Jan-2022 21:49:35] NOTICE: fpm is running, pid 1
[11-Jan-2022 21:49:35] NOTICE: ready to handle connections
[11-Jan-2022 21:57:12] WARNING: [pool www] child 7 said into stdout: "-> Starting StoredDatasetFiles::getDatasetFiles() <-"
[11-Jan-2022 21:57:12] WARNING: [pool www] child 7 said into stdout: "in /var/www/protected/components/StoredDatasetFiles.php (51)"
[11-Jan-2022 21:57:12] WARNING: [pool www] child 7 said into stdout: "in /var/www/protected/components/CachedDatasetFiles.php (53)"
[11-Jan-2022 21:57:12] WARNING: [pool www] child 7 said into stdout: "in /var/www/protected/components/FormattedDatasetFiles.php (50)"
[11-Jan-2022 21:57:12] WARNING: [pool www] child 7 said into stdout: "in /var/www/protected/components/FormattedDatasetFiles.php (93)"
[11-Jan-2022 21:57:12] WARNING: [pool www] child 7 said into stdout: "in /var/www/protected/views/dataset/view.php (5)"
172.16.238.10 -  11/Jan/2022:21:56:44 +0000 "GET /index.php" 200
172.16.238.10 -  11/Jan/2022:21:57:31 +0000 "GET /index.php" 200
[11-Jan-2022 21:58:41] WARNING: [pool www] child 8 said into stdout: "-> Starting StoredDatasetFiles::getDatasetFiles() <-"
[11-Jan-2022 21:58:41] WARNING: [pool www] child 8 said into stdout: "in /var/www/protected/components/StoredDatasetFiles.php (51)"
[11-Jan-2022 21:58:41] WARNING: [pool www] child 8 said into stdout: "in /var/www/protected/components/CachedDatasetFiles.php (53)"
[11-Jan-2022 21:58:41] WARNING: [pool www] child 8 said into stdout: "in /var/www/protected/components/FormattedDatasetFiles.php (50)"
[...]

PR #924, opened 2022-01-19
Decouple debug mode from caching control

rija:
# Pull request for issue: #915

Make sure the existing variables ``YII_DEBUG``, ``YII_TRACE_LEVEL`` and the new ``DISABLE_CACHE``
can be configured like any other variables using gitlab variables or ``.env``
and the ``generate_config.sh`` script.

Also ensure we have sensible default values depending on which environment we are in
(dev/CI or staging/live) for all the above variables. This is done in ``generate_config.sh``
(and also in ``gigadb-build-jobs.yml`` for the default ``YII_DEBUG`` setting for staging).

The real invalidation query is now defined in a configuration file (``main.php.dist``) for convenience of modification.

This work solves the current problem we are seeing on staging where the caching of DB queries stopped working when
we've enabled YII_DEBUG on staging, and the general problem of finding balance between getting a lot of debugging information on a staging environment and the need of live-like performances on that environment.

The work enables the following:

| Key | Role | Default on Dev/CI | Default on Staging | Default on Live |
| ---  | --- | --- | --- | --- | 
| YII_DEBUG | enable debug mode for extra logging | true | true | false
| YII_TRACE_LEVEL | how many lines of context for log entries | 3 | 0 | 0 | 
| DISABLE_CACHE | whether to disable caching of DB queries | true | false | false |

>Note: the value of each of these three variables has impact on website performances. 
> The default values for the live environment offer the maximum performance. 
> While the default values for Dev/CI provide the most debugging information.

>Note: those three variables set the values for PHP constants of the same names that are
> defined in the Yii web application's ``index.php`` file (generated from templates  ``ops/configuration/yii-conf/index.$GIGADB_ENV.php.dist``)

The above table and notes have also been added in the documentation file ``docs/SETUP_CI_CD_PIPELINE.md``

* ``protected/components/DatasetComponents.php``

When ``YII_DEBUG`` is set to true, we display in the log whether a request for data was a cache hit or cache miss.
Depending on whether ``DISABLE_CACHE`` is set to true or false we set the cache invalidation query to current time or the configured real invalidation query.

## Changes to the tests

* ``tests/unit/DatasetComponentsTest.php``

Unit tests for the changes we have made to the ``DatasetComponents.php`` class.

## Changes to the configuration

* ``ops/configuration/yii-conf/index.CI.php.dist``
* ``ops/configuration/yii-conf/index.dev.php.dist``
* ``ops/configuration/yii-conf/index.live.php.dist``
* ``ops/configuration/yii-conf/index.staging.php.dist``
* ``ops/configuration/yii-conf/main.php.dist``

## Changes to the documentation

* ``docs/SETUP_CI_CD_PIPELINE.md``

* ``ops/pipelines/gigadb-build-jobs.yml``
* ``ops/scripts/generate_config.sh``

## Additional changes to fix Centos 8 repo issues

* ``ops/infrastructure/bastion_playbook.yml``
* ``ops/infrastructure/dockerhost_playbook.yml``
* ``ops/infrastructure/roles/centos-eol-fix/tasks/main.yml``
* ``ops/infrastructure/roles/docker-install/tasks/main.yml``
* ``ops/infrastructure/roles/fail2ban/tasks/main.yml``
* ``ops/infrastructure/roles/postgres-preinstall/tasks/main.yml``

Centos 8 has been end-of-lined on 31st January 2022, causing issues with fetching packages from its repository.

Error like this now happen when provisioning a staging or live environment with Ansible:
```
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
```

An approach put forward was to replace the url in the repo with the Centos Vault version of them.
The problem with that, is that packages in the vault don't get updates.
If a security vulnerability occur and is exploited and a patch is made available our host system would not get access to it, 
thus leaving the host vulnerable.

The approach I've taken is to switch the distribution from abandoned Centos 8 to the supported Centos Stream 8, which 
will allow for an upgradable base system.

However, most of the software we need for a GigaDB deployment are not available yet in the supported system repo.
And in any case, they are not affected by the issues related to Centos 8.

But because the  distribution switch is implemented by assigning Centos Stream 8 code to the ``releasever`` variable,
and the repo for those third party software use that variable, we need to hardcode the code for Centos 8 in those repos 
(and only those ones, not the base system ones).

All of that is implemented in a new Ansible role ``centos-eol-fix``, which is used in both playbooks.

To make it work, I had to pull all the tasks related to installing yum repos from our  other custom Ansible roles and put them in the new role.


kencho51 (approved):
Hi @rija,

All tests are passing in dev and in gitlab pipeline.
By looking at the log of application container, cache for dataset_xxx_CachedDataset*  is `MISS` in dev, but `HIT` in staging as expected.
The staging servers are provisioned by using the updated config which replaced the EOL centos 8.
All codes look good to me, so I would like to approve this PR.

ops/scripts/generate_config.sh
```diff
@@ -122,9 +109,19 @@ TARGET=${APP_SOURCE}/protected/config/console.php
...
 
+
+if [[ $GIGADB_ENV != "dev" && $GIGADB_ENV != "CI" ]];then
+  export YII_TRACE_LEVEL=${YII_TRACE_LEVEL:-"0"}
+  export YII_DEBUG=${YII_DEBUG:-"false"}
+  export DISABLE_CACHE=${DISABLE_CACHE:-"false"}
+else
+  export YII_TRACE_LEVEL=${YII_TRACE_LEVEL:-"3"}
+  export YII_DEBUG=${YII_DEBUG:-"true"}
+  export DISABLE_CACHE=${DISABLE_CACHE:-"true"}
```
pli888, line 120: `YII_DEBUG` is in `env-sample`. Do we need to add `YII_TRACE_LEVEL` and `DISABLE_CACHE` into `env-sample` too?

rija: @pli888, from a functional perspective it isn't required as the defaults are already set in the configuration script.
However from a documentation perspective it's better that they are present in ``env-sample``, so I've committed your suggested change.

pli888:
@rija Using at a known bug, I confirm that `YII_DEBUG` variable is able configure debug mode between different environments. Caching of DB queries is also configurable using `DISABLE_CACHE`. I did wonder how the staging environment manages to get `YII_DEBUG="true"` as this isn't defined in `generate_config.sh` but then saw that this is done in` gigadb-build-jobs.yml`. I confirm the new `centos-eol-fix` Ansible role works for me. I approve this PR.

pli888 (approved):
@rija Using at a known bug, I confirm that `YII_DEBUG` variable is able configure debug mode between different environments. Caching of DB queries is also configurable using `DISABLE_CACHE`. I did wonder how the staging environment manages to get `YII_DEBUG="true"` as this isn't defined in `generate_config.sh` but then saw that this is done in` gigadb-build-jobs.yml`. I confirm the new `centos-eol-fix` Ansible role works for me. I approve this PR.

Issue #862, opened 2021-12-01
Error when attempting to submit dataset spreadsheet via webpage


only1chunts:
**Describe the bug**
Error message returned when attempting to submit a prepared spreadsheet submission "Upload failed. Please contact database@gigasciencejournal.com"

**To Reproduce**
1. Go to https://staging.gigadb.org/datasetSubmission/upload
2. Tick agree to terms and conditions
3. add a file to be uploaded
4. click "upload new dataset" button
5. see error message "Upload failed. Please contact database@gigasciencejournal.com"

**Expected behavior**
the website is suposed to send the spreadsheet to database@gigasciencejournal.com AND send a confirmation/thank you letter to the submitter.

** Additional information ** 
Ticket created from smoke testing question "are automatic emails being sent on dataset submission" in #747 

* Spreadsheet is sent as an attachment to database@gigasciencejournal.com
* Probably configuration problem

rija:
reopen as still a problem on staging.
Maybe a problem with the email sending details

pli888:
>reopen as still a problem on staging.

@only1chunts This is working on the staging server now with the spreadsheet sent as an attachment in the email. Note that the attachment does not have the original Excel filename though.

only1chunts:
Yes it sends the attachment, but as you noted it does mess up the file name, including removing any file extension which means its difficult to open it. If there is an easy fix to this it would be great to have it now, but if its something that will take a while then we can go "live" with it for now and add it to the backlog for later.
NB- I also noticed a very Ugly breakage when I accidentally clicked the "Upload dataset" button before selecting a file, so that should be made more elegant in the future.

pli888:
>NB- I also noticed a very Ugly breakage when I accidentally clicked the "Upload dataset" button before selecting a file, so that should be made more elegant in the future.

Clicking "Upload New Dataset" button before selecting a file is also a problem on the live server. It's because the file selector and `Upload New Dataset` button both become enabled in the front-end code after the `terms and conditions` checkbox is checked.

PR #953, opened 2022-02-08
Fix problems with dataset spreadsheet upload

pli888:
# Pull request for issue: #862 Error when attempting to submit dataset spreadsheet via webpage

* Correct behaviour between Terms & Conditions checkbox, input file selector for choosing spreadsheet to upload and Upload New Dataset button on the `/datasetSubmission/upload` page
* The filename of the spreadsheet attachment in the email now has the original spreadsheet filename that has been uploaded by the user

## Changes to the Component classes

The `sendHTMLEmailWithAttachment()` function in `MailService.php` now requires a new argument `$attachmentFileName` which provides the filename that the attachment should have instead of a randomly generated name.

The `actionUpload()` function in `DatasetSubmissionController.php` has been updated so that it can provide the `sendHTMLEmailWithAttachment()` function in `MailService.php` with a filename for the attachment in the email.

The `protected/views/datasetSubmission/upload.php` file has been updated to correct the behaviour between Terms & Conditions checkbox, input file selector for choosing spreadsheet to upload and `Upload New Dataset` button. Checking the Terms & Conditions checkbox will now enable the input file selector but keep the `Upload New Dataset` button disabled. This button will only be enabled when the user selects a file to upload. This has been implemented by separating the control of the input file selector and the Upload New Dataset button.

## Changes to the tests

The functional test `MailServiceTest.php` has been updated so that it can provide the `sendHTMLEmailWithAttachment()` function in `MailService.php` with a filename for the attachment.

A new acceptance test has been added to `DatasetUpload.feature` to ensure that the `Upload New Dataset` button remains disabled even when the Terms & Conditions checkbox has been checked.

### DIFF
--- a/protected/components/MailService.php
+++ b/protected/components/MailService.php
@@ -70,9 +70,10 @@ class MailService extends CApplicationComponent
      * @param string $subject email's subject
      * @param string $content content to send
      * @param string $filepath path to file attachment
+     * @param string $attachmentFileName name for the file attachment
      * @return bool whether sending the email is successful or not
      */
-    public function sendHTMLEmailWithAttachment(string $from, string $to, string $subject, string $content, string $filepath)
+    public function sendHTMLEmailWithAttachment(string $from, string $to, string $subject, string $content, string $filepath, string $attachmentFileName)
     {
         return $this->mailer->compose('template',
             ['top_img' => '/var/www/images/email/top.png',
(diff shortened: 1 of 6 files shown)


kencho51 (approved):
Hi @pli888,

The tests are passing locally and also in my gitlab pipeline. I could see that `Update New Dataset` button is disable by default until an excel file is attached.

Besides I may not have the authority to receive email from database@gigasciencejournal.com, so I could not see changes of the file name in the email.

Other than that,  the codes look good to me, I would like to approve this PR.

Issue #966, opened 2022-02-11
Search must be able to search `genome`


pli888:
On staging server, searching for some common words like "genome" will return no results.

However on local dev environment, the same search will return the expected results.

So it's most likely something specific to AWS RDS's PotgreSQL configuration

PR #968, opened 2022-02-11
Fix problem where search could not find genome

rija:
# Pull request for issue: #966 

There were words (like genome or machine learning) that the GigaDB search returns no results for even though there should be a lot of matching datasets.

This problem manifests itself on staging environment.
However, on local dev environment the same search returns the expected results.
This shows that the crux of the problem lies with AWS RDS.

Strong clue surfaces when running the following query on both environments: 
```
SHOW default_text_search_config ;
```

results on local dev: 
```
gigadb=# SHOW default_text_search_config ;
 default_text_search_config 
----------------------------
 pg_catalog.english
(1 row)
```

results on RDS AWS:
```
gigadb=> SHOW default_text_search_config ;
 default_text_search_config 
----------------------------
 pg_catalog.simple
(1 row)
```

It shows that AWS RDS is using a "simple" dictionary instead of the "english" with which the GIN indexes were created with and the search queries were made with.

RDS AWS doesn't allow changing the default value system wide, but we can change the default for a specific  role/database combination with this query: 
```
alter role gigadb in database gigadb SET default_text_search_config TO 'pg_catalog.english';
```

and this can be made into a Yii migration.

* protected/migrations/schema/m220127_150844_search_materialized_views_gin_indexes.php

### DIFF
--- a/protected/components/DatabaseSearch.php
+++ b/protected/components/DatabaseSearch.php
@@ -69,6 +69,14 @@ class DatabaseSearch extends CApplicationComponent {
 		    return $command->join("dataset_type dt","d.id = dt.dataset_id")
                 ->join("type t", "dt.type_id=t.id")
                 ->where(array('in', 't.name', $types))
+                ->andWhere("d.upload_status = 'Published'")
+                ->queryAll();
+        }
+
+		if($author_id) {
+		    return $command->join("dataset_author da","d.id = da.dataset_id")
+                ->where('da.author_id=:id', array(':id'=>$author_id))
+                ->andWhere("d.upload_status = 'Published'")
                 ->queryAll();
         }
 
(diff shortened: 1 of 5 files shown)


pli888 (approved):
The new code in the migration script looks good to me and I confirm it works on my personal staging server by searching for `genome` on the GigaDB website. I approve this PR.

rija:
Hi @pli888, @kencho51

I've discovered three issues with the search.

* It cannot find any results when searching a keyword that's only in a dataset's dataset type (e.g: epigenomic)
* the links to dataset types on the home page weren't working
* the links on author names in either search results or on dataset page were not working

The last push of 3 commits I've just made to this branch should fix all those issues (and includes corresponding acceptance scenarios)

kencho51 (approved):
Hi @rija,

The updated acceptance test is passing locally and also in my gitlab pipeline.
All scenarios you mentioned have been tested manually in dev and staging gigadb website, they are working as expected.

Side note, I have to destroy all resources to get rid of `network kencho-gigadb-website_back-office is ambiguous (2 matches found based on name)` problem, then this PR could be re-deployed to staging, but it may not be related to this PR. And I do not need to `sd_reset_migration` before `sd_gigadb` as mentioned by @pli888 .

All code looks good to me, I approve this PR.

pli888:
>I've discovered three issues with the search.

I confirm these 3 issues are now fixed on my personal staging server. I approve the new changes.

PR #970, opened 2022-02-22
Fix to out of shared memory/disk space errors when running Yii migrations for large database

rija:
This PR is to fix issues with running database migrations on large database (like the ``production_like`` test data or the CNGB database import).
The problem manifested in two situations:

(1) On my local development environment, when running ``setup_devdb.sh`` against ``production_like`` test data

(2) On staging environment, after triggering "sd_reset_migration" manual gitlab job and then triggering a new deployment

In both cases, the symptom are that the process will take forever, won't finish and if we wait, will crash due to exhausted resources (no more memory, or no more disk space).

The root cause are the database triggers, they need to be disabled before running migrations, as every insert from the migration script will trigger them, and to make it worse, the operation they perform is an expensive one.

Database constraints and indexes also need to be disabled, it was already the case in the gitlab production deployment job, but not in the local/CI environment that uses ``setup_devdb.sh``.

One challenge to overcome on local development environment is that because we have to run the data migrations after the schema migrations have run 
and the disabling need to happen before running the data migrations, we cannot rely on schema migrations to restore the disabled structure like we do 
on gitlab production deployment job.

The solution is to pre-build a list of drop queries and add queries before starting migrations.
Then run the drop queries just before running the data migrations.
and run the add queries after running the data migrations.

Why you cannot just dynamically re-add the constraints, indexes and triggers after the data migrations have run (instead of pre-building the queries)
is because once they are dropped, they don't exist anymore, and we wouldn't have an easy way to know what structure to re-create.
That's why both the list of drop queries and add queries are created beforehand while we still have information about them.

>Aside: While investigating, I've stumbled on something that can be a suggestion for future improvement (it's not in scope with this PR):
>In the ``csv_yii_migration.js`` and its ``ops/configuration/yii-conf/migration.php.dist`` template, 
replacing individual ``INSERT`` statements  with  a bulk ``COPY ... FROM '.../myfile.csv' DELIMITERS ','`` 
will improve import performance by **orders of magnitude** (the challenge there would be ensuring that the test data CSV files are visible to the database engine)

* ops/deployment/docker-compose.yml

### DIFF
--- a/ops/deployment/docker-compose.yml
+++ b/ops/deployment/docker-compose.yml
@@ -127,7 +127,7 @@ services:
       - ${DATA_SAVE_PATH}/postgres/${POSTGRES_VERSION}/data:/var/lib/postgresql/data
       - ${APPLICATION}/fuw/app/common/config/bootstrap.sql:/docker-entrypoint-initdb.d/3-fuw.sql
       - ${APPLICATION}/ops/configuration/postgresql-conf/pg_hba.conf:/etc/postgresql/pg_hba.conf
-    command: postgres -c 'hba_file=/etc/postgresql/pg_hba.conf' -c 'stats_temp_directory=/tmp' -c 'log_statement=none' -c 'log_min_duration_statement=1000ms' -c 'max_locks_per_transaction=1024'
+    command: postgres -c 'hba_file=/etc/postgresql/pg_hba.conf' -c 'stats_temp_directory=/tmp' -c 'log_statement=none' -c 'log_min_duration_statement=1000ms'
     networks:
       - db-tier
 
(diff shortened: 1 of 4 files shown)


pli888 (approved):
I'm not able to reproduce the out of memory or disk problems on my local environment, probably because I have ample memory and disk space on my development machine. I also cannot replicate the problem on my staging server after triggering a `sd_reset_migration` job followed by a new deployment.

However, after reading about the use of database triggers after insert statements for creating materalised views (which I believe are used for the database search functionality), I understand how they can be the cause of memory and disk problems and therefore need to be dropped and re-added. Since the code changes look good to me and all tests are passing, I approve this PR.

kencho51 (approved):
Hi @rija,

### Before applying your fix
I could reproduce this error `Out of memory: 7 ERROR:  out of shared memory` in my local dev environment when I tried to `./ops/scripts/setup_devdb.sh production_like`.  But for the staging case, I could still `sd_gigadb` even after `sd_reset_migration`.

### After applying your fix
I could set up local database against `production_like` using `./ops/scripts/setup_devdb.sh production_like`, everything works as expected, besides it takes ~5 mins to finish.
This fix also passes my gitlab pipeline, and code looks good to me, so I would like approve it.

Btw,  have you made any changes to `ops/deployment/docker-compose.yml` file? I could not see it, but you have mentioned it this PR description.

rija:
> Btw, have you made any changes to ops/deployment/docker-compose.yml file? I could not see it, but you have mentioned it in this PR description.

Hi @kencho51 

no, that file hasn't changed. 
the reason it was listed is because an early attempt at fixing problem was made in that file (additional configuration of postgres server) but in the end it did not work and i forgot to remove that mention. thanks for spotting.

PR #978, opened 2022-02-25
Update image in admin dataset update form

kencho51:
# Pull request for issue: #964 

This is a pull request for the following functionalities:

<Insert description of the proposed new functionalities and how they can be tested>
This PR is for updating the image in the admin dataset form, as part of the work for https://github.com/gigascience/gigadb-website/issues/964

`Remove image!!!` button allow users to remove image from a dataset associated with custom image.

### Before the fix
A. Dataset with image

image

image
B. Dataset with no image

image

image

### After the fix

image

image

image

image

## Changes to the view
Implemented fix to the misalignment issues reported in #926 by specifying the form to use current stylesheet file `current.less`.

## Changes to the model
Models `Images` and `ImageHaver` have been replaced by `Image` which extends `CActiveRecords`, but these two files are still remained in the codebase which could avoid other other functionality from broken. 

`DatasetKeywordsTest::testItShouldUpdateKeywordsOnCreate1OnManagement` is one of the tests found broken, but it needs further work to find out the root cause.  

## Change to the controller
The `actionCreate` and `actionUpdate`  in the `AdminDatasetContorller`  are updated to allow file upload and record the file path to the table `image` url column.  

## Changes to the tests

<Insert description of changes to the test code>

A new acceptance test `AdminDatasetForm.feature` has been created, to make sure that:

1. Buttons `Remove image!!!` are available for dataset with custom image associated. 
2. Image meta data fields will be displayed when image is uploaded or the dataset is associated with custom image
3. Uploaded image can be previewed
4. Can save image to no image dataset

Issue #964, opened 2022-02-11
Enable current admin controller for uploading images to write image to S3 using FlySystem


pli888:
* [x] 1. Show image  if it exist, next to "Image upload" in the dataset update form #976 
* [x] 2. 1st case (no image, - i.e: generic image or empty url/location column): show button "Upload new image" #978 
* [x] 3. 2nd case (image associated with the dataset): show two buttons, "Replace image", "Remove image" #978 
* [x] 4. Configure flysystem for S3
* [x] 5. Refactor Image model class using Yii2's ActiveRecord #978 
* [x] 6. Uploaded files go to S3 (using flysystem)

rija:
Hi @kencho51, @pli888


Below, I reckon,  is a summary of remaining  work to do (step by step) on dataset form. Let me know whether it makes sense or not

(1) Create dataset context

* [ ] Don't display metadata fields by default (use CSS class to control the visibility and set ``display`` selector to none by default)

if file chooser changes to have a file selected (detect using an ``onChange`` event):
 
* [ ] Display metadata fields (create a  javascript function to turn on the visibility of the fields)

(2) Update dataset context: 

if dataset is  associated with custom image, do: 

* [ ] Show "Remove image" button
* [ ] Show metadata fields filled with value related to custom image

```
if ($dataset->image !== null && "no_image.png" !== $dataset->image->location) { // dataset has a custom image
	// show remove image button
	// Write javascript that will set metadata fields visibility on	
}
```


(3) Processing the removal of image

When curator click on "Remove image" button, 
It makes an AJAX call to the  adminController's ``actionUpdate()`` method and change the dataset' image_id to that of the generic image
and it replaces the custom image displayed in the form with the generic image.
That is done with something like:

```
document.querySelector('image_preview').innerHTML = '<div><img src=....  style="..."></div>';
```



I reckon that's a lot for one person given the number of days left in the sprint.
We should share the work.

rija:
forgot one task, but we can do that at the end:
 * [ ] Switch dataset form workflow to AWS S3 flysystem adapter

rija:
Hi @kencho51

Below's the first patch. Copy it into a file and then from your branch run:
```
git apply --ignore-space-change --ignore-whitespace /path/to/the/file.patch 
```

It's the Yii migration that create the official row in image table for the generic image:

```
diff --git a/protected/migrations/schema/m220309_165454_generic_image.php b/protected/migrations/schema/m220309_165454_generic_image.php
[...]

PR #984, opened 2022-03-15
Preview image in admin dataset update and create forms in staging

kencho51:
# Pull request for issue: #964 

The major fix is in PR #978.
This PR is to update CSP to include img-src `blob:` to allow previewing of uploaded image.


kencho51:
This fix will be merged as part of https://github.com/gigascience/gigadb-website/issues/991.

Issue #933, opened 2022-01-20
Admin pages - Misalignment of Dataset Funder


ChrisArmit:
**Describe the bug**
In the Admin pages, there are misalignments in the Update Dataset Funder page.

**To Reproduce**
1. Go to 'http://staging.gigadb.org/datasetFunder/update/id/22'
2. Look at the webpage. You will see that the underlying table is msialigned.

**Expected behavior**
The items on the Update Dataset Funder page should be aligned such that additional columns do not appear to the right.

**Screenshots**
Screen captures attached of the issue, and also of how the page should look.

**Desktop (please complete the following information):**
 - OS: MacOSX
 - Browser: Firefox
 - Version: 96
 
![funder](https://user-images.githubusercontent.com/33310581/150361090-2bde6965-fcb8-40d1-9b81-d833ff1c65ce.png)

![funder_correct](https://user-images.githubusercontent.com/33310581/150361123-121dbc7a-bc04-4d96-bbec-458baf568a51.png)

Issue #927, opened 2022-01-20
Admin pages - Misalignment of File Attributes


ChrisArmit:
**Describe the bug**
In the Admin pages, there are misalignments in the Manage Files page.

**To Reproduce**
1. Go to 'http://staging.gigadb.org/adminFile/update/id/121404'
2. Scroll to the bottom of the page.
3. Click on 'Edit'
4. Look at the File Attributes on the webpage. You will see that they are misaligned.

**Expected behavior**
The items on the Manage Files page should be aligned such that additional columns do not appear to the right of the table.

**Screenshots**
Rectangle on the attached screen capture highlights the issue.

**Desktop (please complete the following information):**
 - OS: MacOSX
 - Browser: Firefox
 - Version: 96

![file_attributes](https://user-images.githubusercontent.com/33310581/150350373-a33952e2-b443-4641-b337-e19c2ada4f90.png)

Issue #926, opened 2022-01-20
Admin pages - Misalignment of Update Dataset page


ChrisArmit:
**Describe the bug**
In the Admin pages, there are multiple misalignments on the Update Dataset page.

**To Reproduce**
1. Go to 'http://staging.gigadb.org/adminDataset/update/id/458'
2. Look at the webpage.

**Expected behavior**
The items on the Update Dataset page should be aligned such that items are on the same row.

**Screenshots**
Arrows on the attached screen capture highlight the issue.

**Desktop (please complete the following information):**
 - OS: MacOSX
 - Browser: Firefox
 - Version: 96

![dataset_test](https://user-images.githubusercontent.com/33310581/150349168-56834cce-f9fc-4616-b932-dd0709dd203c.png)

PR #997, opened 2022-03-22
Fix misalignment in admin dataset, admin file and dataset funder pages

kencho51:
# Pull request for issue: #926, #927, #933

This PR does not make any update of the files in the less folder, so the changes are only effective specifically to the admin dataset page,  admin file page and admin sample page as described in the reported issues.

For issue #926, since the `forms.less` has been reverted as the whole less folder needs further investigation, so it is re-opened and fixed in this PR.

image

When edit attribute on admin file page, the table size would exceed the page size after clicking the edit button #927,  after updating the style in the files `/views/adminFile/_attr.php` and `/view/adminFile/_form`, this issue could then be fixed.

image

For the text boxes are found not under the title in admin dataset funder page #933, this issue is fixed by adding `flex-direction: column;` to the class `controls`. 

image

### DIFF
--- a/features/dataset-admin.feature
+++ b/features/dataset-admin.feature
@@ -7,39 +7,6 @@ Background:
 	Given Gigadb web site is loaded with "gigadb_testdata.pgdmp" data
 	And default admin user exists
 
-@ok @issue-381
-Scenario: form loading with all necessary fields
-	Given I sign in as an admin
-	When I go to "/adminDataset/update/id/210"
-	Then I should see a form element labelled "Submitter *"
-	And I should see a form element labelled "Curator Id"
-	And I should see a form element labelled "Manuscript Id"
-	And I should see a form element labelled "Upload Status"
-	And I should see a form element labelled "Epigenomic"
-	And I should see a form element labelled "Software"
-	And I should see a form element labelled "Genomic"
-	And I should see a form element labelled "Metadata"
-	And I should see a form element labelled "Dataset Size *"
-	And I should see a form element labelled "datasetImage"
-	And I should see a form element labelled "Image URL"
-	And I should see a form element labelled "Image Source *"
-	And I should see a form element labelled "Image Tag"
-	And I should see a form element labelled "Image License *"
-	And I should see a form element labelled "Image Photographer *"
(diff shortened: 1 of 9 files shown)


rija (approved):
Hi @kencho51,

I have tested this with the three major browsers and passed all automated tests.

To be upfront about it, this repetition of inline styling is very much a CSS anti-pattern (from maintainability, readability, extensibility POV) that in normal circumstances would be a no-no BUT: 

* We need to sort out our CSS design system and libraries integration. I understand that ending having to do this way is partly a result of the problem we have with it.
* We have currently other more pressing work to move forward than this cosmetic (albeit unsightly) change
* Your changes visibly fix the errors reported by the users
* Since relaunching beta.gigadb.org got delayed anyways, it would be nice to the users to include that PR before we make a new attempt

Therefore, on the basis that this is a temporary fix, I approve this PR.

(Actually, I'd be more comfortable if you add a ``TODO`` comment in each modified view file stating that this implementation is to be revisited once we've better integrated our UX libraries with our design system so that we don't forget about it and that PHPStorm can remind us)

Another minor niggle: the "(bytes)" text is slightly offset in the dataset update form for the "Dataset Size *" field. 
But instead of fixing it by CSS, I suggest it reads better to change the field label to be "Dataset Size In Bytes *" (you'll probably need to update the acceptance scenario that test for that field label).

Finally don't forget to rebase from the latest changes in ``develop``.

Thanks

rija:
Hi @kencho51,

Thanks for implementing all my remarks.

However on Firefox and Chrome there is an alignment problem with the the new "Dataset Size In Bytes" field.



It seems it is because the code for the corresponding label and input element are in the section for two-columns rows instead of the section for one-column rows
You can see the boundary with the comment ``<!-- end of row of two columns -->``.
The label and input control are in the wrong side of it. It needs to be below in the same section as title, description, etc...

Also you need to add a ``class="span10"`` attribute to the input field like for all controls in that section.

pli888 (approved):
With the new inline styling, I can see that the layout of the HTML form elements is much improved now. The `TODO` comments will help remind us that the inline styling needs to be removed in the future when we have solved our CCS problem. I approve this PR.

Issue #995, opened 2022-03-21
terraform plan causes unsupported argument errors


pli888:
Executing `$ AWS_PROFILE=Gigadb terraform plan` to deploy an upstream live environment causes the errors below:
```
│ Error: Unsupported argument
│ 
│   on ../../modules/rds-instance/rds-instance.tf line 31, in module "db":
│   31:   name                   = var.gigadb_db_database
│ 
│ An argument named "name" is not expected here.
╵
╷
│ Error: Unsupported argument
│ 
│   on ../../modules/rds-instance/rds-instance.tf line 54, in module "db":
│   54:   final_snapshot_identifier = "snapshot-final-${var.deployment_target}-${var.owner}-${local.tstamp}"
│ 
│ An argument named "final_snapshot_identifier" is not expected here.
```

Looks like this is caused by the use of `terraform-aws-modules/rds/aws` version 4.2.0 in my upstream live environment whereas we have been using version 3.5.0 which does not have this problem. The fix to the above problems are to use `db_name = var.gigadb_db_database` and `final_snapshot_identifier_prefix = "snapshot-final-${var.deployment_target}-${var.owner}-${local.tstamp}”` in `rds-instance.tf`.

rija:
Fixed two years ago with PR #1001, so closing

PR #1001, opened 2022-03-30
Update Terraform scripts to work with latest Terraform AWS module

pli888:
# Pull request for issue: #995 terraform plan causes unsupported argument errors

* Terraform scripts now able to work with latest version of Terraform AWS module

<Insert description of any changes related to the building of the GigaDB application>

### DIFF
--- a/ops/infrastructure/terraform.tf
+++ b/ops/infrastructure/terraform.tf
@@ -229,6 +229,6 @@ module "rds" {
   # of bastion server and ec2_dockerhost instance.
 }
 
-output "rds_instance_address" {
-  value = module.rds.rds_instance_address
-}
+#output "rds_instance_address" {
+#  value = module.rds.rds_instance_address
+#}
--- a/ops/infrastructure/terraform.tf
+++ b/ops/infrastructure/terraform.tf
@@ -158,51 +158,49 @@ output "vpc_database_subnet_group" {
   value = module.vpc.database_subnet_group
 }
 
-
-
 # EC2 instance for hosting Docker Host
-#module "ec2_dockerhost" {
-#  source = "../../modules/aws-instance"
-#
-#  owner = data.external.callerUserName.result.userName
-#  deployment_target = var.deployment_target
-#  key_name = var.key_name
-#  eip_tag_name = "eip-gigadb-${var.deployment_target}-${data.external.callerUserName.result.userName}"
-#  vpc_id = module.vpc.vpc_id
-#  # Locate Dockerhost EC2 instance in public subnet so users can access website 
-#  # container app
-#  public_subnet_id = module.vpc.public_subnets[0]
-#}
+module "ec2_dockerhost" {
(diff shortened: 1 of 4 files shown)


rija (approved):
Hi @pli888,

I was able to perform all the provisioning steps and they all worked fine.

I was able to ssh the bastion the usual way I do with no problems: 

```
$ ssh -i *************.pem centos@13.36.171.229
```

and I could ssh the webapp ec2 instance the usual way I do without problems:

```
$ ssh -i *************.pem -o ProxyCommand='ssh -i *************.pem -vvv -W %h:%p -q centos@13.36.171.229' centos@10.99.0.206
```

All the tests are passing locally and on CI, and the staging deployment ``sd_deploy`` was successful and the website on my staging environment and 
the corresponding portainer dashboard seem to work fine.

Finally ``terraform destroy`` did its job without errors.


I'm therefore happy to approve this PR.

Probably a stupid question:
In ``aws_db_parameter_group.gigadb-db-param-group[0].name``,  why is there ``[0]``?  I would think the DB parameter group resource is one thing, not an array of things.

pli888:
>Probably a stupid question:
In aws_db_parameter_group.gigadb-db-param-group[0].name, why is there [0]? I would think the DB parameter group resource is one thing, not an array of things.

It's because we use the `count` meta-argument as a conditional to create the `aws_db_parameter_group` resource when we are creating a staging environment. The problem is that the `count` value will also create that number of resources, hence we need a `[0]` in `aws_db_parameter_group.gigadb-db-param-group[0].name` to identify which `aws_db_parameter_group.gigadb-db-param-group` we want to use.

ops/infrastructure/terraform.tf
```diff
@@ -132,7 +132,7 @@ module "vpc" {
   # You can enable communication from internet to RDS is via an internet gateway
   # to provide public access to RDS instance, but is not recommended for 
   # production! These parameters are all false so no public access to RDS
-  create_database_subnet_group = false
+  create_database_subnet_group = true
```
kencho51, line 135: Is it better to move `create_database_subnet_group = true` above the comment? As it has the value `true`, while comment mentions that all parameters are flase.

kencho51 (approved):
Hi @pli888,

I have gone through all the steps mentioned in `How to test the new feature` and everything is working well as expected.

The `gigadb_db_host` is accessible on bastion server, and dockerhost server is accessible through the bastion server using `ProxyCommadn` from local dev computer.

And I have tried several rounds of `terraform apply` and `terraform apply --destroy`, DB parameter group can be created and destroyed successfully.

 A suggestion is to move `create_database_subnet_group = true` in terraform.tf to above the comment block, so it would make the content below it more consistent. It is just a minor, cosmetic change.

Other than that, all code looks good to me.

ops/infrastructure/terraform.tf
pli888: This is done now.

Issue #987, opened 2022-03-15
Get spreadsheet import tool into codebase


pli888:
- [x] Add sample spreadsheet for testing spreadsheet upload tool into gigadb/app/tools directory
- [x] Add README file to document how to use the tool into gigadb/app/tools directory
- [x] If required, add gitignore file so log files are not saved into main codebase, etc into gigadb/app/tools directory
- [x] Add dependencies of the tool
- [x] Run unit tests

kencho51:
The tool [repo](https://github.com/jessesiu/gigadb-tools-dataupload).

rija:
More tasks to do before this is considered complete:
* [ ] Ensure tool works on cloud deployment (see PR #1051)
* [ ] Ensure MD5 checksum are written to the file table
* [ ] Ensure file size are written to the file table

rija:
closing as this technically done. There will be a separate issue (#1052) for using on cloud deployment

PR #1004, opened 2022-04-06
Use spreadsheet import tool in local environment

pli888:
# Pull request for issue: #987 Get spreadsheet import tool into codebase

* Download and run the consultant's tool from the `gigadb/app/tools/excel-spreadsheet-uploader` directory.
* Documentation in the README.md file provides instructions for running the tool.

## Changes to the database data

New rows have been added into `species.csv` and `external_link_type.csv` in the `data/dev` directory. The example spreadsheet `100679newversion.xls` requires these new data for database upload.

`Dockerfile` and `docker-compose.yml` provides a running container with open JDK 8 that the consultant's tool uses to compile and execute the Java code with the `run.sh` bash script.


rija:
Hi @pli888, @kencho51,

I've been able to install the tool and successfully replicate the demo given by @pli888 during VOH.

I've also used the xlsx formatted spreadsheet emailed by @only1chunts (also called source spreadsheet by me below)
Before copying it into the Uploaddir, I've converted it to the old format (.xls) using Apple Numbers.
The java tools seems to have processed the spreadsheet in the same way as it did with the default example.

There are some things to point out though (but I don't think they have anything to do with format difference):

* There is an "error" in curator supplied spreadsheet in the ``ftp_site`` field causing malformed location value in the file table which is a missing trailing slash  (and it's the wrong url too as we use https prefixes now). Adding a trailing slash in the source spreadsheet fixes the value in the file table's location column.
* I had to replace GIGA-D-21-00197 with the seemingly assigned identifier 102203 in the identifier column in the source spreadsheet
* the accessions link have odd entries that I didn't see with the default example (but the default example's corresponding input fields are too simple so can't compare):


input from spreadsheet:

Screenshot 2022-04-06 at 6 47 47 PM

result in database (the rows of interest are the ones for dataset_id 701):

Screenshot 2022-04-06 at 6 47 33 PM



Is this normal behaviour  or should we expect a distinct link row for each of the 5 links irrespective of which of the two spreadsheet fields they were from?

rija:
@pli888,

here's the patch to fix the oddity I've mentioned in my last comment:

```
diff --git a/src/Excel2Database.java b/src/Excel2Database.java
index 8fccf2a..7231fc5 100755
--- a/src/Excel2Database.java
+++ b/src/Excel2Database.java
@@ -1,4 +1,4 @@
-<U+FEFF>import java.io.BufferedInputStream;
+import java.io.BufferedInputStream;^M
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -3360,8 +3360,15 @@ public void fillTable_sample_attribute() throws SQLException, IOException{
                String[] attributeList1 = null;
                int size_number1=0;
                if (content1.length() != 0) {
+            if(content1.contains(","))^M
+                       {^M
                        attributeList1 = content1.split(",");
                        size_number1 = attributeList1.length;
+                       }^M
+                       else{^M
+                               attributeList1 = content1.split(";");^M
+                               size_number1 = attributeList1.length;^M
+                       }^M
                }
                
                

```

rija:
Here's the patch that:

* sets semi column as valid boundary between accession link
* removes any spaces inside ext_acc_mirror and ext_acc_link's values

[Excel2Database.txt](https://github.com/gigascience/gigadb-website/files/8442858/Excel2Database.txt)

rija:
Here's the patch that:

    * sets semi column as valid boundary between each accession link
    * removes any spaces inside ext_acc_mirror and ext_acc_link's values
    * removes preceding and trailing white spaces in sample_attributes' values

[Excel2Database.txt](https://github.com/gigascience/gigadb-website/files/8444868/Excel2Database.txt)

rija:
Hi @pli888,

I was thinking, may be the application of the patch can be done in ``run.sh`` before the compilation step by adding a line like: 

```
patch -n Excel2Database.txt
```
(the Dockerfile need an ``apt-get install patch`` added)

So that don't have to host any bit of source code from the other repo.



I also still think the git archive passed to the ``curl``command  should be from a fork of the consultant's repo rather than their repo, so that the patch will always work whatever change happens to the consultant's repo.

rija:
Hi @pli888 

I've made another patch that will allow value of ``ftp_site`` to work whether is has trailing slash or not. (@only1chunts's example doesn't have one and which will cause incorrect file locations before this patch).

patch content: 

* sets semi column as valid boundary between each accession link
* removes any spaces inside ext_acc_mirror and ext_acc_link's values
* removes preceding and trailing white spaces in sample_attributes' values
* If there's no trailing slash at the ned of ftp_site's value, we add one in code

[Excel2Database4.txt](https://github.com/gigascience/gigadb-website/files/8466146/Excel2Database4.txt)

rija (approved):
Hi @pli888,

I've tested the workflow again after you have manually added my last patch to the fork's version of the java code.

I've inputed two spreadsheets at the same time in ``uploadDir``: the default example and a manually converted version of Chris' example.

After running the tool, both datasets were ingested with no issues.
The problems that the patch was made for didn't occur in the resulting data.

So the tool appears to work fine. On that basis I'm happy to approve this PR.

I know acceptance tests are still work in progress, but:

* the default example does not give enough coverage as e.g all the behaviours fixed by the patches come from another example
* One curator-supplied example, I doubt is enough to test all possible error modes, so there's likely more patch needed as curators start sending more files to the tool
* We will learn more the sooner we get the tool in curator's hands rather than try to second-guess all possible error modes

So on that basis, I think it's fine to defer completing implementation of acceptance tests for later

kencho51:
Hi @pli888,

I have found the way to apply the patch and hence created a PR https://github.com/pli888/gigadb-website/pull/36 for it.

After applying the patch file, the manually converted `GigaDBUpload_102203_GIGA-D-21-00197_hamster.xls` could be uploaded to the dev `gigadb` database. The newly added attributes could visually confirmed by logging into the dev website as an admin, and the upload process could be confirmed by checking`**End success: GigaDBUpload_102203_GIGA-D-21-00197_hamster.xls` in the `java.log`.

And to confirm that this import tool is able to upload more the 1 xls at the same time, as indicated in the `java.log` as 
```
**Begin: file 1 : 100679newversion.xls in process...
.
.
.
execution time: 3762
**End success: 100679newversion.xls
**Begin: file 2 : GigaDBUpload_102203_GIGA-D-21-00197_hamster.xls in process
.
.
.
execution time: 1469
**End success: GigaDBUpload_102203_GIGA-D-21-00197_hamster.xls
```

The code looks good to me, so I am happy approve this PR.

pli888:
>I have found the way to apply the patch and hence created a PR https://github.com/pli888/gigadb-website/pull/36 for it.

Thanks @kencho51 but the code changes in the patch in your PR were already made to [Excel2Database.java](https://github.com/gigascience/ExceltoGigaDB/blob/develop/src/Excel2Database.java) last week.

pli888:
>So on that basis, I think it's fine to defer completing implementation of acceptance tests for later

@rija Ok, let's get this PR merged. I suggest we fix the bugs I demoed last week re: creating private dataset pages first before the work on the acceptance tests for Excel spreadsheet upload is completed.

Issue #1024, opened 2022-04-19
Time out on asset-packagist.org


rija:
when running ``composer install`` or ``composer update``,
manually or on CI, there is a time out error on https://asset-packagist.org

PR #1025, opened 2022-04-19
Fix asset-packagist.org time out with Yii2 composer dependencies

rija:
# Pull request for issue: #1024

when running composer install or composer update,
manually or on CI, there is a time out error on https://asset-packagist.org/ [1]

In addition to the need of replacing`` asset-packagist.org`` with ``repo.packagist.org`` in ``composer.json`` [2],
we need to install ``yidas/yii2-composer-bower-skip`` composer library so Yii2 dependencies install/update don't throw errors (which is ok for us as we don't use that part of Yii2)

See: 

[1] hiqdev/asset-packagist/issues/146
[2] https://getcomposer.org/doc/05-repositories.md#composer

## How to test

Running the following commands should not yield any errors:

```
$ docker-compose run --rm config
$ docker-compose run --rm test composer update 
$ docker-compose exec -T console bash -c 'composer update'
$ docker-compose exec -T console bash -c 'cd /gigadb-apps/worker/file-worker/ && composer update'
````

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c0a56ae0e4a3a60e34c1b4e044d3826d",
+    "content-hash": "6093427da7b45ece562639bbd5ff5209",
     "packages": [
         {
             "name": "aws/aws-crt-php",
@@ -58,16 +58,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.212.1",
+            "version": "3.219.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
-                "reference": "e4a5f8dca671281e8e2122607a4da2a8348489bf"
+                "reference": "7e8da5b45d545ca3129a14e972001698212b1a00"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/aws/aws-sdk-php/zipball/e4a5f8dca671281e8e2122607a4da2a8348489bf",
-                "reference": "e4a5f8dca671281e8e2122607a4da2a8348489bf",
+                "url": "https://api.github.com/repos/aws/aws-sdk-php/zipball/7e8da5b45d545ca3129a14e972001698212b1a00",
+                "reference": "7e8da5b45d545ca3129a14e972001698212b1a00",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by kencho51 — APPROVED** [review]
Hi @rija,

All steps in `How to test` section are passing locally and `unit_func_tests` is also passing in gitlab pipeline.
But, after all the steps, it will end up with 3 modified `composer.lock` files as below,
```
modified:   composer.lock
modified:   fuw/app/composer.lock
modified:   gigadb/app/worker/file-worker/composer.lock
```
Should we commit and push these changes or restore them to discard them in working directory?

**Comment by rija** [comment]
@kencho51,


Regarding the lock files. At the moment we should follow this guideline:

* (Optionally) commit them deliberately after checking the nature of the changes  as part of pushing to any ready PR (and the reviewers would have the right to veto it if an update is problematic to them)
* Discard them in any other situations

The reasons is that change to lock file means in general (sometimes it's a change of cloning/transfer protocol) 
some dependencies have had a minor update.

We don't want these minor update to end up being the new baseline for the team without some vetting (in case of bug introduction or supply chain attack)
and any pull request is an opportunity to do that check.
Also given the large amount of dependencies, the lock file will gets changed quite frequently and we don't want to make specific PR every time this happen (too time consuming), and we don't want to commit them directly to develop either.

At the same time it is beneficial to keep that ability of composer to continue updating minor versions to ensure we are not potentially blocked by an issue in a given dependency for which a fix is already released.  That latter point we might want to move away from in the future when we have a bigger team and are better at keeping up-to-date with dependencies changes and monitoring supply chain, in which case we could then use only exact versions in composer.json and update composer.json as needed and the lock file won't get updated at all until composer.json changes again. 


Leading by example, I've now pushed the lock files as part of this PR, make sure everything still works

**Comment by pli888** [comment]
@rija @kencho51 Did either of you get an error with passing the SampleTest unit test? I see the error on my local environment and in my CI/CD pipeline:
```
There was 1 error:

---------
1) SampleTest: It should return sample attribute array map
 Test  tests/unit/SampleTest.php:testItShouldReturnSampleAttributeArrayMap
                                                                                                                                                                                      
  [TypeError] Argument 1 passed to Attribute::__construct() must be of the type int, null given, called in /var/www/vendor/yiisoft/yii/framework/db/ar/CActiveRecord.php on line 397  
                                                                                                                                                                                      
#1  /var/www/vendor/symfony/polyfill-php80/Resources/stubs/Attribute.php:18
#2  /var/www/vendor/yiisoft/yii/framework/db/ar/CActiveRecord.php:397
#3  /var/www/vendor/yiisoft/yii/framework/test/CDbFixtureManager.php:306
#4  /var/www/vendor/yiisoft/yii/framework/test/CDbTestCase.php:118

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

kencho51:
Hi @pli888,

> 
> ```
> There was 1 error:
> 
> ---------
> 1) SampleTest: It should return sample attribute array map
>  Test  tests/unit/SampleTest.php:testItShouldReturnSampleAttributeArrayMap
>                                                                                                                                                                                       
> ```

`SampleTest` is passing  on my local environment and in gitlab pipeline.

pli888 (approved):
All tests are now passing on my dev environment and in my CI/CD pipeline. I approve this PR.

SilverFire:
Asset-packagist is up, sorry for the troubles. You can find an explanation in the original issue: https://github.com/hiqdev/asset-packagist/issues/146#issuecomment-1106742011

Issue #1023, opened 2022-04-19
Creation of private dataset pages


pli888:
## Describe the bug

Clicking the create private dataset button in the dataset update admin page does not lead the user to the new private dataset page. The correct behaviour should be that when the button is pressed then the user will see the new private dataset page which has a URL that has a private page token.

PR #1026, opened 2022-04-20
Create private dataset page

kencho51:
# Pull request for issue: #1023

1. A new dataset `100039` with dataset type `Incomplete` is introduced.
2. Upload status `Incomplete` has been included in the admin dataset page.
3. Dataset with upload status `AuthorReview` can be created with private URL.
4. `Create/Reset Private URL` is enable for the  page type `hidden` and `draft` datasets.
5. The private URL will contain `/dataset/DOI/token/$token`. 

## How to test?
1. Dataset page `http://gigadb.gigasciencejournal.com:9170/dataset/100039` will not be working because it is not a public page.
2. Go to `http://gigadb.gigasciencejournal.com:9170/adminDataset/update/id/5`,  you should see upload status is `Incomplete`, and `Create/Reset Private URL` button is available at the bottom. When you click `Create/Reset Private URL` button, you will be redirected to, for example, `http://gigadb.gigasciencejournal.com:9170/dataset/100039/token/LN9aWzJ_y7FpAWv7`.
If you go back to `http://gigadb.gigasciencejournal.com:9170/adminDataset/update/id/5`, you should see `Open Private URL` link is available as well.
3. `Create Dataset` in `http://gigadb.gigasciencejournal.com:9170/adminDataset/admin` page, after filling in all mandatory fields, dataset with upload status `AuthorReview` will be created and redirected to private URL, and buttons `Create/Reset Private URL` and `Open Private URL` will then be available for that dataset.

In `protected/controllers/AdminDatasetController.php`, `hidden` and `draft` page type are explicitly  included  in the `actionPrivate`,  upload status `Incomplete` belongs to hidden page, while `AuthorReview` belongs to draft page, Create/Reset Private URL` button can be triggered and then create a private url with random 16 chars token. Finally, the page will be redirected to `/dataset/DOI/token/$token`.

In `protected/views/adminDataset/_form.php`, make `Create/Reset Private URL` and `Open Private URL` buttons available for creating new `AuthorReview` dataset.

None.

A new test dataset is needed, if any current dataset in `data/dev` changes its upload status, the related scenarios need to be updated, this will shrink the code coverage.
So a new dataset `100039` is introduced to `data/dev` with its upload status is `Incomplete`. Therefore, when creating acceptance tests, current passing scenarios will not be affected.

None.

A draft PR was created to remove extra template files #1027, now the fix is attached to this PR.

Before the fix, unnecessary excel files will be downloaded to the file systems and they need to be removed manually.

In ` tests/functional/DownloadTemplateFilesCest.php`,  an `after` function is added to check for the existence of the excel file and then remove it after each test, so the state of the file systems could remain as before the test.

None.


tests/acceptance/AdminDatasetForm.feature
```diff
@@ -129,5 +129,37 @@ Feature: form to update dataset details
...
+  @ok @issue-1023
+  Scenario: Open private url is working
+    When I am on "/adminDataset/update/id/22"
+    And I press the button "Create/Reset Private URL"
+    And I wait "1" seconds
+    Then I am on "/adminDataset/update/id/22"
+    And I should see "Open Private URL"
+    And I follow "Open Private URL"
+    And I should see current url contains "/dataset/100020/token/"
+    And I should see "Genome data from foxtail millet (<em>Setaria italica</em>)."
```
pli888, line 161: The `I follow "Open Private URL"` step is a `When` event which is then followed by an expected outcome, `I should see current url contains "/dataset/100020/token/"`. Therefore, I think it reads better this way:
```
When I am on "/adminDataset/update/id/22"
And I press the button "Create/Reset Private URL"
And I wait "1" seconds
And I am on "/adminDataset/update/id/22"
And I follow "Open Private URL"
Then I should see current url contains "/dataset/100020/token/"
And I should see "Genome data from foxtail millet (<em>Setaria 
```

I think the `And I should see "Open Private URL"` step can be deleted because the `I follow "Open Private URL"` step implies that you should see the `Open Private URL` link.

tests/acceptance/Search.feature
```diff
@@ -27,22 +27,22 @@ Feature: main search function
...
-    And I should not see "Data and software to accompany the paper: Applying compressed sensing to genome-wide association studies."
+#  @ok
+#  Scenario: Can navigate to the next page
+#    Given I am on "/"
+#    And I fill in the field of "id" "keyword" with "genome"
+#    And I press the button "Search"
+#    And I wait "1" seconds
+#    When I follow "2"
+#    Then I should see a link "Genomic data from Adelie penguin (<em>Pygoscelis adeliae</em>)." to "/dataset/100006"
+#    And I should not see "Data and software to accompany the paper: Applying compressed sensing to genome-wide association studies."
```
pli888, line 45: If we delete this acceptance test then this means we are reducing the test coverage when we should really be trying to increase it. Can we use another dataset instead of foxtail millet for the tests in `AdminDatasetForm.feature` so it does not impact this test? Maybe we need to add a new dataset into the `dev` environment for the new tests in `AdminDatasetForm.feature`?

tests/acceptance/Search.feature
```diff
@@ -168,13 +165,13 @@ Feature: main search function
...
-    And I press the button "Search"
-    And I wait "1" seconds
-    Then I should see a link "Genome data from foxtail millet (<em>Setaria italica</em>)." to "/dataset/100020"
+#  @ok
+#  Scenario: Search for a term that is only in dataset types
+#    Given I am on "/"
+#    And I fill in the field of "id" "keyword" with "epigenomic"
+#    And I press the button "Search"
+#    And I wait "1" seconds
+#    Then I should see a link "Genome data from foxtail millet (<em>Setaria italica</em>)." to "/dataset/100020"
```
pli888, line 174: Deleting this test means we are reducing test coverage?

pli888 (commented):
I think the fix for creating private dataset pages is fine -         basically, a private URL can be created for a non-published or non-submitted dataset by updating its upload_status value to `Incomplete` or `Private`. The private dataset URL will end with a token provided as a random string 16 characters in length which is saved into the `token` column for the dataset record in dataset table.

My concerns are the acceptance tests that have been commented out as this will reduce test coverage when we should be trying to increase test coverage. I've made a couple of suggestions in the code on how to address this.

tests/acceptance/Search.feature
kencho51: A new dataset `100039` has been added into `dev` environment.

tests/acceptance/Search.feature
kencho51: Since new dataset `10039` is included, this commented scenario has been reverted.

data/dev/dataset.csv
```diff
@@ -9,4 +9,5 @@ The assembled scaffolds of high quality sequences total 1.23 Gb, with the contig
...
...semble contigs derived from the short fragment library. The Illumina paired-end and mate-pairs reads were assembled together with Ray (<a href=""http://denovoassembler.sourceforge.net"">http://denovoassembler.sourceforge.net</a>), with the k-mer defined iteratively. In total, given that the genome size is predicted to be 1.58Gb, with the total scaffold length of 1,184, 594,388 bp, the overall coverage of the genome is around 76%, a value that might be slightly overestimated given that some of the scaffolds may be overlapping but could not be assembled. Filtering followed by assembly resulted in 259,423 contigs (N50=6,983 bp, longest = 75,003 bp), which was further combined into 148,255 scaffolds (N50 = 19,470, longest = 206,462 bp). The database contains all of the contigs, scaffolds, corresponding assembly parameters, and the annotations for the known repeats and coding sequences. The assembled scaffolds allow basic genomic annotation and comparative analyses with other available avian whole-genome sequences.",4294967296,ftp://climb.genomics.cn/pub/10.5524/100001_101000/100039/,Incomplete,GigaDBSubmissionForm_Parrot.xls,1586ed77abdec5b9f839be0f5d61f357,2012-09-11,2012-09-14,1,,,,,
```
pli888: I think this record should be moved to the top so the records are in order of their ids.

data/dev/gigadb_user.csv
```diff
@@ -7,3 +7,4 @@ id,email,password,first_name,last_name,affiliation,role,is_activated,newsletter,
 336,test+336@gigasciencejournal.com,5a4f75053077a32e681f81daa8792f95,Hugh,Shanahan,"Department of Computer Science, Royal Holloway, University of London, Egham, TW20 0EX, UK","",f,t,f,,,,,test+336@gigasciencejournal.com,,EBI
 ,admin@gigadb.org,5a4f75053077a32e681f81daa8792f95,Joe,Bloggs,BGI,admin,true,false,true,,,,,admin@gigadb.org,,EBI
 ,user@gigadb.org,5a4f75053077a32e681f81daa8792f95,John,Smith,BGI,user,true,false,true,,,,,user@gigadb.org,,EBI
+345,test+5@gigasciencejournal.com,5a4f75053077a32e681f81daa8792f95,John,Smith,BGI,user,t,f,t,,,,,test+5@gigasciencejournal.com,EBI
```
pli888: This user record should be moved to before the admin@gigadb.org record. Also fix the email address so it is consistent with the id.

data/dev/image.csv
```diff
@@ -4,4 +4,5 @@ id,location,tag,url,license,photographer,source
 147,"",CS icon1,"",public domain,Shashaank Vattikuti ,Gigascience
 212,100142.jpg,Overlap between different protocol search terms,http://gigadb.org/images/data/cropped/100142.jpg,CC0,Alnasir and Shanahan 2015,Alnasir and Shanahan 2015
 669,"",no image icon,var/www/hosts/localhost/htdocs/images/uploads/image_upload/Images_1885.png,Public domain,GigaDB,GigaDB
-700,no_image.png,no image icon,"",Public domain,GigaDB,GigaDB
\ No newline at end of file
+700,no_image.png,no image icon,"",Public domain,GigaDB,GigaDB
+5,100039_Amazona_vittata.jpg,Puerto Rican parrot,http://gigadb.org/images/data/cropped/100039_parrot.jpg,CC0,Taras Oleksyk,"University of Puerto Rico at Mayagüez, PUERTO RICO"
```
pli888: This image record should be moved higher so the ids are in order I think.

tests/acceptance/AdminDatasetForm.feature
```diff
@@ -129,5 +129,36 @@ Feature: form to update dataset details
...
+    And I should see "(2012): Genomic data of the Puerto Rican Parrot (<em>Amazona vittata</em>) from a locally funded project. GigaScience."
+
+  @ok @issue-1023
+  Scenario: Open private url is working
+    When I am on "/adminDataset/update/id/5"
+    And I press the button "Create/Reset Private URL"
+    And I wait "1" seconds
+    Then I am on "/adminDataset/update/id/5"
+    And I follow "Open Private URL"
+    And I should see current url contains "/dataset/100039/token/"
```
pli888: I think this step should be the `Then` step.

pli888 (changes requested):
@kencho51 Can you check the appearance of the private URL dataset page please? Mine looks like this:
Screenshot 2022-05-25 at 10 11 46 AM

The purpose of the private dataset page is to check that it looks ok before going live. Some of the information is missing from the private dataset page so I'm wondering what is wrong.

Also, I don't think `gigadb/app/worker/file-worker/composer.lock` should be a part of this pull request.

kencho51:
Hi @pli888 

> Can you check the appearance of the private URL dataset page please? 
> 
> The purpose of the private dataset page is to check that it looks ok before going live. Some of the information is missing from the private dataset page so I'm wondering what is wrong.

The private dataset `100039` was initially created with just minimum information for passing the `Create/Reset Private URL` test. Now all information of `100039` has been supplied.

> Also, I don't think `gigadb/app/worker/file-worker/composer.lock` should be a part of this pull request.

The composer.lock file has been removed.

data/dev/sample_attribute.csv
```diff
@@ -1,4 +1,5 @@
 id,sample_id,attribute_id,value,unit_id
+25158,151,200,female,
```
pli888: This record should be placed lower down so the ids are listed in order.

pli888:
> The composer.lock file has been removed.

@kencho51 When you look in the `develop` branch, you will see that it has a `gigadb/app/worker/file-worker/composer.lock` file. Instead of deleting it in your feature branch, you should have reverted this file back to the state it was in the `develop` branch.

Also, check if you need to rebase your feature branch with the latest code in the upstream develop branch.

rija (commented):
Hi @kencho51,

The primary functionality works as expected.
I noticed you've fixed the spurious excel files leftover from functional tests. The code that does that looks fine to me too.

Locally, the unit and functional tests passes. 
Acceptance tests won't run properly for me locally until you rebase with ``develop``'s tip to get the fix for Apple Silicon based Mac.

I've tried on my old Intel based Mac, and the acceptance test run is failing for ``features/dataset-admin.feature:51``

I tried manually to create new dataset by filling just mandatory fields like the scenario would do.
And I get a blank page as result (which match the scenario failure mode)

Looking at the log, nothing wrong there, and the dataset is added to the database correctly.

The problem is that your change to ``AdminDatasetController.php`` has moved instructions that was always called, into an ``elseif`` block
that isn't selected when creating a new dataset from admin form. (And you may have not noticed that the new dataset creation will redirect to the private action method).

When a new dataset is created, it is assigned the "AuthorReview" and for this type, instantiation of ``DatasetPageSettings`` will select the "draft" type, not the "hidden" one (as to why ? It was part of File Upload Wizard worflow's implementation which is not entirely correct).

A simple fix is to modify the ``elseif()`` block you have added to select both "hidden" and "draft" internal types. 

That's a good approach because if you look at the database record created for a new dataset using the form, no token is created. So you'd need to create one anyway. (That seems like a bug which existed before your changes that would be fixed as a bonus) so that block would create the token and redirect appropriately.

An alternative approach,  is to ensure new dataset creation through the form also use the "Incomplete" upload_status instead of "AuthorReview"

The drawback with that approach, is we need to be extra conscientious with running and augmenting the tests in case they are side effects, as "AuthorReview" was implemented as part of the File Upload Wizard workflow.

However that latter approach seems the most correct one (I think there should not be different statuses when a dataset is created by form or by bulk upload), but we would need later ask confirmation from Chris H as to which status is correct (it is related to both File Upload Wizard workflow - which is not entirely correct on that matter - and River Valley integration workflow which revamps most of the statuses).

Anyways, because the former approach is simpler, as less risk of side effects and does fix another existing bug, I think you should implement that for now. When we get to demo that functionality we can ask Chris H about those statuses and schedule a proper refactoring of the confusing DatasetPageSettings/actionPrivate() code.

tests/acceptance/AdminDatasetForm.feature
```diff
@@ -126,8 +126,72 @@ Feature: form to update dataset details
     And I fill in the field of "name" "Dataset[ftp_site]" with "ftp://test"
     And I fill in the field of "name" "Dataset[title]" with "test dataset"
     And I press the button "Create"
-    Then I am on "dataset/view/id/400789"
+    Then I wait "1" seconds
```
pli888, line 129: I see there is no step here anymore to check the new redirect to `/dataset/'.$model->identifier.'/token/'.$model->token` that you have defined in `AdminDatasetController.php`. You could add a new function in `AcceptanceTester.php` that uses `seeCurrentUrlMatches` from Codeception and a regular expression to test this.

pli888 (approved):
@kencho51 The create private dataset button and create new database functionalities are working for me. Also, all acceptance, functional and unit tests are passing locally. I approve this PR.

Have a think about improving `tests/acceptance/AdminDatasetForm.feature` though as described in my comment.

rija (approved):
Hi @kencho51

All tests are now passing for me.

I concur with @pli888's observation regarding scenario "Can create dataset with image" of ``AdminDatasetForm.feature``.

Especially odd that the step that's missing is present for the scenario "Open Private URL from AuthorReview dataset".

However given that the scenario "new dataset with mandatory fields filled in" from the legacy ``dataset-admin.feature`` has the step performing the missing check
we are not really loosing coverage. 
It's a consistency issue. Also the legacy test suites need to be retired at some point, so we should avoid relying on the coverage they provide
We can do that whenever implementing a story or repaying tech debt.

Thus, since coverage is not that much impacted, the rest of code is fine and functionality works, I'm happy to approve this PR.

 
From a bigger picture perspective, I believe we now have all the pieces in place for dataset spreadsheet workflow:
* master template of the spreadsheet layout is in ``develop``
* the spreadsheet uploader tool can ingest spreadsheet based on the template, that's on ``develop`` too
* we have a file urls checker tool to check the file urls are correct and that's on ``develop`` as well
* we can preview the dataset resulting from the spreadsheet and that was the last piece.

Next step is to schedule for next sprint the writing and testing of an SOP for performing the job.

tests/acceptance/AdminDatasetForm.feature
kencho51: A step `I should see current url contains` has been added to this scenario now.

PR #1027, opened 2022-04-21
TestDebt: To remove excel files after a functional test

kencho51:
# Pull request for functional test debt

Before the fix, every time after running

```
% ./tests/unit_functional_runner
OR
% docker-compose run --rm test ./vendor/codeception/codeception/codecept run functional tests/functional/DownloadTemplateFilesCest.php

```
unnecessary excel files will be downloaded to the file systems and they need to be removed manually.

After the fix, an `after` function is added to check for the existence of the excel file and then remove it after each test, so the state of the file systems could remain as before the test.

### DIFF
--- a/tests/functional/DownloadTemplateFilesCest.php
+++ b/tests/functional/DownloadTemplateFilesCest.php
@@ -6,6 +6,17 @@ class DownloadTemplateFilesCest
     {
     }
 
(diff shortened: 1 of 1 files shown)


kencho51:
Close this PR because it is included in #1026.

Issue #1031, opened 2022-04-25
spreadsheet tool workflow has stopped checking whether files'url resolve to a successful file download


rija:
Since tech team took over the dataset spreadsheet uploading tool, the systematic check of files url for the files added from the spreadsheets is not performed

PR #1033, opened 2022-04-26
Add command to check files url for a dataset

rija:
# Pull request for issue: #1031

The command will make a HEAD request to the urls of files associated to a dataset and output thoses that
don't have a 200 response code or have an ftp scheme.

To test run the following command:

```
$ docker-compose exec application  ./protected/yiic files checkUrls --doi=<DOI>
```

where <DOI> is replaced by a DOI for which to test the files' url.

### DIFF
--- a/protected/commands/FilesCommand.php
+++ b/protected/commands/FilesCommand.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Command to check fils url
+ *
+ * @author Rija Menage <rija+git@cinecinetique.com>
+ * @license GPL-3.0
+ */
+
+
+class FilesCommand extends CConsoleCommand
+{
+
+    public function getHelp()
+    {
+        $helpText = "checks files url for a specific dataset in the database" . PHP_EOL;
+        $helpText .= "Usage: ./protected/yiic files checkUrls --doi=<DOI>" . PHP_EOL;
+
+        return $helpText;
+    }
+
+    /**
+     * @return int
+     * @throws CException
+     */
+    public function actionCheckUrls($doi) {
+        $sql =<<<END
+SELECT f.location
+FROM file f, dataset d
+WHERE f.dataset_id = d.id and d.identifier = '$doi';
+END;
+
+        try {
+            $rows = Yii::app()->db->createCommand($sql)->queryAll();
+            foreach ($rows as $row) {
+                $parts = parse_url($row['location']);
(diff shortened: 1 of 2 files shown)


protected/commands/FilesCommand.php
```diff
@@ -0,0 +1,68 @@
...
+        try {
+            $rows = Yii::app()->db->createCommand($sql)->queryAll();
+            foreach ($rows as $row) {
+                $parts = parse_url($row['location']);
+                if ( "ftp" === $parts['scheme']) {
+                    echo $row['location'].PHP_EOL;
+                    continue;
+                }
+                $headers = get_headers($row['location'], self::RETURN_ASSOCIATIVE_ARRAY, stream_context_create(array('http' => array('method' => 'HEAD'))));
+                if ($headers[0] !== HttpCode::OK)
```
pli888, line 52: The expected behaviour is that if a file URL resolves OK then it will not be outputted. However, resolvable file URLs are printed in my terminal when I run the tool, e.g. `https://ftp.cngb.org/pub/gigadb/pub/10.5524/100001_101000/100020/readme.txt` when I run this command against the `dev` GigaDB website:
```
$ docker-compose exec application  ./protected/yiic files checkUrls --doi=100020
```

Maybe the problem is because the value of `HttpCode::OK` is 200 which is not equal to the `HTTP/1.1 200 OK` value of `$headers[0]` for a resolvable URL. The `!==` comparison operator will therefore always output `true` even if the file URL is resolvable.

rija: Thanks @pli888, your observation was correct and  the last commit should have fixed the problem.

pli888 (approved):
The code looks good to me and I confirm the new functional tests are passing locally and on my CI/CD pipeline. I approve this PR.

rija:
Thanks @pli888,

I've just realised there was a missing test coverage for the special case of ftp links, so I made one last commit for an update to FilesCommandCest.php that add a test for those links.

pli888:
>I've just realised there was a missing test coverage for the special case of ftp links, so I made one last commit for an update to FilesCommandCest.php that add a test for those links.

@rija Ok, the new functional test method passed using:
```
$ docker-compose run --rm test ./vendor/codeception/codeception/codecept run --debug tests/functional/FilesCommandCest.php:^tryToOutputNonResolvableFTPLinks$
```

Issue #1016, opened 2022-04-14
[S1009] Write SOP for rolling back to a specific good deployment


rija:

rija:
done

PR #1035, opened 2022-05-06
Roll back to previous deployment SOP

pli888:
# Pull request for issue: #1016 [S1009] Write SOP for rolling back to a specific good deployment

* An ROLL_BACK_PREV_DEPLOY.md SOP for rolling back to a previous `live` deployment

### DIFF
--- a/docs/sop/ROLL_BACK_PREV_DEPLOY.md
+++ b/docs/sop/ROLL_BACK_PREV_DEPLOY.md
@@ -19,7 +19,9 @@ stages.
 this roll back procedure. If so, then trigger the `ld_teardown` job in the 
 pipeline. N.B. triggering the `ld_teardown` job will delete all Docker volumes
 including the Let’s Encrypt one, which will mean a new deployment will create a 
-new certificate.
+new certificate. More often than not, `ld_teardown` is not required but there
+might be situations that it is desirable to execute this step.
 4. Click on the final stage of the pipeline and select `ld_deploy` to re-deploy 
 the code that this pipeline is associated with onto the beta-gigadb.org server.
-5. Go to https://beta.gigadb.org in your browser to check the site is running.
+5. Go to https://beta.gigadb.org in your browser to check the live website is 
+running.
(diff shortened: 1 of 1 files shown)


rija (commented):
Hi @pli888,

As it stands currently in this branch and on the  ``develop`` branch,

the action performed in ``ld_teardown`` is already performed as one the first steps in ``ld_deploy``


``.deploy`` job in /Users/owner/code/gigadb-website/ops/pipelines/gigadb-deploy-jobs.yml:

```
    # shutdown currently running container and its volumes
    - docker-compose --tlsverify -H=$REMOTE_DOCKER_HOST -f ops/deployment/docker-compose.production-envs.yml down -v
```

``.teardown`` job in ops/pipelines/gigadb-operations-jobs.yml:

```
    - *configure-docker-compose
    - docker-compose --tlsverify -H=$REMOTE_DOCKER_HOST -f ops/deployment/docker-compose.production-envs.yml down -v
```


So the instructions could just be about running deployment job for the desired pipeline.

However, as part of PR #1038 (certificates renewal fix), a change will be made to the deployment jobs. In order to avoid docker volumes to be deleted (root cause of certs problems)
the service tear down in ``sd_deploy`` and ``ld_deploy``  won’t have the ``-v`` parameter.
But we will keep it on the manual tear down jobs ``sd_teardown`` and ``ld_teardown``, as sometimes we need to wipe out everything including docker volumes. 
The majority of time though,  it is not necessary so it’s fine for deployment job to maintain Docker volumes around (and the Let’s encrypt volume need to stay around)

So the instructions should probably have nota benes mentioning that:

* If docker volume needs to be removed when rolling back, we need to trigger ``ld_teardown`` job
* if triggering ``ld_teardown`` job, all docker volumes will be deleted including Let’s Encrypt one, which mean new deployment will create a new certificate (there could be situations where that’s desirable but often it’s not)

pli888:
@rija Thanks for the comments. I've made your suggested changes now.

Issue #989, opened 2022-03-15
Deploy spreadsheet upload tool with sample spreadsheet and use with RDS


pli888:

only1chunts:
we're using the tools on live so this ticket must be complete

PR #1036, opened 2022-05-08
Detect directories in files url and fix SSL handshake error

rija:
# Pull request for issue: #989

* Show directory url in the report of urls for curators to double check
* Fix SSL handshake when checking urls with PHP's ``get_headers``
* Fix ``setup_dev.sh`` to allow gigadb to build and deploy on Apple Silicon based macs

## Changes to ``protected/commands/FilesCommand.php``:

The first change is to detect if the url is likely to be a directory url by checking if the last component of the path has an extension or not. If not, we add it to the report.

The second change is to fix an intermittent error when running the url checker:

```
get_headers(): SSL: Handshake timed out
```

This is not related to our FTP server on CNGB infrastructure as I met that error with urls to third party web sites too.

I don't know the root cause which is related to how PHP negotiate the SSL connections in ``get_headers``, but since we control the urls we check, we can remove the condition for the error by not peforming SSL verification.
This is done by adding an other option to the stream context parameter of ``get_headers()``

## Changes to ``tests/functional/FilesCommandCest.php``:

When testing checking of HTTP links to files, we use internal URLs to gigadb, initially to avoid the SSL handshake error but we keep them to speed up the tests and avoid potential idiosyncracies of our FTP servers.

When testing FTP links, the ftp link can be dummy as we our just capturing the scheme without actually making  a connection to the url.

Finally, added a test scenario for directory link

## Changes to ``ops/scripts/setup_devdb.sh``

Minor change to the script so it works on Apple Silicon based macs.

Before the change, the script would fail with error "no TTY input device" on those machines.

### DIFF
--- a/ops/scripts/setup_devdb.sh
+++ b/ops/scripts/setup_devdb.sh
@@ -37,9 +37,9 @@ $DOCKER_COMPOSE run --rm  application ./protected/yiic migrate mark 000000_00000
 $DOCKER_COMPOSE run --rm  application ./protected/yiic migrate --connectionID=db --migrationPath=application.migrations.schema --interactive=0
 
 # Drop constraints, indexes and triggers to allow data migrations
-docker-compose run --rm test psql -h database -U gigadb < protected/runtime/dropConstraintsQuery.sql
-docker-compose run --rm test psql -h database -U gigadb < protected/runtime/dropIndexQuery.sql
-docker-compose run --rm test psql -h database -U gigadb < protected/runtime/dropTriggerQuery.sql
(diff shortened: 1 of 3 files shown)


pli888 (approved):
The code looks good to me and I have checked that the tool and its tests are working fine using the commands below:
```
$ docker-compose exec application  ./protected/yiic files checkUrls --doi=100020
$ docker-compose run --rm test ./vendor/codeception/codeception/codecept run --debug tests/functional/FilesCommandCest.php
```

I approve this PR.

PR #1125, opened 2022-09-26
Fix: Upstream failed jobs

rija:
# Pull request for issue: #<insert relevant github issue numbers that this PR is related to>

Fix Pipeline failure on Upstream.
The issues were:

* gigareview test job fails 
* files-urls-updater test job fails
* build and deploy job for the main app will fail

## How to test?

This branch is the pipeline that's already deployed on staging.gigadb.org and beta.gigadb.org.
You can re-execute that pipeline at https://gitlab.com/gigascience/upstream/gigadb-website/.

You can also clone this branch and add a remote on your own Gitlab project and ensure the pipeline still execute there without errors.

The issues were caused by the coumpound effect of the following problems:

* The url of gitlab variables API was hardcoded to ``forks``, so it fails on Upstream pipeline
* Jobs were added to DAG to allow the Gigasciences apps' pipeline to be independent of each other, but some common base job used for caching docker base image was missing in the ``needs`` keyword of the Gitlab config
* Some variables were not set in Upstream's CI/CD settings variables

The ``generate_config.sh`` in those new apps were changed to detect whether it's running as part of Upstream pipeline or not. It does so by checking the value of ``CI_PROJECT_URL`` (which therefore needs to be made available in the ``.env`` file by adding it to the filter used in Gitlab config steps that generate that file).

The build and deploy jobs for the main app need to depends on the ""base_images" job, so that its artifacts are available, this is done  by adding that job to the array value for the ``needs`` keyword in ``ops/pipelines/gigadb-build-jobs.yml`` and ``.gitlab-config.yml``.

Finally, I've updated ``gigareview/README.md`` to mention the variables that were missing on Upstream

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -114,6 +114,7 @@ ld_gigadb:
   stage: live deploy
   tags:
     - $GITLAB_USER_LOGIN
+  needs: ["base_images","build_live"]
   environment:
     name: "live"
     deployment_tier: production
--- a/gigadb/app/tools/files-url-updater/generate_config.sh
+++ b/gigadb/app/tools/files-url-updater/generate_config.sh
@@ -49,7 +49,7 @@ if ! [ -s ./.secrets ];then
 
     # deal with special case we are in Upstream
     if [[ $CI_PROJECT_URL == "https://gitlab.com/gigascience/upstream/gigadb-website" ]];then
-      PROJECT_VARIABLES_URL="https://gitlab.com/api/v4/projects/gigascience%2Fupstream%2F$REPO_NAME/variables"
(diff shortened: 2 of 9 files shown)


pli888 (approved):
From looking at the `Upstream` gigadb-website CI/CD pipelines, I can see the pipeline for the `upstream-failed-jobs` code now successfully runs all the way to `live deploy` stage when it was previously failing. I have also been able to run the `upstream-failed-jobs` code on my CI/CD pipeline to staging deployment and passing acceptance tests.

From looking at the code, I can see how the devops code now takes into account when pipelines are running in `Upstream` thereby allowing deployment on staging and live, and enabling tests for tools and sub-applications to pass in this GitLab sub-project.

I approve this PR.

PR #1166, opened 2022-11-21
Stop legacy DB container service before each run

rija:
# Pull request for issue: #<insert relevant github issue numbers that this PR is related to>

* Always try to stop legacy DB docker server at the start of each run in ``databaseReset.sh``

## How to test?

* Connect to the bastion server
* Ensure there is no`` pg9_3`` named container running
* in ``/home/centos``, run ``./databaseReset.sh``
* Notice that the script started and send a stop command to a non existent container with no impact to the rest of the workflow
* [Optional: if the vsftpd error is still happening, this step won't be necessary] Comment out ``docker stop pg9_3`` under the **Houseekeeping** section at the end of ``databaseReset.sh``
* Run ``./databaseReset.sh``
* Notice a lingering container named `` pg9_3`` still present when doing ``docker ps``
* Run ``./databaseReset.sh``
* Notice it had stopped the server and started a new one, then went on with the workflow

When a run of ``databaseReset.sh`  exits due to error, the final housekeepting that stop the legacy DB service is not performed, causing subsquent runs to fail a start as the still running container causes a name conflict with the container the script want to start. The fix is to always stop the container associated with the legacy DB service before we spin it at the beginning of the script's workflow.

To avoid and error causing the script to exit when trying to stop a container that don't exist (happy path), we add a ``|| true`` command to ensure stopping the container before start a new one always succeed.

### DIFF
--- a/gigadb/app/tools/files-url-updater/databaseReset.sh
+++ b/gigadb/app/tools/files-url-updater/databaseReset.sh
@@ -14,6 +14,9 @@ else
   backupDate=$1
 fi
 
+# Stop any existing legacy database server if it hasn't done so in previous run (e.g: due to errors)
+docker stop pg9_3 || true
+
 # Spin up legacy database (PosgresQL 9.3)
 # Make sure the port 5432 is accessible from host
 docker run --rm --detach --name pg9_3 -p 5432:5432 registry.gitlab.com/$GITLAB_PROJECT_NAME/production_pg9_3:$GIGADB_ENVIRONMENT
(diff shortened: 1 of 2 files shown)


pli888 (approved):
After following the instructions, I confirm the updated `databaseReset.sh` script is able to successfully execute even when a legacy DB container has been left running. I approve this PR.

kencho51 (approved):
Hi @rija,

After deploying this PR to my staging in Seoul region and execute `./databaseReset.sh` in bastion server , the legacy database could be dumped and loaded into the RDS, and the legacy DB container was removed.

PR #1167, opened 2022-11-22
Add disk usage calculator and Gitter notifier using swatch

rija:
# Pull request for issue: 

* Setup a disk usage monitoring and alerting for the bastion server

## How to test?

After building an AWS environment for staging, 
Log in to the bastion with SSH, and run the following command:

```
$ swatch -c check-swatch.conf -p ./diskUsage
```

Then you should see a message in the Gitter notification channel showing disk usage alongside a **(drill)** flag.

There are three moving pieces to the workflow:

1. finding out disk usage
2. notifying that value to the Gitter notification channel if the value is above a threshold
3. setup a cronjob to have that health check happen daily on bastion (no need to do more frequently, as disk usage increase is mostly the result of daily cronjob).

### Details

1. ``df`` command will show information about disk usage, we pipe it through various command to extract just the value for the EBS volume (the one whose path contain ``/dev/``)
2. Install swatch and:
	a. wraps a curl call to the Gitter api into a notify script. It needs Gitter room id, gitter api token and the GigaDb environment variable 
	b. configure ``swatch.conf`` with different thresholds (< 50 %: green, no notifcation; < 70 %: yellow, notify ; > 70 %: red, notify)
	c. call swatch using the ``-p`` argument as we want to pipe the value of disk usage to the notify script
3. Setting up the cronjob, require updating bastion playbook to do: 

	a. Create the diskUsage monitoring script (it does 1.)
	b. Create the Gitter notify script (it does 2.a.)
	c. Copy swatch config file (that's 2.b. as well as a test config for drills)
	d. Install swatch package (use ``dnf`` to install the appropriate RPM)
	e. Setup cronjob to automate daily disk usage monitoring

Finally, other files were changed to enable the environment variables required for 2.a. to be available in the ``.env`` file on the bastion server.

Added two new variables to the **Forks** section of: 

* docs/variables.md


kencho51 (approved):
Hi @rija,

**praise:** It a good idea to have a new variable `$DEPLOYMENT_TIER` and call it in the `notify` script together with `$GIGADB_ENVIRONMENT`, so the alert message is more informative and specific to the deployment environment.

**praise:**  The scripts`diskUsage` and `notify` are created using ansible, they are only existed in the deployed environment, it helps to keep the codebase as small as possible.

**praise:** The cron job functionality is working, as I have rescheduled it 1400 everyday, and I got the alert message without `drill` at that specific time.  

**question(non-blocking):** Why you would use `check-swatch.conf` for testing, and `swatch.conf` for the actual cron job, given that they perform the same function? 

This PR looks good to me, I am happy to approve it.

pli888 (approved):
I confirm that the swatch tool can call the `diskUsage` shell script. With its `conf` file, swatch can then take the output of  `diskUsage` to generate a message that is sent to the Gitter notification chat room using the `notify` script. The code looks good to me too.

Presumably, two swatch configuration files are required, one for manual testing and one for live use because it is not easy to process variables and conditionals in the `exec` line so it knows to send a `drill` message in manual testing.

rija:
Hi @kencho51 

>question(non-blocking): Why you would use check-swatch.conf for testing, and swatch.conf for the actual cron job, given that they perform the same function?

@pli888 gave the correct answer. I would add the motivation was to reduce risks of human errors that can occur when messing with live config, and  yet make it easy to perform integration testing at any time since there are several moving pieces.

Issue #1130, opened 2022-10-05
[S1123] Create master wasabi account and allow developer accounts to access its buckets


rija:
* [x] Create  a new GigaDB account on Wasabi (using tech@ email address)
* [x] Create a bucket for dataset files
* [x] Secure that bucket
* [x] Add permission on that bucket to allow access developer accounts

only1chunts:
all tasks ticked and merged so closing this ticket

PR #1172, opened 2022-11-30
Migration script for transferring datasets from CNGB backup server to Wasabi

pli888:
# Pull request for issue: #1130

* Batch copy of GigaDB dataset files to `live` directory in Wasabi `gigadb-website` bucket
* Default test mode that copies files to `gigadb-website/dev` and `gigadb-website/staging`bucket directories.
* Logging of status messages
* Docker deployment of dataset migration tool on CNGB backup server

## How to test?

See [README.md](https://github.com/pli888/gigadb-website/blob/rclone-cngb2wasabi/gigadb/app/tools/wasabi-migration/README.md) for documentation on how to test and use this tool to migrate GigaDB data.

The main functionality for copying dataset files to Wasabi is provided by the `rclone_copy.sh`. This script is deployed with the `rclone` tool in a Docker container which comes with test data to perform test dataset uploads to the Wasabi `gigadb-datasets` bucket and also has access to live data when installed on the CNGB backup server.

Configuration to use rclone on the CNGB backup server needs to be generated on your local machine.

Information on how to use this Wasabi migration tool is provided by the [README.md](https://github.com/pli888/gigadb-website/blob/rclone-cngb2wasabi/gigadb/app/tools/wasabi-migration/README.md) file.


gigadb/app/tools/wasabi-migration/generate_config.sh
```diff
@@ -0,0 +1,74 @@
...
+fi
+echo "Sourcing secrets"
+source "./.secrets"
+
+# Restore default settings for variables
+set +a
+
+# Generate rclone configuration
+SOURCE=${APP_SOURCE}/rclone.conf.dist
+TARGET=${APP_SOURCE}/config/rclone.conf
```
kencho51, line 72: Hi @pli888,

Dir `${APP_SOURCE}/config/` is not present, when `% docker-compose run --rm config` would get this error:
```
+ SOURCE=/app/rclone.conf.dist
+ TARGET=/app/config/rclone.conf
+ VARS='$WASABI_ACCESS_KEY_ID:$WASABI_SECRET_ACCESS_KEY'
+ envsubst '$WASABI_ACCESS_KEY_ID:$WASABI_SECRET_ACCESS_KEY'
/app/generate_config.sh: line 74: /app/config/rclone.conf: No such file or directory
ERROR: 1
```
If `mkdir config` and then `docker-compose run --rm config`, `rclone.conf` would then be generated.

rija (changes requested):
Hi @pli888,

**praise**: Got functionality working as described in the README file, and the code looks fine to me.

**issue**: In the ``rclone.conf.dist``, the value of the ``acl`` key should be ``public-read``
Otherwise the files can't be downloaded by end users using the public URL

**suggestion**: as already alluded to by @kencho51, it might be a good idea to add ``mkdir config`` in the README file before the line indicating how to generate rclone configuration

**thought**: I see that the max batch size is hard-coded. I wonder whether we might in some situation want to override it using an environment variable

pli888:
>issue: In the rclone.conf.dist, the value of the acl key should be public-read
Otherwise the files can't be downloaded by end users using the public URL

>suggestion: as already alluded to by @kencho51, it might be a good idea to add mkdir config in the README file before the line indicating how to generate rclone configuration

>thought: I see that the max batch size is hard-coded. I wonder whether we might in some situation want to override it using an environment variable

All above suggestions have been implemented.

gigadb/app/tools/wasabi-migration/README.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+#### Batch size for file uploads
+
+The maximum number of datasets that can be uploaded has a default value of 100.
+This can be overridden using the `--max-batch-size`. For example, to increase
+the batch size to 200:
+```
+$ docker-compose run --rm -e HOST_HOSTNAME=`hostname` rclone_cngb /app/rclone_copy.sh --starting-doi 100000 --ending-doi 100300 --max-batch-size 300
+```
```
kencho51, line 298: Hi @pli888,

**non-blocking**: the batch size to 200 => `300`?

gigadb/app/tools/wasabi-migration/README.md
```diff
@@ -0,0 +1,298 @@
...
+Creating wasabi-migration_rclone_run ... done
+    25964 dermatology.data
+    39975 pub/10.5524/100001_101000/100002/CR.kegg.gz
+      649 pub/10.5524/100001_101000/100002/readme.txt
+```
+
+The `rclone_copy.sh` script can batch upload a set of datasets as defined by a 
+starting DOI and an ending DOI. The script will look for the existence of all 
+datasets between these two DOIs. For example, the following command will upload 
+two datasets which are fouund in the `tests/data/gigadb/pub/10.5524/100001_1010001`
```
kencho51, line 58: Hi @pli888,

**typo**: should be `found`.

kencho51 (approved):
Hi @pli888,

**praise:** the logic in the `rclone_copy.sh` considers every environments( dev, staging and live), and allows copying with least human interruption, which means less error.

**praise:** allowing command line options in the `rclone_copy.sh` to optimize the `rclone copy` functionality is elegant.

**thought:** wondering if `--config=` option was allowed the `rclone copy` command, and pointed to customized `rclone.conf` file containing different end points, this tool would then be more versatile as it can copy `from/to` from designated regions.

By following the steps in the readme, the functionalities are working in my local dev environment and in the cngb backup server. And in the cngb backup server, `Migration` user is used, `rclone delete/purge` cmd would result as `Forbidden` or `Access Denied` in the log file when tried to delete files/folders or list non-live directory. And I could see the `--max-batch-size` is working as well.

The code looks good to me, I'm happy to approve it.

rija (changes requested):
Hi  @pli888:  

**praise**: suggested changes have been implemented and I tested them working  

**praise**: good idea for logs naming scheme, make it easy where to find the log for a specific run of the tool 

**issue**: The error message on line 130 is not written to the log when exit code is non-zero.  
 
The exit code shows up on the command line, but if it's not in the log, the notification script won't see it. 
Additionally,  the error message string is missing the value of the exit code even though the intent seem to be there (as indicated by the presence of a colon at the end). 

Console output:
```
~/c/p/g/a/t/wasabi-migration►docker-compose run --rm rclone /app/rclone_copy.sh --starting-doi 100001 --ending-doi 100002                                                   4.144s
Creating wasabi-migration_rclone_run ... done
ERROR: 8
```

Logs output: 
```
2022/12/02 17:25:22 ERROR : CR.kegg.gz: Failed to copy: Put "https://s3.ap-northeast-1.wasabisys.com/gigadb-datasets/dev/pub/10.5524/100001_101000/100002/CR.kegg.gz": max transfer limit reached as set by --max-transfer
2022/12/02 17:25:22 ERROR : Cancelling sync due to fatal error: Put "https://s3.ap-northeast-1.wasabisys.com/gigadb-datasets/dev/pub/10.5524/100001_101000/100002/CR.kegg.gz": max transfer limit reached as set by --max-transfer
2022/12/02 17:25:22 ERROR : readme.txt: Failed to copy: RequestCanceled: request context canceled
caused by: context canceled
2022/12/02 17:25:22 ERROR : Fatal error received - not attempting retries
2022/12/02 17:25:22 Failed to copy with 2 errors: last error was: Put "https://s3.ap-northeast-1.wasabisys.com/gigadb-datasets/dev/pub/10.5524/100001_101000/100002/CR.kegg.gz": max transfer limit reached as set by --max-transfer
```

>note: I've artificially created the error mode by adding ``--max-transfer 1KiB \`` parameter to the rclone copy command, knowing from [the docs](https://rclone.org/docs/#exit-code) that it will cause that specific non-zero exit code.

  The problem's root cause is the `set -e` directive at the top that cause the script to exit after rclone has exited meaning any instructions after the rclone command won't be parsed.  
  A possible fix is to add a `set +e` just  before the call to `rclone copy`. That approach allows the top `set -e` to still protect the code block before the call (the operator don't want rclone to be called at all if there are any problems in the code above, and since at that stage the operator is still interacting, there is no need to log those)

pli888:
>A possible fix is to add a set +e just before the call to rclone copy

@rija This fix has been added. I confirm the exit code is now written into the log file in case there is a problem with running rclone.

rija (approved):
Hi @pli888 

**praise**: it looks good to me now. Happy to approve

Issue #1149, opened 2022-11-08
[S1123] Write acceptance tests for secure and reliable connectivity to Wasabi


rija:
* [x] Identify and setup all Wasabi policies that need tested
* [x] Write  wasabi.feature:

Examples:

Given I have a file "dummy.txt"
And I use developer configuration
When I transfer it to "test/gigadb-datasets"
Then It should succeed

Given I have a file "dummy.txt"
And I use developer configuration
When I transfer it to "live/gigadb-datasets"
Then It should not succeed

Given I have a file "dummy.txt"
And I use migration configuration
When I transfer it to "live/gigadb-datasets"
Then It should succeed

Given a file "dummy.txt" exists on Wasabi in "live/gigadb-datasets"
And I use migration configuration
When I delete it from "live/gigadb-datasets"
Then It should not succeed

* [x] Implement the steps

PR #1177, opened 2022-12-11
Wasabi policies and acceptance tests

rija:
# Pull request for issue: #1149

* persistence of all policies in use in single JSON files
* Automated acceptance tests that test and document those policies

## How to test?

The acceptance tests defined in ``tests/Wasabi.feature`` will exercise the applied policies and must no fail:

```
$ docker-compose run --rm codecept run --no-redirect --debug --fail-fast -g wasabi acceptance
```

For each capability:
 
* we create a single JSON file policy
* ensure it's not hardcoding any user or group, so the policy can be reused
* mention it's in filename the Wasabi environment it applies to
* create the policy in Wasabi
* associate the policy to the appropriate user or group

Then we write the test scenario corresponding to the application of the policy.

A prerequisite for acceptance tests to run is configuring rclone with the appropriate credentials.
This is done by storing credentials for each policy beneficiary that need tested in Gitlab Variables.
To avoid each developers running the tests to have to set them, they are already set up on the ``Forks`` and ``Upstream`` projects. **There's nothing for you to do**. Below is the list of variables

| Variable name | comment |
| -- | -- |
| MIGRATION_ALT_WASABI_ACCESS_KEY_ID  | dangerous to use real access/secret key, so create an ALTernative set for running the tests |
| MIGRATION_ALT_WASABI_SECRET_ACCESS_KEY | dangerous to use real access/secret key, so create an ALTernative set for running the tests | 
| CODECEPTDEV_WASABI_ACCESS_KEY_ID  | the default set of keys for a user of Developer group dedicated to testing |
| CODECEPTDEV_WASABI_SECRET_ACCESS_KEY | the default set of keys for a user of Developer group dedicated to testing |  
| CODECEPTCUR_WASABI_ACCESS_KEY_ID  | the default set of keys for a user of Curator group dedicated to testing |
|CODECEPTCUR_WASABI_SECRET_ACCESS_KEY | the default set of keys for a user of Curator group dedicated to testing |  

Rclone is then called with appropriate commands and test data in each step implementation to test the effect of the policy.

The scenario for the admin role, required programmatically requesting assuming that role in order to obtain temporary credentials (access key, secret key and session token) that embody the elevated permission.
This is done by [using the StsClient class from the AWS SDK for PHP](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/php/example_code/sts/AssumeRole.php).


pli888 (changes requested):
I think the code looks good and I managed to get all the tests passing. Before executing the command to run the tests though, you need to `./up.sh` the docker services but this is where I had a couple of problems:

I encountered this error:
```
Error response from daemon: pull access denied for deployment_application, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
This was fixed by editing `docker-compose.yml` on line 159 to:
```
image: ${COMPOSE_PROJECT_NAME}-application:latest
```

I also had this error when running `./up.sh`:
```
yiisoft/yii2-composer contains a Composer plugin which is blocked by your allow-plugins config.
```
I ran the command below to fix this error:
```
$ docker-compose run --rm console composer config --no-plugins allow-plugins.yiisoft/yii2-composer true
```

All of the Docker services were then able to start and all the Wasabi acceptance tests passed after executing them using:
```
$ docker-compose run --rm codecept run --no-redirect --debug --fail-fast -g wasabi acceptance
```

rija:
Hi @pli888, @kencho51  

Thanks, I've updated the ``composer.json`` for FUW so it won't show the error message for ``yii2-composer``. All our other composer config files already had the ``allow-plugins`` section with ``yii2-composer`` in it.

kencho51 (approved):
Hi @rija,

**praise:** It is brilliant that implementing automatic tests on a backup process regarding different user group and different environment, which makes our tool reliable as it has been tested thoroughly.

**praise:**  The implementation in the `DeveloperSteps.php` is inspiring,  the way of how we are going to use a tool actually could be tested.

**question(non-blocking):** In policies `AllowReadWriteContentOn*.json`, each action has an empty object `"Condition": {}`, what's the  purpose of this? Or it can be removed as it is empty?

**polish(non-blocking):** I executed `phpcs --standard=PSR12 tests/_support` and it returned:
```
FILE: /gigadb-website/tests/_support/DeveloperSteps.php
--------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AND 15 WARNINGS AFFECTING 16 LINES
--------------------------------------------------------------------------------------------------------
  13 | ERROR   | Each class must be in a namespace of at least one level (a top-level vendor name)
``` 
And line 13 is `class DeveloperSteps extends \Codeception\Actor `.

After getting the latest commits, the `Wasabi.feature` is passing on both my dev and gitlab pipeline. The code looks good to me too, so I am happy to approve this PR.

pli888 (approved):
I confirm there is no plugin error message now when I execute `./up.sh`. All Wasabi acceptance tests are still passing for me. I also confirm that when logged into the Wasabi web console using my developer account, I confirm I cannot upload or delete files in the `live` directory.

Issue #1056, opened 2022-06-13
bug found when viewing dataset after adding a new dataset-manuscript link via admin pages


only1chunts:
## User story

>As a curator
>I want to add a manuscript DOI link via the admin page https://beta.gigadb.org/adminManuscript/admin
>So that when a manuscript is given a DOI we can include that in the GigaDB entry

I suspect the issue is that the manuscript DOI I have entered is not YET published by OUP, they have only pre-assigned it, this means that the DOI is not valid and will not appear in cross-ref yet, effectively its a broken link, but we know that it will (eventually) be published and provide the correct link to the MS within the next X days/weeks depending on how fast/slow OUP production is or if there is any embargo on the manuscript.
The example that brought about this ticket is: https://beta.gigadb.org/dataset/view/id/102215
But if my suspicion is correct then that page will not be broken as soon as the DOI is minted by OUP, so depending on when this ticket gets looked it, it might not be showing an error message. Screen attached below shows what I see at the moment:

![image](https://user-images.githubusercontent.com/6037145/173389680-9310bbea-6ce7-4b85-b282-c83800945463.png)

## Acceptance criteria

>Given I add a manuscript link that is currently not "live"
>When I visit the relevant public dataset page
>Then the web page is still displayed but not displaying the link to the manuscript (until that MS DOI is live)

## Additional Info

Seems like the DOI is being used to retrieve citation information. If DOI link is broken then the dataset page is not displayed. Need to check the code that fetches information catches this particular problem.

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
[...]

PR #1187, opened 2023-01-04
Retrieve datacite text using try catch

kencho51:
# Pull request for issue: #1056 and #1203

This PR tries to fix:
1. #1056

After logging in as admin, and update the `identifier` in `/adminManuscript` page:
  - if the identifier is not found in https://doi.org, the request exception `404 Not Found` would be handled, and the dataset page (/dataset/view/id/$doi) would work.
 
2. #1203

In dataset page, and try to get the citation text in the `Cite Dataset` box:
  - if the identifier is not found in https://data.datacite.org/, the request exception `404 Not Found` would be handled and trigger the action of hiding the `Cite Dataset` box.

##### In dev:
For #1056:
a. Manual test
```
% cd /gigadb-website
% ./up.sh
# update the identifier to `10.1186/s13742-015-9999-9` in http://gigadb.gigasciencejournal.com:9170/adminManuscript/update/id/281
# go to http://gigadb.gigasciencejournal.com:9170/dataset/view/id/100142, the page would work
```

b. Automatic test
```
# Functional test
 % docker-compose run --rm test ./vendor/codeception/codeception/codecept run functional StoreDatasetConnectionsCest -v                                            
Creating deployment_test_run ... done
Running with seed: 


Functional Tests (1) ---------------------------------------------------------------------------------------------------------------------------------
Modules: Db, \Helper\Functional, \Helper\MailHelper, Asserts, PhpBrowser
------------------------------------------------------------------------------------------------------------------------------------------------------
✔ StoreDatasetConnectionsCest: Try view the dataset page with not found identifier (4.99s)
------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 30.07 seconds, Memory: 16.00MB

OK (1 test, 1 assertion)
```

2. For #1203 
a. Manual test
```
% cd /gigadb-website
% ./up.sh

 # then go to http://gigadb.gigasciencejournal.com:9170/dataset/300070, and the `Cite Dataset` button could not be found.
 ```
b. Automatic tests:
```
# Acceptance test
 % docker-compose run --rm codecept run --no-redirect -g cite-dataset-button acceptance
Creating deployment_codecept_run ... done
Codeception PHP Testing Framework v5.0.4 https://helpukrainewin.org
[Groups] issue-1056 

Acceptance Tests (3) -----------------------------------------------------------------------------------------------------------------------------------
✔ a user visit the dataset page: Display the cite dataset dropdown box content (21.17s)
✔ a user visit the dataset page: Display the citation text from datacite (20.46s)
✔ a user visit the dataset page: Display no Cite Dataset button when bad request (4.37s)
--------------------------------------------------------------------------------------------------------------------------------------------------------
Time: 01:13.112, Memory: 12.00 MB

OK (3 tests, 6 assertions)
```

##### In staging:
1. Manual test for #1056:
  - go to staging gigadb website and log in as admin.
  - update the identifier to `10.1186/s13742-015-9999-9` or any other random one in `/adminManuscript/view/id/38`
  - go to `/dataset/view/id/100142`

For #1056 
For updating the `identifier` in `/adminManuscript/update/id/$id`, an additional  catch block `ClientException` is implemented to capture the `404 Not Found` exception.

For #1203 
For `Cite Dataset` box in the dataset page,  the `Text` button will make a request to `https://data.datacite.org/` and return the text citation in a new page. By implementing  `DownloadService` component class,  the exception would be caught, and then `Cite Dataset`  button would be hidden.

The exception message will be logged using `yii::log` method.

Difficult to identify the dataset in staging that the DOI has not been minted by `https://data.datacite.org/` for manual testing.

None.

None.

##### Bring in legacy unit tests
1. Bring in 39/48 of legacy unit test files in `protected/unit/` to `tests/unit/`. 
2. Update the class name from `AffilateUserIdentityTest` to `AffiliateUserIdentityTest` for the `tests/unit/AffiliateUserIdentityTest.php`
5. Update the file name from `DatasetFildedropTest.php` to `DatasetFiledropTest.php` for the `protected/tests/unit/DatasetFiledropTest.php`

```
# Run the updated unit tests
% docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit
....
-----------------------------------------------------------------------------------------------------------------------


Time: 53.83 seconds, Memory: 28.00MB

OK (219 tests, 637 assertions)

# Run the legacy unit tests
% ./tests/unit_runner
....
Time: 18.67 seconds, Memory: 24.00MB

OK (81 tests, 251 assertions)
```

None.

### CODE REVIEW

File: protected/views/dataset/view.php [hunk 1]
```diff
@@ -51,15 +51,29 @@
                                             <div class="dropdown-box">
                                                 <button id="CiteDataset" class="drop-citation-btn dropdown-toggle" type="button" data-toggle="dropdown">Cite Dataset<span class="caret"></span></button>
                                                 <?php
-                                                $text = file_get_contents('https://data.datacite.org/text/x-bibliography/10.5524/' . $model->identifier);
-                                                $clean_text = strip_tags(preg_replace("/&#?[a-z0-9]+;/i", "", $text));
+                                                $url = 'https://data.datacite.org/text/x-bibliography/10.5524/' . $model->identifier;
+                                                $webClient = new GuzzleHttp\Client();
```
pli888: **thought:** If we want to fetch information from an external source such as DataCite - this functionality is related to what the [DownloadService](https://github.com/gigascience/gigadb-website/blob/develop/protected/components/DownloadService.php) component class provides?

pli888 (commented):
There are two parts to this PR:
  * Conditional display of the `Cite Dataset` button based on whether citation for the dataset can be fetched from a [datacite.org](http://datacite.org/) URL containing the dataset identifier. If URL is not resolvable, e.g. 404 error message then error is caught and button is not displayed.
  * Display citation for the manuscript publication reporting the dataset in the Additional details section. The manuscript citation comes from https://doi.org/. If the [doi.org](http://doi.org/) URL does not work with the manuscript identifier then the error is caught and dataset page is displayed instead of hanging.

**praise:** The functionality works. To test the PR, I updated dataset 100142's associated manuscript with a fake DOI: 10.1186/s13742-015-9999-9. When I go to http://gigadb.gigasciencejournal.com:9170/dataset/view/id/100142, the new code catches the error when trying to fetch manuscript information and dataset page fully loads.

**praise:** The Cite Dataset button is now not displayed for http://gigadb.gigasciencejournal.com:9170/dataset/300070, because it does not an associated manuscript.

**praise:** All acceptance tests are passing. The functional test is passing.

**issue:** `Read the peer-reviewed publication(s):` is displayed on the dataset page even when there is no manuscript citation in the Additional details section. For example, see http://gigadb.gigasciencejournal.com:9170/dataset/view/id/100142. Is it possible to not have `Read the peer-reviewed publication(s):` displayed when there is no manuscript information to be shown?

only1chunts:
Hi,
Maybe i'm mis-understanding something here but I would like to claify....
1 - 

> praise: The Cite Dataset button is now not displayed for http://gigadb.gigasciencejournal.com:9170/dataset/300070, because it does not an associated manuscript.

This is not the correct behaviour? The Cite Dataset button has nothing to do with the linked manuscript and should always be displayed.

2 - 

> issue: Read the peer-reviewed publication(s): is displayed on the dataset page even when there is no manuscript citation in the Additional details section. For example, see http://gigadb.gigasciencejournal.com:9170/dataset/view/id/100142. Is it possible to not have Read the peer-reviewed publication(s): displayed when there is no manuscript information to be shown?

100142 does have a linked manuscript in live, so unless its been deleted from this test server then there should be something displayed.

kencho51:
Hi @only1chunts,

> This is not the correct behaviour? The Cite Dataset button has nothing to do with the linked manuscript and should always be displayed.

The `Cite Dataset` button on dataset page will try to retrieve citation information from `https://data.datacite.org/$identifier`, if the link is broken, `404 Not Found` would be resulted when `Text` is clicked.
Take this url (https://data.crosscite.org/text/x-bibliography/10.5524/109999)  as an example, you would get  get this error `The resource you are looking for doesn't exist.` 
By hiding the `Text` option can prevent someone seeing that error page. For the same reason, the options for `RIS` (https://data.crosscite.org/application/x-research-info-systems/10.5524/109999) and `BibTex` ( https://data.crosscite.org/application/x-bibtex/10.5524/109999) could also be hidden. So, as a result, hiding the `Cite Dataset` button seems to be a fix for all.

only1chunts:
There is something very off with this ticket. I think we need to discuss the actions as it doesn't all make sense. The two parts to this ticket are completely separate, yet they are being merged together here which makes it impossible to tell which fixes/tests are for which of the issues.
The Cite Dataset Button has nothing to do with manuscript links, so why does the manual test say "change the manuscript link in the admin page and look at the Cite dataset button on the staging server"? this is not logical as they are unrelated so one should not affect the other. If it does then we have a problem.
Maybe the implemented functions are correct, i.e. If any call to datacite fails then dont show that part of the page, but the testing of it is definately questionable.

pli888:
@only1chunts I think the PR could have been better structured so it is more clearer. The way I understand it is that this PR is to do with fixing the display of the manuscript citation in a dataset page in the Additional information section. The work to do with the `Cite Dataset` button and its manual testing is technical debt repayment and would have been better placed under this heading in the PR.

tests/acceptance/DatasetView.feature
```diff
@@ -143,3 +143,29 @@ Feature: a user visit the dataset page
     When I am on "/dataset/100935"
     Then I should see "Github links"
     And I should see "https://github.com/cihga39871/Atria"
+
+  @ok @issue-1056
```
pli888: **:suggestion** In a few months time, I will have forgotten what issue 1056 was about. I think a better tag to use would be something like `@cite-dataset-button`

rija (commented):
**praise:** successfully replicated the instructions in "How to test" sections
**praise:** the new functional tests and acceptance tests scenarios pass and look fine to me
**issue:**  broken publication links should not be present in the returned array
the `getPublications()` function return a list of links to be ultimately shown by the view for display, if there is no link nothing is displayed,
If a link has a problem (and therefore caught by a try...catch block), it should not be part of the returned array.
As it stands now, I think there still is a link row added to the array even if the publication link has errored. That's the probable cause of the issue @pli888 referred to
**polish:** check PSR12 on `protected/components/StoredDatasetConnections.php`, `protected/views/dataset/view.php`, and `tests/_support/AcceptanceTester.php`

rija (commented):
**praise:** good job moving most of the legacy tests over to Codeception

kencho51:
Hi @pli888 and @rija,

By adding condition `&& null !== $publications['citation']` [5db2a50](https://github.com/gigascience/gigadb-website/pull/1187/commits/5db2a50903b38b12c9882ccc7d86826d6a8d7191) in the `/protected/view/dataset/view.php`, the `Read the peer-reviewed publication(s):` on the dataset page now will be hidden if the link is broken.

But the array returned by the `getPublications()` still contains the ['citation'] entry with empty value.

rija:
@kencho51 remember the try/catch is inside a loop, if you end up in a catch block you should skip that iteration, no reason to perform the rest  of the loop block in that case. It's not correct to return an array with dodgy data in it

kencho51:
Hi @rija,


> @kencho51 remember the try/catch is inside a loop, if you end up in a catch block you should skip that iteration, no reason to perform the rest of the loop block in that case. It's not correct to return an array with dodgy data in it

In this [eb9e76f](https://github.com/gigascience/gigadb-website/pull/1187/commits/eb9e76fb9ff0f71177550d6c546b9868fdd7239c), the loop is inside the try/catch block, so if there is an exception, the loop would be skipped and the array returned would be empty.

The functional test and the unit test have been updated aw well.

rija:
Hi @kencho51,

I think putting the try/catch block outside a loop may cause problems:
If we have a list of links, and the first link is problematic, an Exception will trigger the catch block which is outside the loop.
That means, that none of the remaining links will make it to the global returned array, which is not good.

When I said "you should skip that iteration", I meant it literally, using the [continue](https://www.php.net/manual/en/control-structures.continue.php) directive.

rija (changes requested):
Hi @kencho51,

**praise:** Can see only working links are returned to the views from domain layer for publication links
**issue:** there is block of code in `protected/views/dataset/view.php` that still has the potential to blow up to the user screen:

```
 57.   $textFileExist = DownloadService::fileExists($url);
 58.  if ($textFileExist) {
 59.  	$textFile = DownloadService::downloadFile($url);
```
If a network failure happen between the moment line 57 is executed and the moment line 59 is executed, an uncaught exception will be displayed to the user 
Because `DownloadService::downloadFile` throws a generic PHP exception (which is also not optimal) and the view code is not catching it.
**suggestion:** One way out of it, is you could forgo the use of `DownloadService::fileExists` and just wrap line 59 in try/catch
With the `$showButton = false;` in the catch block, and the `$showButton = true;` in the try block

`DownloadService::downloadFile()` throws a generic PHP Exception, which could be then one you need to catch, but it's not the correct exception for it to throw.
Ideally, in that function (I know it's not code you wrote, but I missed it last time), we should check using a `switch case` block, the different response codes (not just 200 vs everything else) and then throw the appropriate exception as listed here: https://docs.guzzlephp.org/en/stable/quickstart.html#exceptions
But an easier and still correct for now implementation would be to throw `GuzzleHttp\Exception\BadResponseException` instead of the generic PHP exception and catch it in the code in `protected/views/dataset/view.php`
You DON'T need to modify `protected/commands/FilesCommand.php`'s `actionUpdateMD5FileAttributes()`, the original function that needed `DownloadService::downloadFile()`, because that function is somehow catching `\GuzzleHttp\Exception\GuzzleException` which is the interface also implemented by `GuzzleHttp\Exception\BadResponseException`'s parent class, so the suggested change will still work with that code (because of the L in SOLID principles)

kencho51:
Hi @rija and @pli888 ,

The `downloadFile()` has been updated to throw `GuzzleHttp\Exception\BadResponseException`, and it will be caught in the `protected/views/dataset/view.php`.

Although the `downloadFile()` function may throw other `GuzzleException`, the caller method needs to have  a catch block to handle it, such as `actionUpdateMD5FileAttributes()` already has one.

So, please have a look of this PR again.

rija (approved):
Hi @kencho51

praise: the code and usage of `throw` and `try...catch` in the `DownloadService` and the view look good to me
praise: the `DownloadService`'s `downloadFile()` method now throws a relevant exception

LGTM, happy to approve

pli888 (approved):
**praise:** I can see the downloadFile function in DownloadService class now returns a more appropriate BadResponseException if the response status code is not 200. The BadResponseException will be caught in the dataset view class if it fails to retrieve citation data for a manuscript from DataCite and disable display of the cite dataset button due to the `showButton` Boolean variable being false.

@kencho51 I approve this PR.

Issue #1135, opened 2022-10-05
[S1123] Create SOP to operate the transfer to Wasabi


rija:
* [x] Document configuration and access prerequisites
* [x] Document the process for transfering dataset files to wasabi.
* [x] Explain what happen when there are errors
* [x] Include Backup server to Zabbix dashboard
* [ ] Add link to Backup server Zabbix dashboard
* [ ] Add link to Wasabi Status page
* [ ] Add link to Wasabi console
* [ ] Add link to download urls

rija:
wasabi status page: https://status.wasabi.com/

PR #1195, opened 2023-01-21
SOP for migrating live data to Wasabi

pli888:
# Pull request for issue: #1135

* Instructions for performing dataset migration from CNGB backup server to Wasabi bucket

## How to test?

Perform the steps in [WASABI_DATA_MIGRATION.md](https://github.com/pli888/gigadb-website/blob/wasabi-sop/docs/sop/WASABI_DATA_MIGRATION.md)

The `migrate.sh` script has been updated with functionality to parse a fourth command-line argument for parameterising a boolean value that determines whether to migrate `live` data or not. Without this, it is not possible to migrate production data on the CNGB backup server.

The `send_notification.sh` shell script has been updated so that the `if` conditional block will send a partially-hardcoded notification message that tells you if an error originated from a local dev Swatchdog Docker container or from the CNGB backup server.

The README.md has been updated as well as the new SOP document.


docs/sop/WASABI_DATA_MIGRATION.md
```diff
@@ -0,0 +1,353 @@
...
+```
+[centos@ip-xxx-xx-x-x]$ sudo docker-compose run --rm rclone /app/rclone_copy.sh --use-live-data --starting-doi 100002 --ending-doi 100020
+Cannot copy live data because we are not on backup server - exiting...
+```
+
+#### Batch size for file uploads
+
+The maximum number of datasets that can be uploaded has a default value of 100.
+This can be overridden using the `--max-batch-size`. For example, to increase
+the batch size to 200:
```
kencho51: Change `200` to `300` to match with the cmd below?

kencho51 (approved):
Hi @pli888,

**praise:** Change the name from `start_backup_proccess.sh` to `migrate.sh` is more concise.

**praise:** I could follow the steps in `WASABI_DATA_MIGRATION.md` to replicate the migration process in the live backup server.

**praise:** I further confirmed that user `Migration` cannot delete files in both `gigadb-datasets/staging/pub/10.5524/` and `gigadb-datasets/live/pub/10.5524/100001_101000/`.

**praise:** I further confirmed that developer can delete the live dataset files only after `Switch Role` to `Admin`.

**praise:** It is helpful to mention the source of error (eg. server name) in the `send_notification.sh`.

**polish(non-blocking):** In the `WASABI_DATA_MIGRATION.md` line 251, should change `200` to `300` to match with the cmd below.

Changes look good to me, I'm happy to approve it.

rija (approved):
praise: the SOP make sense to me
praise: the code change looks good
nitpick (non-blocking): same ones already noted by @kencho51

happy to approve

docs/sop/WASABI_DATA_MIGRATION.md
pli888: @kencho51 This has been corrected now.

Issue #1178, opened 2022-12-12
Create Console Controller class


pli888:
- [x] Use model classes to retrieve readme information
- [x] Place new created readme.txt file in /home/curators directory
- [x] Create unit tests
- [x] Create functional test to ensure script outputs a file

Try using dataset 100314 which is a bugbear dataset.

Also try using dataset 100310 which has >19k files.

Points to note:

* Use model classes in `/gigadb/app/worker/file-worker/models` not the model classes that come with GigaDB repo.

rija:
May need to add a PSR-4 clause to composer json to access model class:
```
 "autoload": {
        "psr-4": {
            "app\\models\\" : "/gigadb-apps/worker/file-worker/models"
        }
     }
```

only1chunts:
everything is ticked and merged so I think we can close this ticket

Issue #1073, opened 2022-06-29
Dataset 100314 not working


tuli:
**Describe the bug**
GigaDB dataset 100314 is bad on live, but non-existent on beta, see ticket #676 for live bug already reported.

**To Reproduce**
1. Go to http://beta.gigadb.org/dataset/100314

**Expected behavior**
Should look more like other datasets e.g. http://gigadb.org/dataset/102236.

**Desktop (please complete the following information):**
- iOS 10.12.6 n.b. I know I need to update this.
- Chrome 103.0.5060.53
- Safari 12.1.2

**Screenshots**
![image](https://user-images.githubusercontent.com/6037145/176832742-80a8515f-13c0-472b-9ecb-6518d17bbc80.png)

Even on Live website it doesn't work properly, see also ticket #676 
![Screen Shot 2022-06-29 at 13 01 23](https://user-images.githubusercontent.com/1359170/176431562-f1e4c75b-e9e9-47ba-8c31-a00c2d5a94ea.png)

![Screen Shot 2022-06-29 at 13 01 38](https://user-images.githubusercontent.com/1359170/176431549-abdf6595-0a50-4092-8b9c-5e08777b33c9.png)

rija:
The log for the webserver is showing this
```
2022/12/20 07:38:33 [error] 29#29: *13712 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/vendor/yiisoft/yii/framework/base/CBehavior.php on line 111PHP message: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 65536 bytes) in /var/www/vendor/composer/ClassLoader.php on line 571" while reading response header from upstream, client: 81.220.58.148, server: beta.gigadb.org, request: "GET /dataset/100314 HTTP/2.0", upstream: "fastcgi://172.16.238.2:9000", host: "beta.gigadb.org"
```

pli888:
Sometimes dataset contains lots of files or lots of samples. Suggestion is to show the first 1000 items. Need to provide a way to show users how to get hold of the rest of the files/samples, e.g. click here to get the remaining items.

Issue #676, opened 2021-06-15
Bug? display of dataset 100314


only1chunts:
**Describe the bug**
dataset 100314 is not displaying as expected, it appears to have lost all GigaDB style and is showing as a plain html page (see attached screenshot.

**To Reproduce**
1. Go to http://gigadb.org/dataset/100314
2. See error

**Expected behavior**
The dataset page should disaply in the same way all other GigaDB datasets do, using the GigaDB CSS style sheets

**Screenshots**
![Screenshot_100314](https://user-images.githubusercontent.com/6037145/122064803-54e06c00-cde9-11eb-8ed0-bba981c5c9d4.png)

only1chunts:
same issue appears in dataset [100310](http://gigadb.org/dataset/100310)

only1chunts:
It has been noticed that the above issue may be related to the very large number of either; samples (in the case of 100314) or, files (in the case of 100310).

only1chunts:
this is a duplicate of #406

PR #1202, opened 2023-02-09
Allowed memory size exhausted

rija:
# Pull request for issues: #1073 and #676

1. The main change provided by this PR is to allow visitors to successfully load the page at `/dataset/100314` which has 65K+ samples attached which is the issue raised in the issue #1073

To achieve that a number of the following technical problem also had to be fixed before: 

2.  Fix a test data error when using `production_like` test data
3.  Fix incorrect interface used for cache classes
4.  Streamlined how memory usage for PHP is configured
5.  Introduced `t3.small` as the default type of EC2 instance to use for deployment

Additionally,
 
6. because it is the same code solution,  the same fix to the above bug was adapted to fix similar problem accessing `/dataset/100310` which has near 200K files attached and is mentioned in issue #676

Finally,

7. Upgraded the main application  to PHP 7.4 for better performance, security, functionally and as a step toward migrating to PHP 8 later

## How to test?

* Deploy this branch on your staging environment
* navigate to the urls  `/dataset/100314` and `/dataset/100310`
* they should load successfully

1. The main change is straightforward and has four parts

First part: It consists in adding a `LIMIT` clause to the sql query used for retrieving the underlying samples or files data from the databases. This is done in the `protected/components/StoredDatasetSamples.php` and the `protected/components/StoredDatasetFiles.php`

Second part: Increase the memory available to PHP by ensuring the `memory_limit` in [PHP configuration](https://www.php.net/manual/en/ini.core.php#ini.memory-limit) is quadrupled on AWS deployments

Third part: Double the memory available to the EC2 instance from 512 GB to 1 GB RAM. Moving from `t3.micro` to `t3.small` would do the job, but at double the cost, so in theory it's better to use `t3a.small` which is just slightly more expensive than `t3.micro`, but give us the 1 GB RAM. The Only different otherwise is we get an AMD processor instead of Intel.
However, `t3a.small` is not available in Hong Kong, and its comparative price advantage is less pronounced in Tokyo and Seoul regions, so we will go with `t3.small` for now.

The threshold was set to 3000 for samples and 8000 for file, see next section for implementation issues for more about those thresholds.

Fourth part: To alleviate the slow loading of those datasets, I have added a new Gitlab job for `stage deploy` stage (`sd_warmcache`) and for `live deploy` stage (`ld_warmcache`) to load the pages in the cache preemptively just after deployment. This required increasing the value of  `fastcgi_read_timeout` in Nginx config file to avoid intermittent "504 Gateway Timeout" when those jobs run.

2. See the "Any changes to automated tests?" Section

3. Cache object in the codebase: 

The caching layer used in components created to serve DatasetController was meant to be flexible, so we could change the cache implementation without code changes. This is done by using the [`CCache`](https://www.yiiframework.com/doc/api/1.1/CCache) interface wherever we need a Cache object in the components code. The concrete cache is being injected from configuration file. In particular I wanted to debug the cache layer in order to gauge if it had any role in the problem were were trying to fix here. One way if to replace our [Apcu cache](https://www.yiiframework.com/doc/api/1.1/CApcCache) with  a [dummy cache](https://www.yiiframework.com/doc/api/1.1/CDummyCache). It didn't work because the latter implement `ICache` interface which is a superclass to `CCache`. So for maximum flexibility (and to get on with my debugging) I have replaced `CCache` with [`ICache`](https://www.yiiframework.com/doc/api/1.1/ICache) in the method signature for methods that need  a cache object.

4. See the "Any improvements to CI/CD pipeline?" Section
5. See the "Third part" paragraph above
6. Because I remembered there was several occurrences of dataset pages not loading because there are too many files associated with them, I dug out the DOI from various bug tickets:  first #1178 which give the DOI, which is also mentioned in #676 alongside the DOI of ticket #1073. The fix is the same  and is described above in the 1. paragraph
7. See the "Any technical debt repayment?" Section

The current solution (which is a stop-gap one), involved limiting the numbers of rows to display in the tables on the dataset page.
The cutoff threshold was chosen to be the highest possible before the page can no longer load properly.

The two tables below (one for samples, one for files) show the small number of datasets that are impacted with this truncation.

| "identifier" | "nb_sample" |
| -- | -- |
| "100314"	| 65535 |
| "100193"	| 11670 |
| "100824"	| 6467 |
| "100855"	| 4959 |
| "100332"	| 4033 |

| "identifier" | "nb_file" |
| -- | -- |
| "100396"	| 197214 |
| "100310"	| 19981 |
| "100707"	| 16675 |
| "100608"	 | 12864 |
| "100512"	| 12238 |

Temporary palliative measure will be taken by the curators team to address this situation with the non-complete sample table.
For the file table, the visitors can click on the "Ftp site" link to access all the files.

Additionally, this fix only make those problematic DOIs to load in the dataset view, they are still going to be loading very slowly in the browsers for now.

The test data for `production_like` was causing errors:

```
% ./up.sh production_like
...
*** applying m200529_050060_insert_data_dataset_tab
ParseError: syntax error, unexpected 'methylation' (T_STRING), expecting ')' in /var/www/protected/migrations/data/production_like/m200529_050060_insert_data_dataset_tab.php:15
Stack trace:
#0 /var/www/vendor/yiisoft/yii/framework/cli/commands/MigrateCommand.php(414): MigrateCommand->instantiateMigration('m200529_050060_...')
#1 /var/www/vendor/yiisoft/yii/framework/cli/commands/MigrateCommand.php(109): MigrateCommand->migrateUp('m200529_050060_...')
#2 [internal function]: MigrateCommand->actionUp(Array)
```

@kencho51 confirmed that it's not just me.

I needed to use that test data because I needed production like datas and identifiers, so I can easily import canned data from staging tables in order to reproduce the problem on my local environment.

The root of that problem was the fixture file ``data/production_like/dataset.csv`` which lots of quotes issues that become  syntax errors when transformed into PHP code for reasons I don't know. I've corrected the file, so it generate valid PHP migrations.

Documented the changes to AWS policies to allow use of the `t3a.small` and `t3.small` EC2 instances.

1. Added a missing repo configuration.

*  `.github` directory exists on `master`, but is missing on `develop` and it's needed there so we could retire `master` later in favour of a `main`

2. Fixed failed Gigareview test job in CI pipeline (missing dependency in Gigareview's composer.lock)
3. Make code in `protected/components/` and in `DatasetController.php` to conform to PSR12 coding guideline

### Migration to PHP 7.4:

Improved performances, security, and new useful language features. Also a stepping stone towards migrating to PHP 8+
I had to solve the following issues to make it work:

* Problems with "leafo/lessphp":
	 * https://github.com/leafo/lessphp/issues/208, which was fixed by implemented [this](https://gist.github.com/barryvdh/2029466)
 	* "PHP Error[8]: Trying to access array offset on value of type bool" which is explained  [here](https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.non-array-access) and fixed by giving the less docker container service its own compatible php image
 	* `sh: /var/www/protected/yiic: not found` error on stage deploy was fixed by not calling less in deploy job as the CSS files are already built and baked in Production docker file during build job
 * `GD` installation directive in Dockerfile are different for PHP 7.4+
 * `mcrypt` is no longer needed in PHP 7.4+
 
Streamlined the configuration of the PHP memory limit across all environments.
As described above, changing the value of PHP's `memory_limit`configuration was key to getting the fix working.
The problem is that there were several places in our codebase where that configuration key could be changed, 
Creating confusion and time wasting when changing it in the wrong place, and the right place differed depending on the environment.
Now there is only one way place to change it, which is the php.ini file bind-mounted in the container services from either `ops/configuration/php-conf/php-dev.ini` or `ops/configuration/php-conf/php-production.ini` depending on whether you are on local environment or not.
I have removed `ops/configuration/php-conf/gigadb.ini` as it wasn't unnecessary, since the content could go in the ini files aforementioned.
I have also removed the possibility to change the value in `ops/configuration/yii-conf/yiic.php.dist`.


kencho51 (commented):
Hi @rija,

**praise:** After deploy to staging in seoul region, `/dataset/100311` and `dataset/100314` can be loaded now, but very slowly (~3 minute loading time).

**praise:** `Gigareview` pipeline is passing is gitlab.

**praise:** Confirmed that in the AWS console the instance type for the docker_host in seoul is `t3a.small`. 

**issue:** AWS EC2 instance type `t3a.small` is  unavailable in HK region.

The changes look good to me, the slow loading speed is not a blockage as it is an expected event. But `t3a.small` instance type is unavailable in HK region, this fix could not be applied to the beta if it is deployed from HK?

pli888 (changes requested):
praise: Both /dataset/100314 and /dataset/100310 pages are loading and don't result in a time out error.

issue: The above dataset pages can still take up to several minutes to load which will still be problematic to end-users. The tech team discussed using pre-caching to resolve this problem but this infrastructure work will take time and effort to implement so will be done as future work.

praise: All tests are passing again in CI pipeline.

issue: As raised by @kencho51, `t3a.small` EC2 instance type not available in HK region. Will use `t3.small` instance type as discussed.

rija:
Hi @pli888, @kencho51,

I have updated this PR with: 

* replacing `t3a.small` with `t3.small` for now
* preload the urls for the datasets 100314 and 100310 after deployment, on staging and live 
* updated the PR's description to better reflect  the changes

kencho51 (approved):
Hi @rija,

**praise:** Confirmed that in the AWS console the instance type for the docker_host is `t3.small` after deploy to staging or live in seoul region.

**praise:**  After finished the job `sd_warmcache` or `ld_warmcache` which takes about 18mins, the dataset pages for `100314` and `100311` can be loaded much faster on staging and live.

**praise:** All the files in the component dir have been conformed the PSR-12 standardd.

**question(non-blocking):** Why after `curl staging.url/dataset/100314` in the server, the dataset page of it would be loaded much faster? `curl`just fetch the HTML content and display it in the terminal.

The code looks good to me, happy to approve it.

pli888 (approved):
**:praise** The web pages for the two datasets: 100314 and 100310 now load in seconds with them both pre-loaded in the cache. I can also see that the application EC2 server is a t3.small instance type. I approved this PR.

Issue #1102, opened 2022-08-25
unusual behavior with attributes on beta site sample update page


only1chunts:
## User story

>As an admin user
>I want to be informed by the website of erroneous attributes being used
>So that I can correct my mistake

## Acceptance criteria

>Given I am editing a sample, e.g. https://beta.gigadb.org/adminSample/update/id/1018
> And I include an attribute name that does not exist in the database
>When I click save
>Then the system informs me which attribute name is not valid and asks me to choose a valid attribute name

## Additional Info
Currently the live system fails with a nasty CDb exception.
Currently the beta site does not fail, instead, it appears to insert a new erroneous attribute into the attribute table without informing the user, this must not happen.

We should check if the same thing happens on the file admin page?

This is related to [#1088](https://github.com/gigascience/gigadb-website/issues/1088).

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
* [ ] The product is in a releasable state (i.e. the increment has not broken anything)

rija:
If this problem is real, look into foreign constraints in database schema.
They should prevent this from happening

kencho51:
In `/AdminSampleController.php`:

It seems that non-exist attribute is expected to create and save.

```
 // Get attribute, if not exist we create
$attribute = Attribute::model()->findByAttributes(array('structured_comment_name' => trim($data[0])));
if (!$attribute) {
    $attribute = new Attribute();
    $attribute->structured_comment_name = trim($data[0]);
    $attribute->attribute_name = str_replace('_', ' ', trim($data[0]));
[...]

PR #1231, opened 2023-03-08
only attribute name exists in database can be created and saved

kencho51:
# Pull request for issue: #1102 

Display error message `Attribute name $attribute  is not valid - please select a valid attribute name!` when creating a sample object or updating an existing sample object with attribute not found in `attribute` table, and do not create/save it

#### In dev

Go to http://gigadb.gigasciencejournal.com:9170/adminSample/admin after logging in as an admin, and try to update the **Attributes List** of one sample, for example http://gigadb.gigasciencejournal.com:9170/adminSample/update/id/432, to `animal="tiger"`, then click `Save`, the following error messages can then be seen at the top of of the form:
```
Please fix the following input errors:

    Attribute name for the input animal=tiger is not valid - please select a valid attribute name!
    Attribute name for the input plant=rose is not valid - please select a valid attribute name!
```

Or try to create a new sample at here, for example http://gigadb.gigasciencejournal.com:9170/adminSample/create with the **Species * ** value as `20` or any other valid one found the in `Species` table and the **Attributes List** value as  `animal="tiger"`, then click `Save`, the following error messages can then be seen at the top of of the form:
```
Please fix the following input errors:
 
    Attribute name for the input animal=tiger is not valid - please select a valid attribute name!
    Attribute name for the input plant=rose is not valid - please select a valid attribute name!

```

#### In staging and live
1. Provision gigadb-website.
```
# staging servers
% cd ops/infrastructure/envs/staging
% ../../../scripts/tf_init.sh --project gigascience/forks/kencho-gigadb-website --env staging
% terraform plan
% terraform apply
% terraform refresh
% TF_KEY_NAME=private_ip ansible-playbook -i ../../inventories webapp_playbook.yml -v
% ansible-playbook -i ../../inventories bastion_playbook.yml -e "backupDate=latest" -v

# live servers
% cd ops/infrastructure/envs/live
% ../../../scripts/tf_init.sh --project gigascience/forks/kencho-gigadb-website --env live
% terraform plan
% terraform apply
% terraform refresh
% TF_KEY_NAME=private_ip ansible-playbook -i ../../inventories webapp_playbook.yml -v
% ansible-playbook -i ../../inventories bastion_playbook.yml -e "backupDate=latest" -v
```

2. Make sure in gitlab pipeline, `build_staging`, `sd_staging`, `build_live` and `ld_live` have been successfully deployed.

3. Then follow the steps in `In dev` section to test the new feature.

#### Running the tests
 
Updated the acceptance test `AdminSample.feature` to test for inputting non exist attribute in update and create pages.

```
% docker-compose run --rm codecept run --no-redirect acceptance tests/acceptance/AdminSample.feature 
Codeception PHP Testing Framework v5.0.4 https://helpukrainewin.org

Acceptance Tests (11) -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
✔ admin page for samples: sample attributes full description (3.92s)
✔ admin page for samples: display 1 input error message when update (3.31s)
✔ admin page for samples: display error message for empty taxon id when update (3.14s)
✔ admin page for samples: display input format error when update (3.31s)
✔ admin page for samples: display 2 input error messages when update (3.66s)
✔ admin page for samples: display error message for empty taxon id when create (2.95s)
✔ admin page for samples: display error message for non numeric taxon id when create (2.98s)
✔ admin page for samples: display error message for non exist taxon id when create (3.04s)
✔ admin page for samples: display 1 input error message when create (3.84s)
✔ admin page for samples: display 2 input error messages when create (3.89s)
✔ admin page for samples: Create new sample with exist sample attribute (3.62s)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Time: 01:07.390, Memory: 10.00 MB

OK (11 tests, 35 assertions)

```

Created functional test `AdminSampleCest` to test for the non exist attribute would not be saved into the database. 

```
% docker-compose run --rm test ./vendor/codeception/codeception/codecept run functional AdminSampleCest                 
Running with seed: 


Functional Tests (9) ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
✔ AdminSampleCest: Try update sample with one non exist attribute (0.75s)
✔ AdminSampleCest: Try update sample with two non exist attribute (0.69s)
✔ AdminSampleCest: Try update sample with exist attribute (0.58s)
✔ AdminSampleCest: Try create sample with empty s taxon id (0.50s)
✔ AdminSampleCest: Try create sample with non numeric taxon id (0.52s)
✔ AdminSampleCest: Try create sample with non exist taxon id (0.51s)
✔ AdminSampleCest: Try create sample with one non exist attribute (0.56s)
✔ AdminSampleCest: Try create sample with two non exist attribute (0.53s)
✔ AdminSampleCest: Try create sample with exist attribute (0.64s)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 36.27 seconds, Memory: 14.00MB

OK (9 tests, 50 assertions)

```

The attribute name from user's input will be used to locate the `Attribute` model. If that attribute is not in the `attribute` table, there would be no `Attribute` model returned by the `updateSampleAttributes` in `AdminSampleController` class, an error would then be added to the `Sample` model.

The `Sample` model could only be created or saved if the `tax_id` is found in `Species` table. After that, the `sample id` will be available for updating the `SampleAttribute` table. 

A function `findSpeciesRecord()` is created to replace the ducplicated codes in `actionCreate()` and `actionUpdate()`, the function has also been refactored with more checking and saving errors to sample model. The duplicated codes are:
```
$species = Species::model()->findByAttributes(array('tax_id' => $tax_id));
$model->species_id = $species->id;
$model->attributesList = $_POST['Sample']['attributesList']
```

If the `Sample` model contains errors, `$form->error($model,'attributesList')` would display the error message in update or create pages if there is any error of `attributeList` in the model. If no errors, the page will be redirected a view page with the updated values.

Moreover, the `SampleAttribute` model has been updated with two custom validators created, which would check if the `attribute id`  exists in `attribute` table   and `sample id` exists in the `sample` table before saving values into the `sample_attribute` table.

The acceptance test `AdminSample.feature` has been updated to test the update and create pages.

 

Describe any problems with your implementation

None.

Describe changes to the documentation

Describe changes to code that repays any technical debt

Describe any improvements to the Gitlab pipeline

Currently, the `container_scanning` in gitlab pipeline is failing because the scanner is trying to scan the image `registry.gitlab.com/gigascience/forks/kencho-gigadb-website/sprinttask-1102-choose-a-valid-attribute-name:100872f562f9f9ca470f05fc5e3177ee29499acd` which does not exist. To make the scanner to scan the images that we want, the variable `CS_IMAGE` needs to be used for it.

And in order to allow 1 container_scanning job to scan multiple images, a `.scanner_matrix` is introduced, which specify the scanner type and the images we want to scan. The container scanning job would then become:

image

If there are any critical vulnerabilities found by the scanner, they could be found in the `Security` tab in that pipeline, for example at here: https://gitlab.com/gigascience/forks/kencho-gigadb-website/-/pipelines/995302401/security
 
## CI/CD results
- [x] All green


pli888 (commented):
**praise:** These code changes fix the problem described in issue #1102

**praise:** The new acceptance tests look fine to me and are passing.

**nitpick:** It was difficult at first to review this PR because there is a lot of noise when you look at code changes on https://github.com/gigascience/gigadb-website/pull/1231/files. On closer look of the code changes, you see that they are due to PSR-12 code formatting which I thought we agreed to do after the first round of reviewing.

**thought:** I revised what the application logic is with samples. Basically, a sample is associated to an organism species. A sample can have attributes which provides metadata about the sample, e.g. if DNA sample came from a female organism therefore the attribute will be sex:female. On the Update Sample admin page, if an attribute key value pair is inputted into the Attributes List text field, the Attribute name is checked if it exists by the updateSampleAttributes($model) function in AdminSampleController class. If it does not exist then AdminSampleController will add an error to the Sample model. Therefore, it is line 376 that provides the new functionality to add an error to the Sample object if a specific attribute name cannot be found to annotate the sample object with a sample attribute key value pair. There is also line 14 in protected/views/adminSample/_form.php which has been updated to display the warning message in red colour in the admin sample page.

**question:** Since this PR updated updateSampleAttributes($model) in AdminSampleController class, it might have been worthwhile tidying up the code in this function, for example, the variable names could be improved, e.g. $data, $sAttr and $sampleAttr as they sound ambiguous and generic. See if @rija thinks this is worth doing?

**suggestion:** Might be better to update the error message according to what is specified in the acceptance criteria: Attribute name is not valid - please select a valid attribute name.

rija (changes requested):
Hi @kencho51,

**praise**: all tests are passing, locally and on CI
**praise**: error message when inputing an non-existent error is shown and clearly visible

**issue**: the error handling doesn't follow the expected behaviour of web forms:
When an input doesn't validate, we show an error message, but we don't remove the erroneous value from the form's field and we don't allow save to proceed.  Once we've signalled an error, it's up to the user to make the appropriate correction.
There are several reasons for that:
imagine the user enters an attribute with a very long value and they makes a one character typo. With current implementation the whole attribute/value pair is deleted and they might be surprised and annoyed by that. If the implementation were following typical web forms user experience, they would have the opportunity to go back and just correct that one error typo.
Also, allowing form submission to proceed while they were error message on the form is confusing and the user may think there is a problem because that's not how web forms are supposed to function.


**suggestion**: Given the legacy nature of this part of the codebase, an OK alternative to keeping the user input intact in the form fields would be to show both attribute and value in the error message with the same format as in the field, instead of just the attribute name, so that users can at least copy and paste and then make correction.

**issue**: the validation code is not implemented in the correct area of the codebase:
Imagine we need to be able to update sample attributes from the command line. We would use the `SampleAttribute.php` model for handling persistence, but we would have to implement again the validation (which is code duplication).
Validation code always goes inside the model class being validated because the model class is the gatekeeper to the database table and it might be used in different places (web, api, CLI, automated tests).

**suggestion**: With the Yii framework, you can use a custom validator [1] in the model (there are already examples like `validateCaptcha` in `User.php`) then in `function updateSampleAttributes($model)`, line 382 becomes:

```
If (!$sampleAttribute->save()) {
	$model->addError(...);
}
```

**polish**: Legacy code issues in the `protected/controllers/AdminSampleController.php` class:

 * Agree with @pli888's comment regarding obscure variable names 
 * Also along the same line, PHPStorm is detecting duplicated code fragments in the controller class starting line 67, which suggests the duplicated block can be extracted as a private method maybe.
 * Also there are big commented blocks which `git blame` says date back from 2016, so we should delete them. They serve no purpose and are ugly and confusing
 * When creating a new sample, if the input to the Species field is missing or is not recognised, the form throws a `CDbException` error to the user upon submission (the line 67 of the controller indicated in the stack trace should be inside a try/catch block that will call addError)


**thought (non-blocking, security)**:
I was surprised to see that the "Species" field, when creating a new sample, allows the user to input the `id` of the `species` table. This is a security issue because it exposes internal database id scheme to website users and expose the web app to risks of SQL injections.
There's no fix to be done here, because it may be used by the curators (who might have commissioned that behaviour in the first place), and we should talk to them first

[1] https://www.yiiframework.com/doc/guide/1.1/en/form.model#declaring-validation-rules

kencho51:
Hi @rija and @pli888,

This PR has been updated regarding all the comments above, all the details can be found in the updated description section. So, please have a look again.

protected/controllers/AdminSampleController.php
```diff
@@ -2,126 +2,108 @@
...
+        ));
+    }
+
+    /**
+     * Creates a new model.
+     * If creation is successful, the browser will be redirected to the 'view' page.
+     */
+    public function actionCreate()
+    {
+        $model = new Sample();
```
pli888, line 57: **thought**: Instead of `$model`, I would name it `$sample` so I know what type of model object I am dealing with lower down.

protected/controllers/AdminSampleController.php
```diff
@@ -2,126 +2,108 @@
...
+        // Uncomment the following line if AJAX validation is needed
+        // $this->performAjaxValidation($model);
+
+        if (isset($_POST['Sample'])) {
+            $model->attributes = $_POST['Sample'];
+            $model->name = $_POST['Sample']['name'];
+            $array = explode(":", $_POST['Sample']['species_id']);
+            $tax_id = $array[0];
+            try {
+                $species = $this->getSpecies($tax_id, $model);
```
pli888: **praise**: I see the /adminSample/create page is now asking the website user to provide a Taxon Id to specify the species when creating a new sample record which seems safer security-wise.
**thought**: The `views/adminSample/_form` class should be updated to say that Taxon Id is required instead of `Species*`.

protected/controllers/AdminSampleController.php
```diff
@@ -205,189 +189,188 @@ public function actionChoose() {
...
+     * @param $tax_id
+     * @param $model
+     * @return array|CActiveRecord|mixed|Species|null
+     */
+    private function getSpecies($tax_id, $model)
+    {
+        $species = Species::model()->findByAttributes(array('tax_id' => $tax_id));
+        if (!$species) {
+            $model->addError('error', 'Species id ' . $tax_id . ' is not found!');
+        }
```
pli888: **thought**: This conditional statement seems like a validation rule for creating new Sample model objects such that the new sample must be of an existing species in GigaDB. In which case, this functionality should be in the Sample model class? Possibly use the unique core validator to check that an attribute value is in the species database table?

protected/controllers/AdminSampleController.php
```diff
@@ -205,189 +189,188 @@ public function actionChoose() {
...
+        $model->species_id = $species->id;
+        $model->attributesList = $_POST['Sample']['attributesList'];
+        $this->updateSampleAttributes($model);
+        if (!$model->errors) {
+            $model->save();
+            $this->redirect(array('view', 'id' => $model->id));
+        }
```
pli888, line 385: **thought**: Saving the sample model here feels out of place in this function. Should it be in the `actionCreate` function above?

protected/controllers/AdminSampleController.php
rija: Hi @pli888, @kencho51,

Validators' remit is scoped to when an object is about to be created or updated.
Once an object is persisted to the database, it's no longer the validators' jurisdiction.
Getters must assume that the data they try to retrieve has gone through validation already,
so I'd say there is no need to add any additional model validation here.

The name of the function (`getSpecies`) is confusing though because it's not a real getter:

Getters don't change the state of the database (that's the role of setters), they just "get", and this function is doing a `$model->save()` inside it on line 372 and calling `$this->updateSampleAttributes($model)` on line 369.

I think you should take those two lines that change the database state outside that function, and looking at what's left,
I actually think there is no point to its existence as its body can just be put as-is in the parent `try` block.

I think I understand what you were trying to do but the `actionCreate()` here is not big, so there's not much benefit in trying to factor out its content.

protected/controllers/AdminSampleController.php
rija: yes, agreed, see my comment above about getters and setters

rija (changes requested):
Hi @kencho51,

**praise**: the workflow issues I raised have been corrected
**issue**: the `getSpecies` function in `protected/controllers/AdminSampleController.php` has a few problems:
* the phpdoc block doesn't explain what the parameters are for in @parameter lines. (Look at the other functions in that same class for good examples) and the declared @return type doesn't match what is actually returned
* the function is missing type declaration in its signature: see https://www.php.net/manual/en/language.types.declarations.php
* the function name start with `get*` which is the convention for getters but it modifies the database state which getters must not do
*  it doesn't bring much value to the already small `actionCreate()` function, and the  extra level of indirection make `actionCreate()` less readable, not helped by the confusing name

**suggestion**: get rid of the `getSpecies` function, as its body is fine directly inside the `try` block of `actionCreate()`
**issue**: when inputing an attribute that doesn't exist already, it should only display the appropriate error message `- Attribute name for the input animal=tiger is not valid - please select a valid attribute name!`, currently it also displays 6 additional error messages that are only consequences of the first issue, that's very confusing
**suggestion**: In `protected/controllers/AdminSampleController.php`, in the function `updateSampleAttributes()`, the block after `// Let's save the new sample attribute` should not be executed if `$attribute` is empty

protected/controllers/AdminSampleController.php
kencho51: This $model is declared inside  `AdminSampleController` class, which makes it clear that it is referred to the `Sample` model, and I think all `$model` in this controller class is referred to the `Sample` model as well.
So, I would better not change it, otherwise many `$model` in this class have to be re-declared.

kencho51: This $model is declared inside  `AdminSampleController` class, which makes it clear that it is referred to the `Sample` model, and I think all `$model` in this controller class is referred to the `Sample` model as well.
So, I would better not change it, otherwise many `$model` in this class have to be re-declared.

protected/controllers/AdminSampleController.php
kencho51: The `Species*` label is clearer, so I want to keep it,  but I have updated the placeholder to `Taxon Id` as a reminder to users.

kencho51:
Hi @rija,

> **suggestion**: get rid of the `getSpecies` function,

The  private `getSpecies()` function was supposed to replace the ducplicated codes found in the `actionCreate()` and `actionUpdate()`, and the name was given by the IDE when I tried to refactor it using phpStorm.

Now the function has been refactored and renamed to `findSpeciesRecord()` which will not modify database.

rija (approved):
Hi @kencho51,

**praise**: code structure remarks have been fixed
**praise**: user experience when inputing non-existent attribute is improved
**praise**: good thinking regarding changes to remove duplication
**nitpick** (non-blocking):  Both `phpcs --standard=PSR12` and my IDE still throw a bunch of errors and warnings on `protected/controllers/AdminSampleController.php`.php, but I reckon it's not worth pursuing this right now
**praise**: All tests are passing


Based on the above, I'm happy to approve

pli888 (approved):
**praise**: Functionality looks good now that only a single error message is display if there is an attribute problem inputted by user
**praise**: Code looks good now.
**praise**: Functional and acceptance tests updated by this PR are passing.

Issue #1428, opened 2023-09-14
4b. Move the saving of EC2 ip address to ansible_init.sh


rija:
We need to move the update of the gitlab variables that represent EC2  instances' ip addresses  into `ansible_init.sh` to improve deployment reliability when there are issues with re-running terraform or the Ansible playbook

PR #1464, opened 2023-09-28
Ansible init to upload EC2 IP address to GitLab variables

rija:
# Pull request for issue: #1428

* Create a distinct bootstrap_playbook.yml to save EC2 IP addresses to GitLab variables
* Update ansible_init.sh to run the bootstrap playbook
* Clean up most of the warnings thrown when executing Ansible playbooks

### staging

Destroy and recreate a AWS staging environment.
```
$ ../../../scripts/tf_init.sh --project \<your gitlab project name\> --env staging
$ terraform plan
$ terraform apply
$ terraform refresh
$ terraform output

```
Take note of the IP addresses shown in the output.

Force (-f) reinstall all the third party Ansible roles:
```
$ ansible-galaxy install -fr ../../../infrastructure/requirements.yml 
```

Then run `ansible_init.sh`
```
```
There should be no error. Ansible ok-executed the first time tasks and skipped the subsequent ones.

Go to Gitlab, your fork's CI/CD settings, you should now see those IP addresses in GitLab variables for your fork's staging under the variables: 
* `remote_private_ip`
* `remote_public_ip`
* `remote_bastion_private_ip`
* `remote_bastion_public_ip`

Now, run `ansible_init.sh` again
```
```

There should be no error, Ansible ok-executed all tasks from the playbook.
The same values should still be seen in GitLab variables.

Notice that Ansible doesn't throw any warnings at all while executing `bootstrap_playbook.yml`

You can then continue provisioning the environment as usual:
```
$ env TF_KEY_NAME=private_ip OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES  ansible-playbook -i ../../inventories webapp_playbook.yml
$ env OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES ansible-playbook -i ../../inventories bastion_playbook.yml -e "backupDate=latest"
```

You will also notice that some of the warnings that used to be thrown by Ansible with both playbooks have now been corrected.

### live

Repeat the previous section with the live environment, the result should be the same.

Moved the tasks that were in `ops/infrastructure/roles/docker-postinstall/tasks/main.ym` that save EC2 IP addresses to GitLab into a new role `ops/infrastructure/roles/tf-output-to-gl-variables/tasks/main.yml` which is then used in a distinct, dedicated, playbook `ops/infrastructure/bootstrap_playbook.yml`.
The playbook uses the role twice, to make two tagged plays, once for the webserver IP addresses, and once for the bastion IP addresses.

We then make two calls to the playbook with the appropriate tags at the end (it must be done after the calls to `ssh-keygen` otherwise it won't work) of `ops/scripts/ansible_init.sh`.

The playbook `bootstrap_playbook.yml` can be executed without warnings as it's new playbook which is called 
from a shell script rather than human operator.

However, executing the `bastion_playook.yml` or `webapp_playbook.yml` playbooks still throw some warnings because:

* A  warning is thrown by third party roles which is already at their latest versions
  * See: https://github.com/prometheus-community/ansible/issues/144 
* the warning about pattern not being matched could be fixed the same way it has been done in `bootstrap_playbook.yml`, but it wasn't done as that file is something @kencho51 is working on currently on its current PR and it requires passing an extra variable to the call to the playbook (` --extra-vars="gigadb_env=staging"`) which requires updating SOP which is also something @kencho51 is working on at the moment.

Fixing "broken windows": Remove the warnings thrown while executing Ansible (all of them for the new bootstrap playbook and some of them for the previous web app and bastion playbooks)

### before

```
[WARNING]:  * Failed to parse /Users/owner/code/gigadb-website/ops/infrastructure/inventories/hosts with yaml plugin: We were unable to read either as JSON nor YAML, these are the errors we got from each: JSON: Expecting value: line 1 column 2
(char 1)  Syntax Error while loading YAML.   did not find expected <document start>  The error appears to be in '/Users/owner/code/gigadb-website/ops/infrastructure/inventories/hosts': line 2, column 1, but may be elsewhere in the file
depending on the exact syntax problem.  The offending line appears to be:  [monitoring_hosts] monitoring ansible_host="{{ lookup('ini', 'MONITORING_HOST type=properties file=../../../../.env') }}" ^ here We could be wrong, but this one looks
like it might be an issue with missing quotes. Always quote template expression brackets when they start a value. For instance:      with_items:       - {{ foo }}  Should be written as:      with_items:       - "{{ foo }}"
[WARNING]:  * Failed to parse /Users/owner/code/gigadb-website/ops/infrastructure/inventories/hosts with ini plugin: /Users/owner/code/gigadb-website/ops/infrastructure/inventories/hosts:17: Section [type_aws_instance:vars] not valid for
undefined group: type_aws_instance
[WARNING]: Unable to parse /Users/owner/code/gigadb-website/ops/infrastructure/inventories/hosts as an inventory source
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
[DEPRECATION WARNING]: "include" is deprecated, use include_tasks/import_tasks instead. See https://docs.ansible.com/ansible-core/2.15/user_guide/playbooks_reuse_includes.html for details. This feature will be removed in version 2.16. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[WARNING]: Collection prometheus.prometheus does not support Ansible version 2.15.2
[WARNING]: Could not match supplied host pattern, ignoring: name_gigadb_server_staging*
```

### after

```
[WARNING]: Collection prometheus.prometheus does not support Ansible version 2.15.2
[WARNING]: Could not match supplied host pattern, ignoring: name_gigadb_server_staging*
```

This PR makes the deployment more resilient by shielding pipeline deployment from potential malfunctions in the early stages of web app and bastion playbook execution


ops/scripts/ansible_init.sh
```diff
@@ -121,3 +122,8 @@ ssh-keygen -R $webapp_ip
 ssh-keyscan -t ecdsa $bastion_ip >> ~/.ssh/known_hosts
 new_host=$(ssh -i $aws_ssh_key centos@$bastion_ip ssh-keyscan -t ecdsa $webapp_ip)
 echo $new_host  >> ~/.ssh/known_hosts
+
+# Bootstrap playbook
+echo "Saving EC2 IP addresses to GitLab"
+env TF_KEY_NAME=private_ip ansible-playbook -i ../../inventories bootstrap_playbook.yml --tags="webapp_ips" --extra-vars="gigadb_env=$target_environment"
+ansible-playbook -i ../../inventories bootstrap_playbook.yml --tags="bastion_ips" --extra-vars="gigadb_env=$target_environment"
```
kencho51, line 129: Hi @rija,

I am thinking the bootstrap playbook can only be executed 1 time only if the `--tags` option is removed, like this:
```
env TF_KEY_NAME=private_ip ansible-playbook -i ../../inventories bootstrap_playbook.yml --extra-vars="gigadb_env=$target_environment"
```
And the output will be:
```
PLAY RECAP ************************************************************************************************************************************************************************************************************************
webapp_private_ip                : ok=5    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
bastion_private_ip                : ok=5    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```
The only difference is that user needs to allow the connection to bastion server through private IP at the first time.

I understand why you would execute the `ansible-playbook` 2 times, as bastion server does not need private ip for connection, but it would help by reducing the execution time and have a more informative `PLAY RECAP` at the end.

It is just a thought and I would like to see what you think.

kencho51 (approved):
Hi @rija,

**praise:** The EC2 IPs are put to gitlab after ansible init.
**praise:** All ansible warnings have been addressed, the only two left warnings: `[WARNING]: Collection prometheus.prometheus does not support Ansible version` has been reported already, while `[WARNING]: Could not match supplied host pattern, ignoring: name_gigadb_server_*` will be fixed in #1475.
**praise:** It is a great idea to supply `--extra-vars` option to the playbooks, so the host will be specific, and can remove the redundant host.
**issue(non-blocking):** The boostrap playbook will be executed 2 times in the ansible init step, which can be executed once only if the `--tags` option is removed, please see my comment for details.
 
Code changes LGTM, so happy to approve.

ops/scripts/ansible_init.sh
rija: Hi @kencho51,

I understand what's bothering you with this approach, and it does annoy me too.
It's a limitation of the dynamic inventory script we are using that can only return either private ips for all hosts or public ips for all hosts, not mix-and-match.
I should have probably listed that as a "known issue" too.

1) In the case of this PR, the Ansible playbook calls are not meant to be made manually and separately, and performance overhead is negligible, so I deem the situation acceptable


2) The most promising solution to the issue involves changing the dynamic inventory script, but:
* that would make this PR bigger than it should
* it should be done as a way to also fix the need to set `TF_KEY_NAME` env to the `webapp_playbook.yml` as the root cause is the same
* If solution is still built on the open source terraform-inventory project we are using, we should aim to submit a PR to that project

So, because of 1)  and 2), better leave this for another time

pli888 (approved):
**praise**: I confirm Gitlab variables for EC2 IP addresses are now updated by `ansible_init.sh` for both staging and live environments.
**praise**: There is a reduction in the number of warning messages when running playbooks.
**praise**: Code looks good to me - I approve this PR.

PR #1516, opened 2023-10-24
Sample attributes list lost and not saved correctly

rija:
# Pull request for issue: #1102

This is a pull request for fixing the following problematic situation:

1) After an invalid entry the form shows in the field what's from the database instead of what was inputed by the user
2) When the form is submitted, whether or not there are errors in the attribute list, only the last valid attribute is saved in the database
3) Because all sample attributes in the databases are deleted first after form is submitted,  if there is an error in the list of attributes, due to 1) and 2), we lose existing attributes from the database with no way to recover them

The fix will ensure that:
*  the data entered by user in the from stays in the form after updating with error
* all valid sample attributes will be saved in the database upon form submission

## How to test?

I updated the scenarios in `tests/acceptance/AdminSample.feature` to cover the situation describe above.
All the tests should pass.

### manually

* Navigate to http://gigadb.gigasciencejournal.com:9170/adminSample/update/id/154
* to the existing list of attributes in the "Attributes List" field, add `animal="tiger"` and submit the form
* You should see:
  * `- Attribute name for the input animal=tiger is not valid - please select a valid attribute name!`
  * the content of the  "Attributes List" field should stay the same as when the form was submitted (including the invalid attribute)
  
Then two parallel tracks to test:
1) Click "Cancel" and return to http://gigadb.gigasciencejournal.com:9170/adminSample/update/id/154, all the valid attributes from the database should be  listed in the "Sample Attributes" column and navigating to the form will show them all in the "Attributes List" field
2) Replace the `animal="tiger"` with `age="5"` then submit again, you should be on http://gigadb.gigasciencejournal.com:9170/adminSample/view/id/154 and the original attributes and the new one should be listed in the view

The most critical issue was in `protected/controllers/AdminSampleController.php`'s `updateSampleAttributes` function where new sample attribute CActiveRecord models are not created correctly.

The line 
```
$sampleAttribute = clone $sampleAttribute;
```
Will not work properly as it reference the same variable name on both side of the equal sign
Furthermore that variable is declared outside of the loop with:
```
$sampleAttribute = new SampleAttribute();
```

So the cloning will create a new object but assign it to the same original variable, at every iteration,
So that when calling `->save()` inside the loop, it keeps saving the same model, and of course the last save win, which is why the latest attribute appears to be the only saved. 
There were all actually saved but into the same sample attribute model so previous save were overwritten.

So the fix is to ensure that the clone is assigned to a variable declared inside the loop, so that at every iteration, a new variable is used.

The other issue is related to what is displayed in the "Attributes list" field after errors are reported.
Previously, it load what's in the database, but that's incorrect. It should show the data that was entered by the user, so that they can correct it by themselves.

n/a

Augmented the acceptance tests scenario in `tests/acceptance/AdminSample.feature` to reproduce the two issues.

n/a

n/a

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
+
 # GigaDB Changelog
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-
+- Fix #1102: On sample admin form, save all valid sample attributes while showing errors for non-existent attributes
 - Feat #1434: Setup playwright local environment for automated accessibility testing
 - Feat #1443: Make ansible playbooks can execute plays separately by adding tags
 - Fix #1428: Increase resilience of provisioning by extracting saving EC2 IP addresses as standalone bootstrap plays
--- a/protected/controllers/AdminSampleController.php
+++ b/protected/controllers/AdminSampleController.php
@@ -333,9 +333,9 @@ class AdminSampleController extends Controller
 
         if (!empty($model->attributesList) && trim($model->attributesList)) {
             // From a model we will clone
-            $sampleAttribute = new SampleAttribute();
-            $sampleAttribute->sample_id = $model->id;
             foreach (explode('",', $model->attributesList) as $attributes) {
+                $sampleAttribute = new SampleAttribute();
+                $sampleAttribute->sample_id = $model->id;
                 $attributes = str_replace('"', '', $attributes);
                 $attributeData = explode('=', $attributes);
                 if (count($attributeData) == 2) {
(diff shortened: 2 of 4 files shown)


kencho51 (approved):
Hi @rija,

**praise:** The acceptance tests are passing
**praise:** The updated logic in the `updateSampleAttributes()` is now better fit for the requirements, and the original sample attributes are preserved.
**praise:**  Follow the `How to test` manually, all steps are passing and all the expect results are observed.
**question(non-blocking):** There is two entries of `Fix #1102:` in the change log, but with difference description, which may cause confusion in future. Therefore, would it be better if we include PR number to each entry? For example, `- Fix #1102 (PR #1516): On sample admin form, save all valid sample attributes while showing errors for non-existent attributes`, this will help to get the commit hash during the release process.

Code changes LGTM, happy to approve.

rija:
> Hi @rija,
> 
**question(non-blocking):** There is two entries of `Fix #1102:` in the change log, but with difference description, which may cause confusion in future. Therefore, would it be better if we include PR number to each entry? For example, `- Fix #1102 (PR #1516): On sample admin form, save all valid sample attributes while showing errors for non-existent attributes`, this will help to get the commit hash during the release process.

Hi @kencho51,

I understand your thinking, however:

* The PRs are a contract between us developers, the CHANGELOG is for all the business to see and use and usually they are the ones who created the Github issues initially (and in all cases, that's their reference for the work), making issue number the only really relevant one in the CHANGELOG. 
* Furthermore, adding both numbers make the CHANGELOG messy as they share the same #-prefix and may confuse people as to which one to follow through when reading hurryingly.
* The fact that a given issue takes several features/fixes contributions to implement/resolve is a good information to know and reason about in roadmap workshops and retrospectives, so it's actually helpful
* I expect these situations to be quite rare anyways
* Updating CHANGELOG is already a chore that we occasionally forget, so adding more rules is not going to help 

So I think it's better to stick to issue numbers only.

Issue #1528, opened 2023-10-30
updates made in admin pages not showing in mockup


only1chunts:
## User story

>As a curator
>I want to see my changes appear in the pre-publication view of the dataset page
>So that I can confirm the changes are correct and show the relevant users the private mockup page displaying the correct information

## Acceptance criteria

>Given I make an update to any dataset metadata in the admin pages
>When I click view "open private URL"
>Then the dataset page is displayed at the private token url with those changes displayed

>Given I make an update to any sample metadata in the admin pages
>When I click view "open private URL"
>Then the dataset page is displayed at the private token url with those changes displayed

>Given I make an update to any file metadata in the admin pages
>When I click view "open private URL"
>Then the dataset page is displayed at the private token url with those changes displayed

>Given I make an update to any author metadata in the admin pages
>When I click view "open private URL"
>Then the dataset page is displayed at the private token url with those changes displayed

## Additional Info
may be related to #963 

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
* [ ] The product is in a releasable state (i.e. the increment has not broken anything)

rija:
Additional notes: 
@only1chunts can reproduce it on Chrome.
Neither Yannan, nor @tuli have noticed the issue

rija:
Hi @only1chunts

1) The main problem: 

I was able to replicate the experiment you showed us during last meeting that reproduce the issue whereby the mockup is not reflecting the changes to the database.

The reason it is happening is because of the way the application decide wether the cached data has expired or not.
[...]

PR #1560, opened 2023-11-07
No caching on mockup page

rija:
# Pull request for issue: #1528

*  Do not use the cache when loading mockup pages
* Always enable caching by default on the development environment

### Automatically

After checking out this branch and starting Gigadb website with `up.sh`,
you can run the automated tests with:
```
$ docker-compose run --rm codecept run --no-redirect --debug acceptance tests/acceptance/DatasetMockupPage.feature
```

### Manually

After login as an admin, navigate to the dataset admin form http://gigadb.gigasciencejournal.com:9170/adminDataset/update/id/668 and click-open on a separate tab the "Open Private URL" link to show the mockup page on a separate tab, then manually perform the acceptance criteria listed in #1528 as guided below

In parallel, do open the Docker Desktop dashboard, ensure you are on the "Containers" screen, and click on the "deployment" stack in the list, then in the list of services from that stack, click on "deployment-application-1", you will see the log of that container service. Make sure it's empty by clicking the trash icon "Clear terminal" on the right  before you run each test. The log should not show anything log entries related to caching (HIT or MISS) after your ran the test (This work for both manual and automated testing).

#### changing dataset metadata

On the dataset update form, just change the content of the title or the description then save, then refresh the mockup page, it should show the changes.

### changing samples metadata 

From the admin dashboard, navigate to the "Dataset:Samples", click the "Edit" button for the first sample and assign it to the "200070" DOI, then refresh the mockup page, it should show the sample in the samples tab.

### changing files metadata 

From the admin dashboard, navigate to the "Dataset:Files", click the "Edit" button for the first file and assign it to the "200070" DOI, then refresh the mockup page, it should show the file in the files tab

### changing author metadata 

From the admin dashboard, navigate to the "Dataset:Authors", click the "Edit" button for the first author and assign it to the "200070" DOI, then refresh the mockup page, it should show the author name in the main section.

### On AWS Deployment

After pushing this branch remotely to you fork, and deploying the resulting pipeline to your staging on AWS, you can reiterate the manual testing steps above on your staging environment

### Enabling caching by default on development environment

Motivations:

* In general, there should no be no reasons for caching to be disabled locally. If there are issues locally that we think is related to caching, chances are there are also happening on production and should be investigated and fixed. Having cache disabled locally increases the risks of not discovering bug in caching before they hit production.
* Locally, if we really need to disable caching, we can do so by replacing `CApcCache` with `CDummyCache` in the Yii configuration file, which is the clean way of doing it
* It makes testing cache related feature easier

### Disabling caching for mockup page

The gathering of the the various sets of information that make a dataset pages is done in the class `DatasetPageAssembly.php`, before being summoned in the `actionView()` method of `DatasetController.php`

It relies on an onion-like architecture that allows:
*  keeping the various concerns related to bringing the data to the users (formatting, authorisation, caching, database query) independent of each other so that if need to change one of them, there is a dedicated layer to go and change without having to touch any other layers
* keeping each domain object (the things we care and talk about as a business: dataset, samples, files,...) in their own vertical structure, so that if a domain object need internal change, only the classes that implement the domain interface are impacted by the change and the calling code (data assembly and controller) should be oblivious to this.
* flexibility that comes from using interfaces. When constructing the layers in  `DatasetPageAssembly.php`, we can use any object that implement the domain interface, so that we can skip and add  layers as needed

The layers that we have are 
```
Formatting -> Authorisation -> Caching -> Retrieval from the database
```
>note: There is a good argument that the caching should be the first layer rather than being in the middle, but that's what we have at the moment. And we are not going to change that here.

Not all domain objects use all the layers (E.g: Authorisation is only used for a handful of them), but at the moment, in `develop`, they all use the Caching layer when displaying the dataset page and the mockup page.

The goal of this PR is to distinguish the code path for showing the dataset page and for showing the mock-up page, and in the latter case, the caching layer should not be used.

This is done by introducing a new configuration option called `skip_cache`, that we wrap in a configuration array passed as parameter to the `assemble()` static method and then to the constructor of `DatasetPageAssembly.php`.

If we take the example of the dataset domain object.
On `develop`, whether we show the dataset page or the mockup page, the assembly will construct the following "onion":
```
$this->_mainSection = new FormattedDatasetMainSection(
            new CachedDatasetMainSection(
                $this->_app->cache,
                $this->_cacheDependency,
                new StoredDatasetMainSection(
                    $this->_dataset->id,
                    $this->_app->db
                )
            )
```

What we want is to keep that for showing the public dataset page, but for the mockup page we want to skip the `CachedDatasetMainSection` layer and do something like:

```
$this->_mainSection = new FormattedDatasetMainSection(
                new StoredDatasetMainSection(
                    $this->_dataset->id,
                    $this->_app->db
                )

```

We can actually do that because all the `*MainSection.php` classes implements the same interface `DatasetMainSectionInterface.php`

And we do the same for all the other page components.

 

N/a

There is a new feature file that implements the user story and acceptance criteria scenarios from #1528:

tests/acceptance/DatasetMockupPage.feature 

The steps that implement the scenarios are in:

tests/_support/Steps/CuratorSteps.php

The one thing to note when coming up with the steps to reproduce the issue are to ensure all the scenarios load the mockup page first. Before the fix is implemented, such action will trigger the page components to be cached, that's why in the "Background" section, we make sure the mockup page is loaded.

The other aspect worth commenting about implementation is that we don't want to create steps that go through the web form to create the changes:
* because introducing changes in the database directly is enough to reproduce the issue
* in real situation, that's how the problem first manifest to users anyway, after the spreadsheet uploader and post-upload script have in sequences updated the database directly.
  

N/a

Once the main purpose of the PR is approved, I will perform the following tech debts repayments:

~~* [ ] Remove the index action from all controllers as following a question from Luis, curators confirmed those endpoints are not needed~~
* [ ] Replace the `Attribute.php` class implementing the Attribute domain object with `Attributes` as PHP 8.0 introduces a non-namespaced `Attribute` class to manage PHP Attributes that combined with other changes/deprecation in PHP 8.1/8.2 causes a [server error](https://github.com/yiisoft/yii/issues/4539) due to conflict between the two classes when running GigaDB against PHP 8.2. 
* [ ] Remove `google/apiclient` as it is not used by feature used by users and will cause dependencies issues with PHP 8.2

N/a

### DIFF
--- a/ops/configuration/yii-conf/index.dev.php.dist
+++ b/ops/configuration/yii-conf/index.dev.php.dist
@@ -4,7 +4,7 @@
 # To be effective, ensure the two lines below are placed before we load Yii.php
 define('YII_DEBUG',$YII_DEBUG);
 define('YII_TRACE_LEVEL',$YII_TRACE_LEVEL);
-define('DISABLE_CACHE',$DISABLE_CACHE);
+define('DISABLE_CACHE',false);
 
 # Load the modified root class for Yii1.1/Yii2.0
 $yii=dirname(__FILE__).'/protected/components/Yii.php';
--- a/protected/components/DatasetPageAssembly.php
+++ b/protected/components/DatasetPageAssembly.php
@@ -36,6 +36,7 @@ class DatasetPageAssembly extends yii\base\Component
     private $_dataset;
     private $_searchForm;
     private $_fileUploadService;
+    private $_skip_cache;
 
 
     public function __construct(Dataset $dataset, CApplication $app, FileUploadService $srv, $config = [])
@@ -44,6 +45,8 @@ class DatasetPageAssembly extends yii\base\Component
         $this->_app = $app;
         $this->_fileUploadService = $srv;
         $this->_cacheDependency = new CDbCacheDependency();
+        $this->_skip_cache = (bool)$config['skip_cache'];
+
     }
 
     /**
@@ -52,11 +55,12 @@ class DatasetPageAssembly extends yii\base\Component
      * @param Dataset $d Dataset instance to pass to the instanciated Page assembly
      * @param CApplication $app Yii web application from which to access cache and database
      * @param FileUploadService $srv GigaDB client to File Upload Wizard API
+     * @param ?array $config Additional options to configure teh assembly (only 'skip_cache' for now)
      * @return DatasetPageAssembly  a new instance of DatasetAssembly
      */
-    public static function assemble(Dataset $d, CApplication $app, FileUploadService $srv): DatasetPageAssembly
+    public static function assemble(Dataset $d, CApplication $app, FileUploadService $srv, array $config = null): DatasetPageAssembly
     {
-        return new DatasetPageAssembly($d, $app, $srv);
+        return new DatasetPageAssembly($d, $app, $srv, $config);
     }
 
     /**
@@ -75,16 +79,26 @@ class DatasetPageAssembly extends yii\base\Component
      */
     public function setDatasetSubmitter(): DatasetPageAssembly
     {
-        $this->_submitter = $datasetSubmitter = new AuthorisedDatasetSubmitter(
-            $this->_app->user,
-            new CachedDatasetSubmitter(
-                $this->_app->cache,
-                $this->_cacheDependency,
-                new StoredDatasetSubmitter(
+        switch($this->_skip_cache) {
+            case true:
+                $dataSource = new StoredDatasetSubmitter(
                     $this->_dataset->id,
                     $this->_app->db
-                )
-            )
+                );
+                break;
+            default:
+                $dataSource = new CachedDatasetSubmitter(
+                    $this->_app->cache,
+                    $this->_cacheDependency,
+                    new StoredDatasetSubmitter(
+                        $this->_dataset->id,
+                        $this->_app->db
+                    )
+                );
+        }
+        $this->_submitter = $datasetSubmitter = new AuthorisedDatasetSubmitter(
+            $this->_app->user,
+            $dataSource
         );
         return $this;
     }
@@ -105,17 +119,28 @@ class DatasetPageAssembly extends yii\base\Component
      */
     public function setDatasetAccessions(): DatasetPageAssembly
     {
-        $this->_accessions = new FormattedDatasetAccessions(
-            new AuthorisedDatasetAccessions(
-                $this->_app->user,
-                new CachedDatasetAccessions(
+        switch($this->_skip_cache) {
+            case true:
+                $dataSource = new StoredDatasetAccessions(
+                    $this->_dataset->id,
+                    $this->_app->db
+                );
+                break;
+            default:
+                $dataSource = new CachedDatasetAccessions(
                     $this->_app->cache,
                     $this->_cacheDependency,
                     new StoredDatasetAccessions(
                         $this->_dataset->id,
                         $this->_app->db
                     )
-                )
+                );
+        }
+
+        $this->_accessions = new FormattedDatasetAccessions(
+            new AuthorisedDatasetAccessions(
+                $this->_app->user,
+                $dataSource
             ),
             'target="_blank"'
         );
@@ -140,15 +165,26 @@ class DatasetPageAssembly extends yii\base\Component
      */
     public function setDatasetMainSection(): DatasetPageAssembly
     {
-        $this->_mainSection = new FormattedDatasetMainSection(
-            new CachedDatasetMainSection(
-                $this->_app->cache,
-                $this->_cacheDependency,
-                new StoredDatasetMainSection(
+        switch($this->_skip_cache) {
+            case true:
+                $dataSource = new StoredDatasetMainSection(
                     $this->_dataset->id,
                     $this->_app->db
-                )
-            )
+                );
+                break;
+            default:
+                $dataSource = new CachedDatasetMainSection(
+                    $this->_app->cache,
+                    $this->_cacheDependency,
+                    new StoredDatasetMainSection(
+                        $this->_dataset->id,
+                        $this->_app->db
+                    )
+                );
+        }
+
+        $this->_mainSection = new FormattedDatasetMainSection(
+            $dataSource
         );
         return $this;
     }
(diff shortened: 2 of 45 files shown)


protected/components/DatasetPageAssembly.php
```diff
@@ -52,11 +55,12 @@ public function __construct(Dataset $dataset, CApplication $app, FileUploadServi
      * @param Dataset $d Dataset instance to pass to the instanciated Page assembly
      * @param CApplication $app Yii web application from which to access cache and database
      * @param FileUploadService $srv GigaDB client to File Upload Wizard API
+     * @param ?array $config Additional options to configure teh assembly (only 'skip_cache' for now)
```
kencho51: **typo:** `teh` should be `the`?

tests/acceptance/DatasetMockupPage.feature
kencho51, line 1: **question:** Why after every `docker-compose run --rm codecept run --no-redirect --debug acceptance tests/acceptance/DatasetMockupPage.feature`, a new empty file with name `sed*` will created in `/protected/config`?

ops/configuration/yii-conf/index.dev.php.dist
```diff
@@ -4,7 +4,7 @@
 # To be effective, ensure the two lines below are placed before we load Yii.php
 define('YII_DEBUG',$YII_DEBUG);
 define('YII_TRACE_LEVEL',$YII_TRACE_LEVEL);
-define('DISABLE_CACHE',$DISABLE_CACHE);
+define('DISABLE_CACHE',false);
```
kencho51, line 7: **suggestion:** Since variable `DISABLE_CACHE` has been hard-coded, so the variable `DISABLE_CACHE` declared in `.env` is redundant and can be removed? 
**suggestion:** Should we hardcode the `DISABLE_CACHE` in `index.staging.php.dist`  and `index.live.php.dist` to `false`, as caching should be always `true` for the production environments as suggested in this PR?

tests/acceptance/DatasetMockupPage.feature
rija: Hi @kencho51,

I cannot reproduce what you are observing when running that command line. And it doesn't make sense. 
There is nothing in the steps implementation of that feature file that write to the file system, let alone the config directory. The Acceptance tests config file, nor the bootstrap script do not do that either.

The only time something is written to `protected/config` is when running `ops/scripts/generate_config.sh`, directly or as part of `up.sh`

Are you sure it's when running that command exactly spell like that ? does it happen when you switch to another branch?

ops/configuration/yii-conf/index.dev.php.dist
rija: Hi @kencho51,

Good thinking, but not the right thing to do here in our case.
On production environments, we need to be able to enable/disable any piece of infrastructure on demand. (e.g: in the case of caching, what if the cache is full, or some bugs not due to caching but made worse by caching happen, etc...)
Environment variables are the way to do that, given that we've adopted an immutable deployment approach (i.e: the codebase+configuration is immutable, so we cannot - and we don't want to do that anyways - connect to the machine and change manually the config). 

So on production, we still need to have this variable coming from the environment, and the functionality being responsive to its value.

tests/acceptance/DatasetMockupPage.feature
kencho51: Hi @rija,

I could reproduce this problem even after switching to `develop` branch, so it is not related to this PR.
And from here: https://forums.docker.com/t/sed-couldnt-open-temporary-file-xyz-permission-denied-when-using-virtiofs/125473, the reason for this is because I was using `VistioFS` in my docker desktop setting, the problem has been gone after I changed it to `gRPC FUSE`.

FYI:
My docker desktop version is `4.24.2 (124339)`.

kencho51 (approved):
Hi @rija,

**praise:** Great to have `skip_cache` variable to separate caching mechanism for mockup page from the main application.
**praise:** Both manual and automatic tests are passing on my local and staging environments.
**praise:** The implementation of switch case in the DatasetPageAssembly class is elegant as it plays the critical control of the caching.

Given all my questions haven been addressed and the code changes LGTM, Im happy to approve this PR.

pli888 (approved):
**praise**: The required functionality is working such that        any changes to dataset, samples, files and author metadata are immediately displayed on mock dataset pages, on both my dev and staging environments.
**praise**: The code looks good to me too. I can see that DatasetController view function checks if a mockup page dataset page is being assembled, it is configured to skip cache. The DatasetPageAssembly class then takes note of the caching configuration when retrieving data for domain objects when assembling dataset pages.

rija:
Hi @pli888, @kencho51,

I've made two additional commits as part of tech debt repayments.
With those two simple commits, I've removed some (not all) of the issues encountered when trying to run GigaDB under PHP 8.2.

It's probably easier to read through each of the commits: [5f95baf](https://github.com/gigascience/gigadb-website/pull/1560/commits/5f95baf8bc486f302efbef186463e7471ec6ab76) and [31c1d2b](https://github.com/gigascience/gigadb-website/pull/1560/commits/31c1d2b8c415dc5768e9d5a171257cecc848743e)

kencho51:
Hi @rija,

To confirm that: 
**prasie:** `google/apiclient-services ` and `google/apiclient (v2.2.2)` have been removed from the composer packages
```
+ docker-compose exec -T application composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Package operations: 0 installs, 0 updates, 5 removals
  - Removing phpseclib/phpseclib (2.0.32)
  - Removing google/auth (v1.16.0)
  - Removing google/apiclient-services (v0.203.0)
  - Removing google/apiclient (v2.2.2)
  - Removing firebase/php-jwt (v5.4.0)
 0/5 [>---------------------------]   0%
 4/5 [======================>-----]  80%
 5/5 [============================] 100%
```
**praise:** after update the model name from `Attribute` to `Attributes`, all tests (unit, functional, acceptance) are passing.

pli888:
**praise**: The code from the two commits look good to me and I can get all tests passing on my dev setup.
**question**: What at the remaining PHP 8.2 issues in GigaDB?

Issue #1680, opened 2024-02-12
MUST Remove dependency on Docker-PHP as it is abandoned and prevent developers using Linux to run FUW workflow


rija:
## Problems

 The current implementation of the feature allowing creation of dropbox FTP account by admin user from GigaDB admin pages:
* Rely on an abandoned, unmaintained open-source dependency
* Prevents developers using Linux as Development environment to execute the FUW workflow

## User Story

>As a developer
>I want to be able to execute file upload workflow from my Linux dev environment 
>And be confident of the maintainability and security of the codebase's dependency
>So that the quality of our software release is not compromised
>And I am not blocked

## Acceptance criteria

>Given I deploy and run GigaDB on my Linux environment
>When I execute file upload workflow
>Then I don't get errors relate to inter-container communication

## Additional Infos

`DockerManager.php` class needs to be deleted and the `docker-php/docker-php` dependency needs to be removed, and we need to implement a new way to manage FTP accounts from the web application.

There are two potential approaches:

1. Install pure-ftpd client in the fuw backend container so the FUW admin API directly use pure-ftpd commands to manage accounts
2. ~~Configure Pure-ftpd to use PostgreSQL for account management and have FUW admin API to interact with the database for account management~~

At the moment, the ftp server run on the same server as the web application, but because we will eventually deploy the ftp sever on its own (#1634) for security reasons, the latter option (2) is the one to implement, as the former (1) won't work when we get to implement that ticket.

Changes: 
* [ ] Install Pure-ftpd in `ops/packaging/Dockerfile` and `ops/packaging/Production-Dockerfile` so CLI tools to manage config are accessible from GigaDB
* [ ] The main file to change is `fuw/app/backend/models/FiledropAccount.php`
* [ ] Add in docker-compose.yml and docker-compose.production-envs.yml the following so the config are shared: 
```
      - ${DATA_SAVE_PATH}/pure-ftpd:/etc/pure-ftpd
      - ${DATA_SAVE_PATH}/pure-ftpd/passwd:/etc/pure-ftpd/passwd
 ```
* [ ] Remove docker-php from composer.json.dist
* [ ] Delete `fuw/app/backend/models/DockerManager.php` 
* [ ] Update functional tests in `fuw/app/backend/tests/unit/FiledropAccountTest.php`

Issue #1659, opened 2024-01-25
FUW local deployment on Ubuntu


luistoptal:
## User story

As a dev I want to deploy the project locally in Ubuntu and be able to follow the FUW flow without encountering errors, as described here: https://github.com/gigascience/gigadb-website/pull/1592

(please see the last issue reported at the bottom, `Update (1-feb)` heading)

I describe the steps I follow to encounter the errors:

- I checkout to the `develop` branch
- `cp ops/configuration/variables/env-sample .env`
- `cp ops/configuration/variables/secrets-sample .secrets`
- Add `FFLAG_API_URL` and `FFLAG_ID` to `.secrets`
- I run `export NODE_OPTIONS=--openssl-legacy-provider` in a terminal
- I confirm that a `chromedriver` is present, e.g. `ls /usr/local/bin/chromedriver`
- I run `docker-compose down -v --remove-orphans`
- I run `sudo ./up.sh` (I need `sudo` because `docker-compose` needs root permissions to run)
- I run `docker-compose run --rm test ./vendor/codeception/codeception/codecept run --debug unit DatasetTest` and get the following output
```sh
Codeception PHP Testing Framework v3.1.3
Powered by PHPUnit 5.7.27 by Sebastian Bergmann and contributors.
Running with seed: 


Unit Tests (4) ------------------------------------------------------------------------------------------------------------------------------
Modules: Asserts, \Helper\Unit
---------------------------------------------------------------------------------------------------------------------------------------------
✔ DatasetTest: Get authors (0.29s)
✔ DatasetTest: Get author names (0.07s)
✔ DatasetTest: Get curator name (0.07s)
- DatasetTest: Get available status list
  *** FUW is enabled ***
✔ DatasetTest: Get available status list (1.35s)
---------------------------------------------------------------------------------------------------------------------------------------------


Time: 1.86 seconds, Memory: 18.00MB

OK (4 tests, 7 assertions)
```
- To test the flow manually I run the up script again (I assume the automated tests change the local DB, because unless I run the script again, I no longer see the `200070` id)
- I access the dataset admin screen as admin (admin@gigadb.org) and see:
![image](https://github.com/gigascience/gigadb-website/assets/143437854/7644f238-70bd-40d6-b341-916217fca3b0)
- In http://gigadb.gigasciencejournal.com/adminDataset/update/id/668 I change Submitter to `user@gigadb.org` and Upload Status to `AssigningFTPBox` and save
[...]

PR #1592, opened 2023-12-18
Initial implementation of Fuw re-enabled

rija:
# Pull request for issue: #1603 

This is a pull request for the following functionalities:

* Re-enable the current state of File Upload Wizard workflow locally
* Add various fixes to the FUW workflow to enable it to function end-to-end as per the video tutorial from two years ago
* Add various fixes to issues that impede on the development
* Implement an easy to operate feature flag for File Upload Wizard using Gitlab Feature flag, currently set up to enable FUW on dev environment only

## How to test?

There are two changes to test: 
1. The feature flag
2. That the instructions shown in the video can still be executed

### Prerequisite

* Update the variable "HOME_URL" for the "dev" environment in the Gitlab variables for your fork of Gigadb website, so it doesn't have a `:9170` in it
* ensure you regenerate your `.env` from `env-sample`
* remove `.secrets`
* Set and export  environment variable `NODE_OPTIONS` with value `--openssl-legacy-provider` in your terminal
* Ensure you have [a recent chrome driver binary](https://googlechromelabs.github.io/chrome-for-testing/) installed at `/usr/local/bin/chromedriver` (you can copy the binary from the downloaded package `chrome-mac-arm64.zip` into `/usr/local/bin`)
* Start the local website with `./up.sh`

#### Automatically

After starting the local website with `up.sh`, run the test:
```
docker-compose run --rm test ./vendor/codeception/codeception/codecept run --debug unit DatasetTest
```

Tests should be passing and a debug output should say: "FUW is enabled "

Navigate to the Feature flag dashboard on Gitlab: https://gitlab.com/gigascience/upstream/gigadb-website/-/feature_flags, and turn the switch off.

**Wait a few seconds (~30s)** for the API to update its state, and run the same test again.
The tests should still be passing but the debug output should say: "FUW is NOT enabled".

Then turn on the switch in the dashboard again.

#### Manually

Navigate as admin to: http://gigadb.gigasciencejournal.com/adminDataset/update/id/668
Click on the upload status drop-down, you should see them all, including the ones used by File Upload Wizard:
 * AssigningFTPbox
 * UserUploadingData
 * DataAvailableForReview
 * DataPending

Navigate to the Feature flag dashboard on Gitlab: https://gitlab.com/gigascience/upstream/gigadb-website/-/feature_flags, and turn the switch off.

**Wait a few seconds (~30s to 1mn)** for the API to update its state, and refresh the dataset form.
The upload status related to the FUW should not be appearing in the drop-down list.

[...]

PR #1656, opened 2024-01-23
Fix acceptance tests for production after merging accessibility feature

kencho51:
# Pull request for issue: #1657 
This is a pull request for the following functionalities:

This PR tries to fix the following broken acceptance tests inherited from the PR #1642:
```
# given the GITLAB_PRIVATE_TOKEN is available and the if block in acceptance_runner is uncommented
% ./up.sh
% ./tests/acceptance_runner
...
There were 6 failures:
1) form to update dataset details: Delete an image's file and then remove the image record
 Test  tests/acceptance/AdminDatasetForm.feature:Delete an image's file and then remove the image record
 Step  I should not see "url" 
 Fail  Failed asserting that on page /adminDataset/update/id/5
--> Skip to main content
Joe's GigaDB Page Admin LogOut Browse Samples
Search
Home
About 
Help 
Terms of use
Admin Manage Update
Update Dataset 100039
Fields with * are required.
Submitter *
admin1@gigasciencejournal.com
admin@gigadb.org
test+14@gigasciencejournal.com
test+163@gigasciencejournal.com
tes
[Content too long to display. See complete response in '/project/tests/_output/' directory]
--> does not contain "URL".

Scenario Steps:

 28. $I->dontSee("URL") at tests/_support/AcceptanceTester.php:63
 27. $I->wait("1") at tests/_support/AcceptanceTester.php:228
 26. $I->acceptPopup() at tests/_support/AcceptanceTester.php:346
 25. $I->seeInPopup("Are you sure? This will take effect immediately") at tests/_support/AcceptanceTester.php:345
 24. $I->click("Remove image") at tests/_support/AcceptanceTester.php:47
 23. $I->wait("2") at tests/_support/AcceptanceTester.php:228

Artifacts:

Png: /project/tests/_output/AdminDatasetForm.Delete.an.image.s.file.and.then.remove.the.image.record.fail.png
Html: /project/tests/_output/AdminDatasetForm.Delete.an.image.s.file.and.then.remove.the.image.record.fail.html
2) form to update dataset details: Click view curation record image with link
 Test  tests/acceptance/AdminDatasetForm.feature:Click view curation record image with link
 Step  I should see an image with alternate text "view" is linked to "http://gigadb.test/curation log/view/id/3" 
 Fail  Failed asserting that an array is not empty.

Scenario Steps:

 14. $I->seeElement("//img[@alt='View']") at tests/_support/AcceptanceTester.php:264
[...]

PR #1688, opened 2024-02-13
Remove dependency on abandoned inter-container-communication project and fix broken tests

rija:
# Pull request for issue: #1659

1. Temporary hack allowing developers on Linux (like @luistoptal) to run the FUW workflow
2. Fix a broken functional test from [[#1592](https://github.com/gigascience/gigadb-website/pull/1592)](https://github.com/gigascience/gigadb-website/pull/1592)
3. Fix Gitlab pipeline that fails on Forks and Upstream for the files-metadata-console tests

## How to test?

Run all the test locally and on CI. They should all pass.

1. I commented out the body of functions defined in `FiledropAccount.php` and used to interact with the Pure-FTP container. I make sure those functions return true so most of the FUW workflow can still be performed. Only the ftp interface won't be available. That allows developer on Linux to work on FUW and prevent abandoned dependency `docker-php/docker-php` to be used.
2. The failure of `tests/functional/FeatureFlagCest.php` is an issue introduced in #1592 and it's caused by me forgetting that GigaDB environment on CI is not the same as on dev. So the fix is to add an if-else block based on which environment we are in
3. Recent build of the container image `composer/composer:2.4.4` pushed to Docker Hub by the composer team had their PHP upgraded to `8.3.2` which caused the composer configuration of the `files-metadata-console` project to fail in its `test` stage during setup and `composer install` on CI because our set of composer dependencies cannot work with that version of PHP. The immediate fix is to use `composer/composer:2.4.3`that runs PHP `8.1.11`. The longer and proper fix is to migrate to recent version of PHP which is scheduled in PR #1702

Yes, it will disable ftp serving for FUW drop-box account.
But it's ok, the current implementation relied on an abandoned open source project (docker-php/docker-php) which was risky, and the correct fix is scheduled  in ticket  #1680 

Furthermore, there are two acceptance tests scenario in `tests/acceptance/AdminDatasetSample.feature` that are now commented out. Their failure is somewhat related to the accessibility and admin style upgrade work, but it fails differently on CI and locally with no obvious cause and require further investigation. But because the functionality they are testing appears to work after manual testing and the fact that the other changes on this PR are blocking FUW development or deployment of GigaDB, it's fine for those scenarios to be temporarily disabled (I have added a TODO so we don't forget about them)

Yes, fixed the feature flag functional tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix #1659: Remove dependency on abandoned inter-container-communication project and fix broken tests
 - Feat #1624: FUW - Migrate to Vue 2.7
 - Feat #1334: Improve accessibility of public and admin pages, revamp admin pages styles, sync style guide with updated styles
 
--- a/fuw/app/backend/models/FiledropAccount.php
+++ b/fuw/app/backend/models/FiledropAccount.php
@@ -259,29 +259,29 @@ class FiledropAccount extends \yii\db\ActiveRecord
     {
         $status = true ;
 
-        $uploaderExists = shell_exec("grep uploader-$doi /etc/pure-ftpd/passwd/pureftpd.passwd");
-        $downloaderExists = shell_exec("grep downloader-$doi /etc/pure-ftpd/passwd/pureftpd.passwd");
-        if ($uploaderExists || $downloaderExists)
-            Yii::warning("!! FTP account already exists");
-        $dryRunModeArray = ["bash","-c","pwd"] ;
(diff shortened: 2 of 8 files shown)


kencho51 (approved):
Hi @rija,

**praise:** All tests are passing locally and also in gitlab pipeline.
**question(non-blocking):** Right the the fuw is turned on in dev only, but what if when it becomes available in production, the featureFlagCest.php will need to update again? 
**note(non-blocking):** The failing scenarios in AdminDatasetSample.feature will be looked into in PR #1656 

LGTM, happy to approve

pli888 (approved):
**praise**: All tests are passing now.
**praise**: Code changes look good to me.
**praise**: CI/CD pipeline is running green now (except for conformance and security stage).
**note (non-blocking)**: There seems to be a problem with container scanning job in CI pipeline.

Happy to approve this PR.

kencho51:
> **note (non-blocking)**: There seems to be a problem with container scanning job in CI pipeline.

This will be fixed in PR #1656 too.

Issue #1639, opened 2024-01-04
Add datasets containing over 2000 files or samples into warm cache job


pli888:
* For example, dataset 100894 contains over 2000 samples
* Take into account the number of datasets that will go into cache as it is memory-based and we might run out of memory on webapp server
* Look into other forms of caching (redis)

rija:
implemented in PR #1745

PR #1745, opened 2024-03-07
Paginate file sample queries on dataset pages

rija:
# Pull request for issue: #1639

* Update the code fetching files for a dataset to be displayed on dataset page to only query the number to be displayed
* Update the code fetching samples for a dataset to be displayed on dataset page to only query the number to be displayed
* Remove the cache warming job in the Gitlab pipeline configuration as no longer needed in that shape
* Fix a couple of failures when running tests in files-metadata-console project

### Prepare your staging environment

1. Checkout my branch and change directory  to it.
2. Generate your staging environment:

```
$ cd ops/infrastructure/envs/staging/
$ ../../../scripts/tf_init.sh --project \<your gitlab project path\> --env staging --region \<your region\> --ssh-key \<path to your key\> --web-ec2-type t3.small --bastion-ec2-type t3.small
$ terraform plan
$ terraform apply
$ terraform refresh
$ env TF_KEY_NAME=private_ip OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES ansible-playbook  -i ../../inventories webapp_playbook.yml -e="gigadb_env=staging"
$ env OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES ansible-playbook -i ../../inventories bastion_playbook.yml -e "gigadb_env=staging" -e "backupDate=latest"
```
   

### Test the functionality on your staging environment

Then navigate to the page of the following 10 datasets on your staging deployment. They should all load without errors:

```
Public datasets with top number of files

DOI		nb_files
100396	197214
100310	19981
100707	16675
100608	12864
100512	12238


Public datasets with top number of samples

DOI		nb_samples
100314	65535
100824	6467
100855	4959
100332	4033
100894	2906

```

The work involve three main changes: 

1. Update the `protected/interfaces/DatasetFilesInterface.php` and the `protected/interfaces/DatasetSamplesInterface.php` interfaces to ensure the data fetching method signature have limit and offset so we can pass the PostgreSQL limit (page size) and offset (starting page) parameters across all layers 

    1a. Special consideration needed to be taken in `protected/components/CachedDatasetFiles.php` and `protected/components/CachedDatasetSamples.php` to change the key used in the cache for files and samples so it includes the page size and starting page otherwise some visitors will get the wrong data displayed in the table fetched from cache. For queries that are not to be paginated the key will include "ALL_0" with means get everything. Those changes required updating a bunch of unit tests. 

    1b. In `protected/components/FormattedDatasetFiless.php` and `protected/components/FormattedDatasetSamples.php`, Untangled the `CPagination` pager class (and make it an injected dependency) from the `CArrayDataProvider` object so that we can obtain the Page size and current page from the view and pass them as limit and offset to the data fetching methods

	1c. Finally in `protected/components/StoredDatasetFiles.php` and `protected/components/StoredDatasetSamples.php`, ensure the SQL query use the limit and offset values that was passed from the upper layer of the onion

2. Update our custom Dependency Injection container `protected/components/DatasetPageAssembly.php`  to inject a `CPagination` pager object  into the new instance of `protected/components/FormattedDatasetFiless.php` and `protected/components/FormattedDatasetSamples.php`

3. Add new test scenarios in `tests/unit/FormattedDatasetFilesTest.php` and `tests/unit/FormattedDatasetSamplesTest.php`to ensure that the above code will make accurate paginated retrievals of data to the layer below (that's the other reason we made the pager an injected dependency, so can create a mock one we can control in test)

Visually, nothing is changing to the users, so the acceptance test scenarios that test the pagination of the files and samples didn't needed changing.
However in order to manually test the work on the dev database, I had to add to `dataset/dev` a new dataset (100035) with enough samples to activate pagination.
That required making minor changes to two legacy acceptance tests that were relying on the number of dataset or making a bet on what the next sample id was going to be (`tests/acceptance/001-Search.feature` and `tests/acceptance/AdminSample.feature`)

1. The cache warming in its current form is useless, so I've removed it, however cache warming as a technique is 
very useful and very beneficial to good end user experience if it's done smartly. 
I feel this out of the scope of this ticket (and it requires the web traffic analytics to be set up and working anyways), so I've created a new user story for it: #1756

2. I've noticed the pager text input below the sample table is not showing, but this issue already occurs on develop branch, so I created a bug issue for it (#1757). It's not going to be done in this PR

Yes, see above sections.

Additionally I've fixed a failing test (affecting develop pipeline too) in `gigadb/app/tools/files-metadata-console/tests/functional/CheckValidURLsCest.php`, because either curl changed or the example.com site changed so the error message returned by curl was no longer that of a page not found error. The fix is to replace the URL with one that will return a 404.

1. A method to DatasetComponent was incorrectly named (`saveLocaldataInCache`), so I've updated the naming (`saveLocalDataInCache`) which required changes to many components.

2. When building or running tests `files-metadata-console`, there was an error message thrown by `symfony/console`: `"Fatal error: Uncaught Error: Attempting to use non-attribute class "Symfony\Component\Console\Attribute\AsCommand" as attribute"`.

This was fixed by running `docker-compose run composer update symfony/console`, which is why `gigadb/app/tools/files-metadata-console/composer.lock` has been updated.


kencho51 (commented):
HI @rija,

**praise:** All the public dataset pages with top number of files/samples can be displayed much faster in my staging (<10 seconds) compared to production staging (>3mins) for the `/dataset/100512`.
**praise:** All tests are passing both in dev and the gitlab pipeline.
**praise:** After deploying this PR to staging and go to `/dataset/100512` Files tab, it would show `Displaying 10 files of 12238`, but just `Displaying 10 files of 8000` in the production staging. After checking the database using `select count(id)  from file where dataset_id = '679';`, the number `12238` is the correct number, which means this PR helps to fix this error.
**praise:** Fixed the error in the gitlab `files-metadata-console` test which would fail randomly sometimes.
**prasie:** The gitlab `cache warming` stage which would fail because of `take too long to complete` is removed.
**praise** The CPagination $pager will pass in limit value and offset value into the functions  `getDatasetFiles()` and `getDatasetSamples()`? The value for $limit is All or from the `$getCurrentPage()`, and for $offset is 0 or from `getCurrentPage * getPageSize()`, it is elegant that the pagination helps to optimize no. of items to be displayed in the dataset page which does not need to get all of it from database. 
**suggestion:** If we go to `/dataset/100894` which contain large number of samples, but there is no `Display x samples of y` like the one in the Files tab,  would it be better to have one?
**question:** From [here](https://www.postgresql.org/docs/14/queries-limit.html), is it better to add in `ORDER BY` in the sql cmd in functions `getDatasetFiles()` and `getDatasetSamples()` to achieve more consistent and predictable query result?

rija:
Hi @kencho51

>**suggestion:** If we go to `/dataset/100894` which contain large number of samples, but there is no `Display x samples of y` like the one in the Files tab,  would it be better to have one?

Yes indeed but it's better to do that as part of the new ticket #1757 alongside the also missing sample pager text input for jumping to specific pages. I have updated that ticket to add the suggested text.


>**question:** From [here](https://www.postgresql.org/docs/14/queries-limit.html), is it better to add in `ORDER BY` in the sql cmd in functions `getDatasetFiles()` and `getDatasetSamples()` to achieve more consistent and predictable query result?

Good spot, and I have made the suggested change.

rija:
Hi @kencho51

>**suggestion:** If we go to `/dataset/100894` which contain large number of samples, but there is no `Display x samples of y` like the one in the Files tab,  would it be better to have one?

Yes indeed but it's better to do that as part of the new ticket  I've referenced in the "Any issues with implementation?" section (#1757) alongside the also missing sample pager text input for jumping to specific pages. I have updated that ticket to add the suggested text.


>**question:** From [here](https://www.postgresql.org/docs/14/queries-limit.html), is it better to add in `ORDER BY` in the sql cmd in functions `getDatasetFiles()` and `getDatasetSamples()` to achieve more consistent and predictable query result?

Good spot, and I have made the suggested change.

kencho51 (approved):
LGTM after the updates, happy to approve.

Issue #1713, opened 2024-02-28
erroneous originator for curation log entry


only1chunts:
**Describe the bug**

**To Reproduce**
1. Go to https://staging.gigadb.org/adminDataset/update/id/2747
2. change something, Click save
3. Scroll down to the curation log
4. See error: the log has a new entry saying created by "system" 

**Expected behavior**
If I am logged in as a curator/admin and I make a change to the dataset I expect the curation log to state that it was me that made an update not "system".

**Additional context**
related to #1615

kencho51:
Close as fixed in PR #1753

PR #1753, opened 2024-03-15
Make curation log recognize user in the created by column

kencho51:
# Pull request for issue: #1713

All the curation log action will be recorded in the curation log table with `created_by` as the logged in user's full name.

## How to test?

Describe how the new functionalities can be tested by PR reviewers

### In dev

```
% ./up.sh
% ./tests/unit_functional_runner
% ./tests/acceptance_runner  
```

1. Then go to http://gigadb.gigasciencejournal.com/ and log in as admin
2. Follow the following steps to test:
```
  Scenario: Display full name after creating a new curation log entry
    Given I am on "/adminDataset/update/id/5"
    When I press the button "Create New Log"
    And I am on "/curationLog/create/id/5"
    And I should see "Create Curation Log"
    And I select "Comment" from the field "CurationLog_action"
    And I fill in the field of "name" "CurationLog[comments]" with "hello world"
    And I press the button "Create"
    And I press the button "Back to this Dataset Curation Log"
    Then I should see "Joe Bloggs"
    And I should see "hello world"

  @ok @curationlog
  Scenario: Display full name after updating the upload status
    Given I am on "/adminDataset/update/id/5"
    When I select "Curation" from the field "Dataset_upload_status"
    And I press the button "Save"
    Then I should see "$YourLogInName"
    And I should see "Status changed to Curation"
```

### In staging

1. After you staing gigadb website has been spun up and running, log in as admin
2. Follow the following steps to test:
```
  Scenario: Display full name after creating a new curation log entry
    Given I am on "/adminDataset/update/id/405"
    When I press the button "Create New Log"
    And I am on "/curationLog/create/id/5"
    And I should see "Create Curation Log"
    And I select "Comment" from the field "CurationLog_action"
    And I fill in the field of "name" "CurationLog[comments]" with "hello world"
    And I press the button "Create"
    And I press the button "Back to this Dataset Curation Log"
    Then I should see "$YourLogInName"
    And I should see "hello world"

  @ok @curationlog
  Scenario: Display full name after updating the upload status
    Given I am on "/adminDataset/update/id/405"
    When I select "Curation" from the field "Dataset_upload_status"
    And I press the button "Save"
    Then I should see "$YourLogInName"
    And I should "Status changed to Curation"
```

Describe how the new functionalities have been implemented by the
changed code at a high level

1. In curation log model class, replace the hard coded user `System` using the full name of the logged in user.
4. Create `getLastName()` method in the `WebUser` components class, so the full name can be obtained by concating `Yii::app()->user->getFirstName()` and `Yii::app()->user->getLastName()`.
5. Update the unit test and acceptance test for checking the functionality of the curation log.

None.

None.

None.

Make the TODO tests in `AdminFileAttributes.feature` passing

None.

## Summary by CodeRabbit

- **Bug Fixes**
  - The curation log now correctly displays the full name of the user responsible for changes instead of a generic system label.

- **New Features**
  - User full names are now recognized and shown in curation log entries.

- **Tests**
  - Acceptance and unit tests updated and expanded to verify the correct display of user full names in curation logs.
  - Test stability improved with additional wait steps and updated tagging conventions.
  - Previously disabled scenarios re-enabled for more comprehensive coverage.

- **Documentation**
  - Changelog updated to reflect the fix for user name recognition in curation logs.


protected/components/WebUser.php
```diff
@@ -27,13 +27,19 @@ public function checkAccess($operation, $params = array(), $allowCaching = true)
...
+    public function getFirstName()
     {
         $user = $this->loadUser(Yii::app()->user->id);
         return $user->first_name;
     }
 
-    function getEmail()
+    public function getLastName()
+    {
+        $user = $this->loadUser(Yii::app()->user->id);
```
alli83: since you are extending CWebUser, you can access the current user directly via `$this` so there is no need to call `Yii::app()->user->id`, just use `$this->id`

protected/models/CurationLog.php
```diff
@@ -109,7 +109,7 @@ public static function makeNewInstanceForDatasetBy(int $id, string $creator): Cu
      */
     public static function createCurationLogEntry(int $id, string $fileName): bool
     {
-        $curationlog = self::makeNewInstanceForCurationLogBy($id, "System");
+        $curationlog = self::makeNewInstanceForCurationLogBy($id, Yii::app()->user->getFirstName() . " " . Yii::app()->user->getLastName());
```
alli83: just take into account that firstName might be null or empty so you could end up with extra spaces when building the full name

kencho51:
@coderabbitai full review

protected/models/CurationLog.php
kencho51: In the `User.php` model class, the firstName and the lastName are the required fields, so the `empty/null` cases would not happen. But, extra spaces around the name could be the case, so I have created function `getCurrentUserFullName(): string` to deal with this case and updated the unit tests accordingly.

kencho51:
@coderabbitai review

rija (changes requested):
Hi @kencho51,

**note**: I think this PR needs rebasing (pipeline fails on debian buster error) and would benefit with some instructions on how to test (other than running the tests)

kencho51:
Hi @rija,

> **note**: I think this PR needs rebasing (pipeline fails on debian buster error) and would benefit with some instructions on how to test (other than running the tests)

Rebase is done. And the how to test has been updated.

rija (approved):
Hi @kencho51,

**praise**: whereas before this PR changing upload status resulted in a curation log entry associated with "system", with this PR, the curation log entry resulting form upload status change is associated with the currently logged-in user

**praise**: with this PR, a manually added curation log comment entry is associated with the currently logged-in user like before 

**praise**: improved test coverage

**nitpick** (non-blocking): the functions signature of the change code are missing the type hints for the return type

happy to approve

protected/models/CurationLog.php
```diff
@@ -100,6 +100,32 @@ public static function makeNewInstanceForDatasetBy(int $id, string $creator): Cu
...
     }
 
+    /**
+     * Builds a full name from first and last names, trimming whitespace and filtering out empty values.
+     *
+     * @param string|null $firstName The first name.
+     * @param string|null $lastName The last name.
+     * @return string The full name, or an empty string if both names are empty.
+     */
+    public static function buildFullName(?string $firstName, ?string $lastName): string
```
alli83: ?string $firstName =null

if it's nullable you should add = null

alli83:
@rija @kencho51 both firstName and lastName are nullable  in DB (is that intentional?). if so, the full name could be empty, so we might need to handle it accordingly, right?

Update: wrong db, firstName and lastName are required. So All good!

kencho51:
@alli83 

> both firstName and lastName are nullable in DB (is that intentional?). if so, the full name could be empty, so we might need to handle it accordingly, right?

I thought the `buildFullName()` can handle the nullable case sufficiently, because:
1. trim(null) returns an empty string ""
2. array_filter() removes empty strings from the array
3. implode() on an empty array returns an empty string

But by adding the `$firstName =null` does make the function more robust and follow the best practice. 

Thanks for the suggestion, please see https://github.com/gigascience/gigadb-website/pull/1753/commits/7d49552e9dfd2345c258fb0ebbe0c989461dd782

alli83:
@kencho51 what I'm wondering is, if firstName and lastName are both empty strings, the the curation log will just show an empty value like "". Shouldn't we display something to indicate that the curator's name is missing or empty?

kencho51:
@alli83,

> if firstName and lastName are both empty strings,

These values actually are getting from the logged in users, so the above case you mentioned will barely happened, because user will not be created during user registration stage.

alli83:
wrong db, firstName and lastName are required. So All good!

pli888 (approved):
**praise**: Functionality works as described on my `dev` and `staging` environments.
**praise**: All tests are passing on my CI pipeline.
**praise**: Code looks good to me.
**issue (non-blocking)**: Executing the tests in my `dev` environment gives me an error: `[Codeception\Exception\ConfigurationException] Params can't be loaded from .env.test`.
**suggestion**: This problem of mine is unrelated to the work in this PR - I will resolve it separately.

This PR looks good to me - happy to approve.

Issue #1800, opened 2024-04-18
Sample tables always show first page


rija:
From https://github.com/gigascience/gigadb-website/pull/1789#issuecomment-2056821321:

>> @rija Can someone that understands better the backend and the controllers take a look at this, I feel like there's potentially a bug in the way the sample paginations results are returned from the controller (I'm not sure)
>
>Hi @luistoptal,
>
>Ok, I'll have a look within a few days.
>
>At a glance, it seems like the offset and limit  are neither calculated correctly nor passed from the frontend to the backend.
>
>The range of data to return from the database (via the cache) is determined by the `offset` and the `limit`.
The former is where the range starts and the latter is how many rows from the offset to returns.
If either parameter is not supplied the default values are used which in the case of `offset` is `0`, which means from the beginning. The default value for `limit` in the database is `ALL` which means all the matching row in the database.
But the frontend has default value for page size which is then passed on to the backend and used as value for limit.
>
>What you describe seems to indicate that effective offset used is always the default value of 0
and the limit, which should not change from page to page, is not picking up on your settings changes. 
Wonder if that could be a caching problem.

rija:
Implemented in #1811

Issue #1783, opened 2024-04-10
issues moving between sample and file view after clicking through some pages of files


only1chunts:
**Describe the bug**
Looking at a dataset with lots of files and lots of samples, e.g. https://staging.gigadb.org/dataset/100894
When I navigate through the sample table, e.g. click page 9 of samples
Then try to navigate around the files table, e.g. click page4 of files
Then try to return to the samples table view and navigate around that, it takes me back to the files table.
There seems to be new parts appended to the URL each time I click a page within a table, e.g.
start on: https://staging.gigadb.org/dataset/100894
click page 9 of samples: https://staging.gigadb.org/dataset/view/id/100894/Samples_page/9
navigate to files table, click page 4 of files: https://staging.gigadb.org/dataset/view/id/100894/Samples_page/9/Files_page/4

It looks like its retaining the previous `Samples_page/9` part of the URL when you navigate around the files table.

**Expected behavior**
I expect to be shown the page/view that I click on

**Additional context**

rija:
Thanks for reporting. 
Two comments on this one:

1. It already behaves that way on CNGB's deployment at http://gigadb.org (see screenshot)
2. It is causing search bots (Google, Bing, ...)  to follow all those urls combinations


<img width="1388" alt="Screenshot 2024-04-10 at 4 08 30 PM" src="https://github.com/gigascience/gigadb-website/assets/11455/121323b4-44bf-474c-90b0-4fe3834fa74d">

rija:
Note: Reported as part of testing GigaDB release [4.2.3](https://github.com/gigascience/gigadb-website/releases/tag/v4.2.3)

only1chunts:
> It already behaves that way on CNGB's deployment at http://gigadb.org/ 

Wow- I never noticed that before! I think the reason is that most datasets dont have both a lot of samples AND a lot of files, so no need to look at multiple pages of each. Also I have my default settings on the tables set to display 100 rows at a time, so very few datasets have multi-page tables for me.

Would adjusting the default number of rows to be 100 make things easier?

luistoptal:
Regarding the issue with the URLs, I have created a PR to fix it: https://github.com/gigascience/gigadb-website/pull/1790

this should also fix the issue reported above, ie 

> causing search bots (Google, Bing, ...) to follow all those urls combinations

So I believe it should be ok to allow indexing by crawlers

[...]

PR #1811, opened 2024-04-19
Fix: Samples table stuck to first page results when paginating

rija:
# Pull request for issue: #1800 

* Ensure sample tables show the correct page of sample result on all pages

## How to test?

Navigate to http://gigadb.gigasciencejournal.com/dataset/100035/

and check that you can paginate through all samples.

The bug was that when loading the current page of result from cache we forgot to pass the limit and offset that consist part of the cache key, so it was retrieving the wrong data from cache

N/a

Updated the unit test `tests/unit/CachedDatasetSamplesTest.php` with a new scenario that reveals the bug by testing various value of limit and offset . 

N/a

N/a

N/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-- Feat #1741L: Add carets to sortable headers in dataset page
 
 ## Unreleased
 
 - Fix #1800: Read the correct sample page from cache on samples tables
+- Feat #1163: Add carets to sortable headers in dataset page
 - Feat #1664: Enable post upload script to work with non-published datasets
 
 ## v4.2.4 - 2024-04-16 - 576757f38
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


kencho51 (approved):
Hi @rija,

**praise:** With the refactored  `getDatasetSamples()`, the pagination in the sample tab is working.
**praise:** The updated unit test is passing
**suggestion(non-blocking):** Since this PR would bring visual changes at the user sides, so would it be a good idea to include the following scenario in the `DatasetView.feature`:

```
 Scenario: Pagination in the sample tab
    Given I have not signed in
    And I am on "/dataset/100035"
    And I follow "Sample"
    And I should see "SRS173539"
    When I follow "2"
    Then I should see "SRS173549"
    And I should not see "SRS173539"
``` 
**

pli888 (approved):
**praise**: Clicking sample pagination links now returns expected results in sample table.
**praise**: New unit test in CachedDatasetSamplesTest.php is passing on CI pipeline.
**thought (non-blocking)**: Like Ken, I wonder if a new acceptance test in DatasetView.feature is useful which would also check if the destination URL is correct, e.g. /dataset/view/id/100035/Samples_page/2, as well as the content of the sample table when clicking on sample pagination links.
**suggestion**: Add a TODO comment for above acceptance test to be implemented by #1783.

rija:
Hi all

thanks from reviewing:

@kencho51:
That suggestion makes sense and I've implemented it and the test passes locally and on CI

@pli888:
Good thinking but that comment is probably more suitable on PR #1790

Issue #1959, opened 2024-07-02
The "Dataset:Relation" function is not available in the admin page


YFan0019:
**Describe the bug**
The "Dataset:Relation" function is not available in the admin page

**To Reproduce**
1. Go to https://gigadb.org/adminRelation/admin
2. Click on "Create a new relation"
3. Try to enter the doi and relation, click create.
4. See error image

![GigaDB-Create-AdminRelation](https://github.com/gigascience/gigadb-website/assets/112144621/489957d9-7abc-4422-bcaf-0f96f2926478)

**Expected behavior**
The new relation should be saved in the system.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:  iOS
 - Browser chrome
 - Version 126.0.6478.127

**Smartphone (please complete the following information):**

**Additional context**

YFan0019:
Please add the relationship to the datasets from the backend server.
https://github.com/gigascience/dataset-tracking/issues/8

102538 HasPart 102546, 102545, 102544, 102543, 102542
https://gigadb.org/dataset/102538/token/h5r1bskBP6UORwni

102542 isPartOf 102538
https://gigadb.org/dataset/102542/token/79Vgzra-97MVu9s4

102543 isPartOf 102538
https://gigadb.org/dataset/102543/token/s7LBQcAmPo_cImVg

102544 isPartOf 102538
https://gigadb.org/dataset/102544/token/vEKNF93Qi945PzNo

102545 isPartOf 102538
https://gigadb.org/dataset/102545/token/hj-FXi-Ni7urZ7_A

102546 isPartOf 102538
https://gigadb.org/dataset/102546/token/A2QvodUl-z1TdXO_

rija:
HI @YFan0019 

I have run the SQL query to create the relationships on Staging (https://staging.gigadb.org/). 
Can you check that it works for you. 
If yes, I will then run the query on the live database.

YFan0019:
Hi @rija 

Looks good for me.
Thank you so much :)

rija:
Hi @YFan0019,

I have now run the query on the live environment.
Let me know if it looks Ok.

For info the query is:
```
insert into relation (dataset_id, related_doi,relationship_id) values (2805,102546,12),(2805,102545,12),(2805,102544,12),(2805,102543,12),(2805,102542,12);

insert into relation (dataset_id, related_doi,relationship_id) values(2809,102538,11);
[...]

PR #1987, opened 2024-07-25
fix 1959: dataset relation wrong attrribute name relationship

alli83:
# Pull request for issue: #1959 

- create relation without error: fix
- add tests
- add checkbox in order to decide if the reciprocal relation should be also created (yes by default)

## How to test?

Go to https://gigadb.org/adminRelation/admin
Click on "Create a new relation"
Try to enter the doi and relation, click create.

- fix attribute name
- fix returnType
- update all relationships
- add checkbox
- handle errors

unit test fixed / functional tests added

### DIFF
--- a/tests/unit/RelationDAOTest.php
+++ b/tests/unit/RelationDAOTest.php
@@ -1,58 +1,66 @@
 <?php
 
-class RelationDAOTest extends CDbTestCase
+/**
+ * run with:
+ * docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit RelationDAOTest
+ */
+class RelationDAOTest extends \CDbTestCase
 {
     /**
      * @dataProvider relationshipProvider
      */
-    public function testItshouldAddReciprocalRelation($relationship, $expected_reciprocal)
+    public function testItShouldAddReciprocalRelation($relationship, $expected_reciprocal)
     {
         $relation_stub = $this->createMock(Relation::class);
+        $relationshipMock = $this->createMock(Relationship::class);
+        $relationshipMock->method('getName')
+            ->willReturn($relationship);
 
         // Configure the stub for the relation for which to create a reciprocal relation
         $relation_stub->method('getRelatedDOI')
-             ->willReturn('100249'); //of id 2
+            ->willReturn('100249'); //of id 2
         $relation_stub->method('getDatasetID')
-             ->willReturn('1'); // of identifier 100243
+            ->willReturn('1'); // of identifier 100243
         $relation_stub->method('getRelationship')
-             ->willReturn($relationship);
+            ->willReturn($relationshipMock);
 
         // Create a mock for the Relation class,
         // only mock the setDatasetID, setRelatedDOI, setRelationship methods.
         $reciprocal_relation = $this->getMockBuilder(Relation::class)
-                         ->setMethods(['setDatasetID', 'setRelatedDOI', 'setRelationship', 'save'])
-                         ->getMock();
+            ->setMethods(['setDatasetID', 'setRelatedDOI', 'setRelationship', 'save'])
+            ->getMock();
 
         // Set up the expectation for the setDatasetID method
         // to be called only once and with:
         // the id of the related dataset (of DOI 100249)
         // as its parameter.
         $reciprocal_relation->expects($this->once())
-                 ->method('setDatasetID')
-                 ->with($this->equalTo(2));
+            ->method('setDatasetID')
+            ->with($this->equalTo(2));
 
         // Set up the expectation for the setRelatedDOI method
         // to be called only once and with:
         // the DOI of the relating dataset (of id 1)
         // as its parameter.
         $reciprocal_relation->expects($this->once())
-                 ->method('setRelatedDOI')
-                 ->with($this->equalTo('100243'));
+            ->method('setRelatedDOI')
+            ->with($this->equalTo('100243'));
 
         // Set up the expectation for the setRelationship method
         // to be called only once and with:
         // the reciprocal relationship supplied by dataProvider
         // as its parameter.
         $reciprocal_relation->expects($this->once())
-                 ->method('setRelationship')
-                 ->with($this->equalTo($expected_reciprocal));
+            ->method('setRelationship')
+            ->with($this->equalTo($expected_reciprocal));
 
         // Set up the expectation for the save method
         // to be called only once
         $reciprocal_relation->expects($this->once())
-                 ->method('save');
+            ->method('save')
+            ->willReturn(true);
 
-        $system_under_test = new RelationDAO() ;
+        $system_under_test = new RelationDAO();
         // createReciprocalTo make use of dependency injection for the reciprocal relation instantiation
         $system_under_test->createReciprocalTo($relation_stub, $reciprocal_relation);
     }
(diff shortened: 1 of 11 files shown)


kencho51 (commented):
Hi @alli83,

**praise:** Good to have a checkbox implemented that allows user to have the flexibility to set up reciprocal relation or not.
**praise:** Good to have error message added into the model class `Relation.php` for the validation.
**praise:** Good to have create constant array `RECIPROCAL_RELATION`, which is clearer and more readable.
**question:** Usually and normally, the `relation_doi` is 6 digits, eg, 100148, which unlikely would exceed 15 char. So, it better to set it to `6`?
**question:** Why you have removed the unit test `RelationDAOTest.php`? That unit tests were to test the `createReciprocalTo()` function itself, while the functional test is more about the integration part.
**suggestion(non-blocking):** Is really not a good practice to make changes in 9 files in just 1 commit. Can look into the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) which we are currently following.

alli83:
@kencho51 if the rule is clear and it doesn't exceed 6 digits, I can update it, no problem 
I added the unit test. 
About the 9 files in 1 commit, I combined them into the same commit because they are quite related and involve small changes

tests/unit/RelationDAOTest.php
```diff
@@ -1,73 +1,81 @@
 <?php
 
-class RelationDAOTest extends CDbTestCase
+/**
+ * run with:
+ * docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit RelationDAOTest
```
pli888, line 5: **note (non-blocking)**: I've been looking into why I sometimes get errors when running this test. In `dev` environment, when I run `./up.sh` to start GigaDB instance and then immediately run this test using `docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit RelationDAOTest` then I see error messages:
```
1) RelationDAOTest: Itshould add reciprocal relation | #0
 Test  tests/unit/RelationDAOTest.php:testItshouldAddReciprocalRelation
                                                                           
  [PHPUnit_Framework_Exception] Trying to get property 'id' of non-object  
                                                                           
#1  /var/www/protected/models/RelationDAO.php:64
#2  /var/www/tests/unit/RelationDAOTest.php:65
#3  RelationDAOTest->testItshouldAddReciprocalRelation
```

However, running `docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit` will allow `RelationDAOTest` to pass. If you then run `docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit RelationDAOTest` then this will pass too. Feels like running `docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit` makes `docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit RelationDAOTest` to use test data in `protected/tests/fixtures`. If you execute `docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit RelationDAOTest` immediately after `./up.sh` then dev data is used in unit tests which may be related to why it was sometimes failing for me.

tests/unit/RelationDAOTest.php
```diff
@@ -1,73 +1,81 @@
...
+ */
+class RelationDAOTest extends \CDbTestCase
 {
     /**
      * @dataProvider relationshipProvider
      */
     public function testItshouldAddReciprocalRelation($relationship, $expected_reciprocal)
```
pli888: **suggestion**: Could capitalise the `s` in `should` in the `testItshouldAddReciprocalRelation` function name. This would fix the output below when running this test:
```
✔ RelationDAOTest: Itshould add reciprocal relation | #0 (0.22s)
```

pli888 (approved):
**praise**: Functionality is working in `dev` and `staging` for me when creating a relation between 2 datasets. Their reciprocal relationship is created too. 
**praise**: Code changes looks good. Good to see code in RelationDAO has been improved with a constant array RECIPROCAL_RELATION describing reciprocal relations enabling removal of if-else statements in createReciprocalTo(Relation $relating_rel, Relation $reciprocal_rel) function.
**note (non-blocking)**: New relationship types defined in constant array RECIPROCAL_RELATION might need to be added into the `relationship` table in order for curators to used them.
**chore**: Rebase with latest `develop` code to remove conflict in CHANGELOG.md.

Happy to approve this PR after rebasing.

rija (dismissed):
Hi @alli83,

**praise**: the bug is fixed in my testing
**praise**: addition of default-checked checkbox for adding a reciprocal relationship is an inspired idea, and it works in my testing
**praise**: the reuse of and the improvement to the Yii control widget for the check-box is nice
**praise**: the creation of an array for all relationships and their reciprocal is an improvement to what we had before 
**issue**: there is a disconnect and redundancy by having the relationships in the array constant and in the DB's `relationship` table: what if a relationship name needs to change (or new ones to be added)? there would be three places where it needs to be modified, one of them being code, one test and the database.
**suggestion**: If we were to fix that in this PR, I'd suggest to: 
* Create a schema migration to add a column to the relationship table for the reciprocal relationship
* Create a data migration to load all the supported relationships and their reciprocal into the DB for all environments
* Replace the array constant with a lookup method on the relationship table
* Update the test dataProvider function to pull the list from the database

Alternatively, and since the main blocker is fixed and it's already an improvement, we could create a new Github issue to perform that suggestion at a later time. See below why this could be the better choice

**issue** (non-blocking): I noticed that the Dataset Relation functionality in the admin dashboard doesn't have acceptance tests associated with it. 
But without a user story and acceptance tests at your disposal I do not expect you to be able to write new ones.
If there are other Github issue associated with this functionality (like the one I suggested to create in my previous point, or the though below), then we get an opportunity to work with the business to come up with new acceptance tests)

**thought**: I would be reluctant at this stage to add the new relationships to the database as suggested by @pli888, because they seem to come from a version of DataCite schema more recent than the one we use, and I don't know the extent of their server side validation:
When submitting a new dataset to Datacite API to have a new DOI minted or the metadata updated, we risk validation errors because our payload would contain relationships that do not match the version of the schema we have implemented and are advertising.
If that wasn't a problem (or when we resolve it), then the proper way to load the relationships in the database would be to create a data migration.

**praise**: In the absence of acceptance tests, the `RelationDAOCest.php` functional test is nice improvement to the coverage


Based on the above, I'm happy to approve this PR, while welcoming your comments on the issue and thought I've raised.

alli83:
I've added the acceptance tests and also ensured that it's not possible to create a relation if the DOIs are the same. 
Regarding the data insertion in db, as discussed, it will be considered in another PR

rija (commented):
Hi @alli83,

**issue**: In `tests/acceptance/RelationDAO.feature` the scenario "Add a relation and the reciprocal relation" fails for me.
The relation view page shows the ID of the initiating dataset (8), and the DOI of the related dataset (100039) but the test scenario is checking for the ID of the related dataset (5) which is not displayed on the relation view page, so the test fails.
The second test scenario fails for the same reason.

**suggestion**: the scenario step should check for the existence of the related DOI instead of the related ID

**praise**: I can confirm it is now disallowed to associate a dataset with itself

alli83:
@rija yep sorry just a bit confused about dataset and doi. It should be good.

rija (approved):
Hi @alli83,

I can see that the failed tests are now passing, 
so I'm happy to approve this PR again.

Issue #2015, opened 2024-08-22
An error statement shows when try to remove files in admin page


YFan0019:
**Describe the bug**
An error shows up when click the delete button in the "Dataset:Files" admin page

**To Reproduce**
1. Go to https://gigadb.org/adminFile/admin
2. Search for the files that I want to remove
3. Click on the delete button
4. See the error window shows up, but the function still works.

**Expected behavior**
Please check the cause and remove the error.

**Screenshots**
<img width="1149" alt="截屏2024-08-22 10 11 18" src="https://github.com/user-attachments/assets/63096ca7-e76e-418f-b794-2a86646f5fd1">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 127.0.6533.120

**Additional context**

rija:
This is likely an error thrown by the Yii framework because of the wrong method being used (GET instead of POST)

pli888:
Work merged in PR #2017 and #2042.

PR #2017, opened 2024-08-26
fix: error when deleting file from admin dashboard

alli83:
#2015

fix the error when deleting file from admin dashboard

## How to test?

Go to https://gigadb.org/adminFile/admin
Search for the files that I want to remove
Click on the delete button
remove the line/file

Needed to remove the final if else statement in order to avoid the error

### DIFF
--- a/protected/controllers/AdminFileController.php
+++ b/protected/controllers/AdminFileController.php
@@ -483,40 +483,56 @@ class AdminFileController extends Controller
     /**
      * Deletes a particular model.
      * If deletion is successful, the browser will be redirected to the 'admin' page.
+     *
      * @param integer $id the ID of the model to be deleted
      */
-   public function actionDelete($id)
+   public function actionDelete(int $id)
     {
-
-        if (Yii::app()->request->isPostRequest) {
-            // we only allow deletion via POST request
-            $file = File::model()->findByPk($id);
-            // $file->fileSamples->delete();
-          foreach ($file->fileAttributes as $fileattributes) {
-              print_r($fileattributes);
-              $fileattributes->delete();
-
-          }
-         foreach ($file->fileSamples as $filesample) {
-              print_r($filesample);
-              $filesample->delete();
-
-          }
-
-         $file->delete();
-
-        }
-
-            // if AJAX request (triggered by deletion via admin grid view), we should not redirect the browser
-
-        if (!isset($_GET['ajax'])) {
-            $this->redirect(isset($_POST['returnUrl']) ? $_POST['returnUrl'] : array('admin'));
-        } else {
+        // we only allow deletion via POST request
+        if (!Yii::app()->request->isPostRequest) {
             throw new CHttpException(400, 'Invalid request. Please do not repeat this request again.');
         }
 
-    }
+        $file = File::model()->findByPk($id);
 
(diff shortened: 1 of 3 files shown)


protected/controllers/AdminFileController.php
```diff
@@ -483,41 +483,35 @@ public function actions()
...
+            $filesample->delete();
         }
 
-            // if AJAX request (triggered by deletion via admin grid view), we should not redirect the browser
+        $file->delete();
 
-        if (!isset($_GET['ajax'])) {
+        // if AJAX request (triggered by deletion via admin grid view), we should not redirect the browser
+        if (!Yii::app()->request->isAjaxRequest) {
             $this->redirect(isset($_POST['returnUrl']) ? $_POST['returnUrl'] : array('admin'));
```
alli83: I reckon this part and this variable are not relevant and should be deleted since they are not being received.

rija: Hi @alli83 , 
yes, make sense to remove that block if that code is never exercised

pli888 (commented):
**praise**: Delete file functionality at `/adminFile/admin` works on my dev environment.
**thought**: The new `AdminFileCest` test works for me but what it is doing may be better suited as an acceptance test. If it was re-written as an acceptance test, there would still be the problem of how to target the trash icon because I think it doesn't have an id attribute in its HTML. Perhaps, update the relevant view class so trash icons have id attributes then their associated links can be clicked in a Codeception acceptance test class.

alli83:
@pli888 it's not going to be a problem.

protected/controllers/AdminFileController.php
alli83: I think it also needs to handle the case if delete return false or throw an exception. should I open a new ticket?

rija (commented):
Hi @alli83,

**praise**: functionality works
**praise**: great to see improved acceptance test coverage
**chore**: add entry to CHANGELOG.md
**nitpick** (non-blocking): a short explanation of the bug and why the changes in this PR fixes it in the "How have functionalities been implemented?" section of the PR would have been welcome
**praise**: code looks fine

pli888 (approved):
**praise**: The new acceptance test looks good to me.

Happy to approve this PR.

protected/controllers/AdminFileController.php
rija: Hi @alli83, you can handle the delete error in this PR. It's a small change, and it's possible for delete to throw exception and return false.

rija (commented):
Hi @alli83,

**praise**: great to see transactions used around cascading delete of files attributes
**praise**: great to see exceptions thrown  by `delete()` being caught
**question**: the phpdoc for `framework/db/ar/CActiveRecord.php`'s `delete()` function is 
```
/**  
     * Deletes the row corresponding to this active record.  
     * @throws CDbException if the record is new  
     * @return boolean whether the deletion is successful.  
     */
```

If the deletion fails because of database issue, will it throw an exception (which the code capture) or returns false (which the code doesn't check for)? 

**praise**: thanks for adding a CHANGELOG.md entry

rija (approved):
Hi @alli83,

**praise**: the changes capture more error conditions
**praise**: Code looks fine

Happy to approve

Issue #372, opened 2019-12-02
The Mint DOI function should save the DataCite payload locally for auditing 


only1chunts:
## User Story

>As a curator
>I want the Mint DOI button to save the payload sent to DataCite
>So that we can audit the medata sent in case of dispute

## Acceptance criteria

>Given I am on the admin page for a dataset
>When I press the Mint DOI button
>Then the DataCite metadata is generated and sent to DataCite API
>And the DataCite metadata is saved in the dataset table
>And an entry linking to the payload is added to the curation log

## Additional Info

**Background**
Currently the Mint DOI button just creates an XML and send it to Datacite to be minted without saving a local, this is fine as long as it works, but when it fails it would be better if we could see the XML to assist with debugging.

**Describe the solution you'd like**
Given that I am logged in as Admin
When I click Mint DOI button
Then DataCite XML is generated and saved to server somewhere, 
And that XML is used to mint DOI at DataCite (as currently happens)

## Implementation

* [ ] #1774
* [x] #1775
* [ ] #1776
* [ ] Add action in AdminDatasetController to return the Datacite Payload for the dataset
* [ ] The Mint DOI entry in the curation log should have a link to andpoint showing the DataCitePayload

## Resources

* https://www.postgresql.org/docs/current/datatype-xml.html

only1chunts:
This could be done in combination with work in ticket #414

alli83:
And an entry linking to the payload is added to the curation log: what is needed? the all xml payload or some info?

only1chunts:
as per the additional info above "Then DataCite XML is generated and saved to server somewhere" i.e. the entire XML payload generated by the Mint DOI button on that occassion.

alli83:
ok so I guess, you want a link and a modal with the all xml saved?

alli83:
of maybe under the Mint button, we could have a link to access the xml saved once it's done instead of having it in the curation log?

rija:
Hi @only1chunts
I think @alli83 meant the content of the curation log entry.
We are already going to save the payload in the database, but does the curation log entry added when minting a DOI needs updating to somehow refers to the DataCite payload?
I don't think it's a good idea to put the full payload in the curation log entry as it can be quite big.
and we don't want to click on a link that will load the XML in the browser because and error will be thrown as the XML doesn't have stylesheet anwyere.

what we could possibly do is trigger the download of a file containing the payload from a link in the curation log entry

[...]

PR #2112, opened 2024-12-01
feature: 372: save dataset as xml in curation log

alli83:
# Pull request for issue: #https://github.com/gigascience/gigadb-website/issues/372

Datacite xml payload saved in log
Add unique constraint to identifier (Dataset model)

## How to test?

Given that I am logged in as Admin
When I click Mint DOI button and refresh the page
Then DataCite XML is generated and saved in the curation log,

acceptance test added


rija (commented):
Hi @alli83,

**praise**: functionality works

**issue**: the type column in the curation log entry is currently "XML Saved", which is too vague (and the fact that its shows in the log implies that is saved, no need to spell that out)
**suggestion**: Use "Sent DataCite XML" 

**issue**: `protected/extensions/CTypeFormatter.php` is missing comments explaining what it is for
**suggestion**: add PHPDoc block to the class with explanation on the reason it exists

**issue**: there are changes to the related identifiers which don't seem to be part of the PR's goal and they are not mentioned in any of the section of the PR description
**suggestion**: Add more info to the PR description about all the changes made in this PR

**praise**: all the tests are passing

**note**: @kencho51, @pli888 when deploying this to live production we need to remember to run DB migrations manually there, because I just noticed  on AWS they are currently run only as part of `databaseReset`, which we cannot run on live production. I'll create a ticket to make the schema migration also a separate gitlab job before deploy going forward

kencho51 (commented):
Hi @alli83 ,

**praise:** Can see the xml was generated and can be seen in the curation log comment column.

**praise:**  The acceptance test is passing.

**praise:** The newly introduced JS code can collapse and expand the xml.

**praise:** The CTypeFormatter checks if the data contains XML-like data and then change the type to raw, which ensure it is displayed as raw HTML (or preformatted text) instead of being escaped or processed as plain text.

**issue:** There is no test for the new component class `LogCurationFormatter.php`.
**suggestion:** Better to have unit test for it.

rija (approved):
Hi @alli83

**praise**: thanks for making the requested changes, looks good now
**praise**: you can ignore my comments about unrelated changes to related identifiers.
After wearing my glasses, I could see it was a minor refactoring out of "suboptimal" code :-)
**note**: @kencho51, I do not think a unit test is necessary as the two pathways of the code are exercised, respectively, by the new acceptance tests in `tests/acceptance/AdminDatasetCurationLog.feature` and the existing acceptance tests from `tests/acceptance/AdminCurationLog.feature`

Based on the above, I'm happy to approve

alli83:
@rija and @kencho51 I added unit tests

kencho51:
@alli83,

Thanks for adding the unit test.

Regarding the unit test, I found:

**issue:** The class name is not reflecting the class that we are testing for
**issue:** The unit test is extending the test framework that we try to replace, as we want to use codeception 
**suggestion:** The above two issues can be simply fixed by executing:
```
% docker-compose run --rm test ./vendor/codeception/codeception/codecept generate:test unit LogCurationFormatter
Test was created in /var/www/tests/unit/LogCurationFormatterTest.php
```
The command will generate unit test with the actual class name and extend the codeception test framework. Then, you just need to move all the existing codes in CustomCFormatterTest.php to it, the test will pass as below:
```
% docker-compose run --rm test ./vendor/codeception/codeception/codecept run unit tests/unit/LogCurationFormatterTest.php 

Running with seed: 


Unit Tests (2) ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
✔ LogCurationFormatterTest: It returns right type (0.00s)
✔ LogCurationFormatterTest: It displays correctly the xml attr (0.00s)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


Time: 108 ms, Memory: 14.00MB

OK (2 tests, 3 assertions)
```

kencho51 (approved):
Hi @alli83,

Thanks for the update, LGTM now, happy to approve

PR #1978, opened 2024-07-08
fix-1727: dataset api sub sql queries for files or samples sorted by id in descending order

alli83:
# Pull request for issue: #1727 

This is a pull request for the following functionalities:

sort samples and files by id in descending order

## How to test?

e.g: /api/dataset?doi=10035&result=sample

## How have functionalities been implemented?

update relations in dataset model

## Any changes to automated tests?

ApiSearchTestCest added

Issue #1975, opened 2024-07-04
API broken


only1chunts:
## User story

>As a user  
>I want to be able to pull GigaDB dataset metadata from the API
>So that I can gather details about datasets for futher analysis

## Acceptance criteria

>Given I follow the instructions on the API help page (https://gigadb.org/site/help#interface)
>When I select the example as shown "http://gigadb.org/api/dataset?doi=100051"
>Then I see the correct results and not an error message

There are 2 issues, the first is that the documentation still uses http instead of httpS, so that does need updating, but even after making that change it still failes with the error shown in additiona info section.

## Additional Info
This is the error message that is shown (after correcting the above call to use https instead of http)

![image](https://github.com/gigascience/gigadb-website/assets/6037145/0807f6cb-6319-4ee9-acb2-6e83c30d664d)

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
* [ ] The product is in a releasable state (i.e. the increment has not broken anything)

rija:
Hi @only1chunts 

the API is not broken, it is not meant to be read by human on a browser, but read programmatically,
if you check the source page, you will see the data returned by the API call.

If you want to view the api response from your browser you should install an API client  browser extension

only1chunts:
programatically I would use this command:
$ wget http://gigadb.org/api/dataset?doi=100051
but that also fails :-( although the error looks different to above:

```
» wget https://gigadb.org/api/dataset\?doi\=100051
--2024-07-04 09:22:58--  https://gigadb.org/api/dataset?doi=100051
[...]

PR #2118, opened 2024-12-04
Fix 1975 remove empty line at the beginning of the xml result

alli83:
# Pull request for issue: #1975

need https://github.com/gigascience/gigadb-website/pull/1978 merged before

remove empty line at the beginning of the xml

## How to test?

Given I follow the instructions on the API help page (https://gigadb.org/site/help#interface)
When I select the example as shown "http://gigadb.org/api/dataset?doi=100051"
Then I see the correct results and not an error message

tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix #1975: Remove empty line at the beginning of the xml result
 - Feat #456: Improve DataCite metadata by migrating to Datacite version 4.6
 - Fix #1727: Sort files and samples by id in descending order when querying
 
--- a/protected/controllers/ApiController.php
+++ b/protected/controllers/ApiController.php
@@ -2,9 +2,6 @@
 
 class ApiController extends Controller
 {
-    // Members
-
-
     const RESULTS = ['file', 'sample', 'dataset'];
 
 	/**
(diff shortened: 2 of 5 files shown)


kencho51 (approved):
Hi @alli83,

**praise:** All tests are passing both in my dev and gitlab CI/CD pipelines.
**praise:** The new functional tests confirm that the xml generated is a valid xml.
**praise:** Was able to make api call in both my dev and staging.
**question(non-blocking):** While using ob_start() and ob_get_clean() can be a quick solution to manage output and clean up unwanted whitespace, why dont generate the XML string directly and then use trim() on clean it? This can ensure any unwanted whitespace is removed without the need for buffering which may consume more memory.

Since no any blocking from my side, happy to approve.

pli888 (approved):
**praise**: Code in this PR allows a call to the `dataset` API  operation to return valid XML containing data about a dataset in GigaDB. There isn't an error message `XML declaration allowed only at the start of the document` anymore when pointing a web browser to http://gigadb.gigasciencejournal.com/api/dataset?doi=102484. Functionality is working on my staging server too. Closer inspection of the XML when saved as a file shows that it no longer contains an empty first line which was the cause of the problem described in #975 API broken.
**praise**: New tests in `ApiSearchTestCest.php` are passing in dev and on my CI pipeline.
**praise**: Code looks good to me
**nitpick**: Would be useful if a bit more documentation was provided in the PR. I tend to, for example, provide commands to help reviewers run tests and test functionality in dev env:
```
docker-compose run --rm test ./vendor/codeception/codeception/codecept run --debug tests/api/ApiSearchTestCest.php
```

Or provide a working URL example to test functionality in dev: http://gigadb.gigasciencejournal.com/api/dataset?doi=102484

Issue #2138, opened 2024-12-20
unchecking a dataset type for a dataset will delete that type from all the other datasets that had it


rija:
**Describe the bug**
```
CONFIRMED
I just tried on staging to remove the Dataset Type Genomic from dataset 102628 and it has removed ALL instances of it.
Incase its relevant, the dataset status was set as "import from EM" 

So we know what the cause of the missing dataset types is, we just need to get it fixed, in the mean time 
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

alli83:
https://github.com/gigascience/gigadb-website/pull/2147

pli888:
Closing this issue because there is a fix in PR #2147

PR #2147, opened 2024-12-23
fix 2138: avoid deleting type when deleting a dataset type

alli83:
# Pull request for issue: #2138

avoid deleting type when deleting a dataset type

## How to test?

Describe how the new functionalities can be tested by PR reviewers

Go to 'dataset'
uncheck a dataset type
Scroll down to save
you should see the dataset type but unchecked

### DIFF
--- a/protected/models/Dataset.php
+++ b/protected/models/Dataset.php
@@ -715,12 +715,14 @@ class Dataset extends CActiveRecord
     public function updateDatasetTypes($postDatasetTypes)
     {
         $actualTypeIdsByDataset = [];
+        //fetch types
         $datasetTypeMaps = $this->datasetTypes;
+        $command = Yii::app()->db->createCommand();
 
         foreach ($datasetTypeMaps as $datasetTypeMap) {
             $actualTypeIdsByDataset[] = $typeId = $datasetTypeMap->id;
             if (!in_array($typeId, $postDatasetTypes, true)) {
-                $datasetTypeMap->delete();
+                $command->delete('dataset_type', 'dataset_id=:dataset_id AND type_id=:type_id ', array(':dataset_id' => $this->id, ':type_id' => $typeId));
             }
         }
 
--- a/tests/acceptance/AdminDatasetForm.feature
+++ b/tests/acceptance/AdminDatasetForm.feature
@@ -521,3 +521,18 @@ Feature: form to update dataset details
     And I wait "5" seconds
     When I am on "/adminDataset/update/id/5"
     Then I should see "Dataset_Genomic" checkbox is checked
+
+  @ok
+  Scenario: Check checkbox for type is working
+    Given I am on "/adminDataset/update/id/8"
+    And I should see "Dataset_Genomic" checkbox is checked
+    When I uncheck "Dataset_Genomic" checkbox
+    Then I should see "Dataset_Genomic" checkbox is unchecked
+    And I should see "Dataset_Workflow" checkbox is unchecked
+    When I check "Dataset_Workflow" checkbox
+    Then I should see "Dataset_Workflow" checkbox is checked
+    When I press the button "Save"
+    And I wait "5" seconds
+    And I am on "/adminDataset/update/id/8"
+    Then I should see "Dataset_Workflow" checkbox is checked
+    Then I should see "Dataset_Genomic" checkbox is unchecked
--- a/tests/acceptance/AdminDatasetForm.feature
+++ b/tests/acceptance/AdminDatasetForm.feature
@@ -496,3 +496,28 @@ Feature: form to update dataset details
     And I press the button "Next >"
     And I wait "1" seconds
     Then I should see "Parrot.k31.NetworkTest.txt"
+
+  @ok
+  Scenario: Check type is not removed when uncheck a dataset type
+    Given I am on "/adminDataset/update/id/8"
+    Then I should see "Workflow"
+    Then I check "Dataset_Workflow" checkbox
+    And I press the button "Save"
+    When I am on "/adminDataset/update/id/8"
+    Then I should see "Dataset_Workflow" checkbox is checked
+    Then I uncheck "Dataset_Workflow" checkbox
+    And I press the button "Save"
+    When I am on "/adminDataset/update/id/8"
+    Then I should see "Workflow"
+
+  @ok
+  Scenario: Check type is not removed from other dataset when uncheck for another dataset
+    Given I am on "/adminDataset/update/id/8"
+    Then I should see "Genomic"
+    Then I should see "Dataset_Genomic" checkbox is checked
+    Then I check "Dataset_Workflow" checkbox
+    Then I uncheck "Dataset_Genomic" checkbox
+    And I press the button "Save"
+    And I wait "5" seconds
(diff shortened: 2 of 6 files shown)


tests/acceptance/AdminDatasetForm.feature
```diff
@@ -496,3 +496,16 @@ Feature: form to update dataset details
     And I press the button "Next >"
     And I wait "1" seconds
     Then I should see "Parrot.k31.NetworkTest.txt"
+
+  @test
```
kencho51: **suggestion:** If the scenario is ready, in our convention, it should be `@ok`.

tests/acceptance/AdminDatasetForm.feature
```diff
@@ -496,3 +496,16 @@ Feature: form to update dataset details
...
+    Given I am on "/adminDataset/update/id/8"
+    Then I should see "Workflow"
+    Then I check "Dataset_Workflow" checkbox
+    And I press the button "Save"
+    When I am on "/adminDataset/update/id/8"
+    Then I should see "Dataset_Workflow" checkbox is checked
+    Then I uncheck "Dataset_Workflow" checkbox
+    And I press the button "Save"
+    When I am on "/adminDataset/update/id/8"
+    Then I should see "Workflow"
```
kencho51, line 511: **issue:** I think this scenario does not catch quite right, as stated in #2138, first, each dataset is by default to have at least one dataset type checked, if someone is trying to uncheck the only checked type, it will throw error `Fail to update your types`, like the above scenario, the update action actually has not triggered yet.

**suggestion:** For example the scenarios can be:

```
 Scenario: Type is not removed when uncheck a dataset type
    Given I am on "/adminDataset/update/id/8"
    And I should see "Workflow"
    And I should see "Genomic"
    When I check "Dataset_Workflow" checkbox
    And I press the button "Save"
    And I am on "/adminDataset/update/id/8"
    And I should see "Dataset_Workflow" checkbox is checked
    And I should see "Dataset_Genomic" checkbox is checked
    Then I uncheck "Dataset_Genomic" checkbox
    And I press the button "Save"
    And I am on "/adminDataset/update/id/8"
    And I should see "Genomic"
```

**issue:** Second, when someone unchecked a type from one dataset and saved, the type of the other dataset containing the same type should not be removed.
**suggestion:** For example the scenarios can be:

```
Scenario: Type of another dataset is not removed when uncheck a dataset type
    Given I am on "/adminDataset/update/id/8"
    And I should see "Workflow"
    And I am on "/adminDataset/update/id/200"
    And I should see "Workflow"
    When I am on "/adminDataset/update/id/8"
    And I check "Dataset_Workflow" checkbox
    And I press the button "Save"
    And I am on "/adminDataset/update/id/8"
    Then I uncheck "Dataset_Genomic" checkbox
    And I press the button "Save"
    And I am on "/adminDataset/update/id/200"
    And I should see "Genomic"
```

protected/models/Dataset.php
```diff
@@ -715,12 +715,14 @@ public  function nullifyDateValueIfEmpty()
...
         $actualTypeIdsByDataset = [];
+        //call getDatasetTypes method  => will return types
         $datasetTypeMaps = $this->datasetTypes;
+        $command = Yii::app()->db->createCommand();
 
         foreach ($datasetTypeMaps as $datasetTypeMap) {
             $actualTypeIdsByDataset[] = $typeId = $datasetTypeMap->id;
             if (!in_array($typeId, $postDatasetTypes, true)) {
-                $datasetTypeMap->delete();
+                $command->delete('dataset_type', 'dataset_id=:dataset_id AND type_id=:type_id ', array(':dataset_id' => $this->id, ':type_id' => $typeId));
```
kencho51, line 725: **question(non-blocking):** Why you would prefer executing the raw sql to deletes records from the database, instead of deleting the specific record by invoking the ActiveRecord instance delete() method? I think through the delete() method, it interacts with the database through the model's abstraction layer and can trigger model level logic.

pli888 (changes requested):
**nitpick**: Tiny bit more constructive information on how to test the functionality in this PR would've been useful. Using the current instructions, I get a `Fail to update your types` error message.
**praise**: Changing the dataset type functionality works as expected after I realised that I need to check another dataset type checkbox after unchecking the current dataset type.
**issue**: I think the acceptance test in `AdminDatasetForm.feature` could be improved so that steps a curator would need to do to change dataset type are defined. Maybe use the `I should see :checkbox checkbox is not checked` assertion to ensure regression of this bug does not happen.
**chore**: Rebase with current develop branch code to resolve conflict in changelog.

tests/acceptance/AdminDatasetForm.feature
alli83: I'm not sure I understand your comment. The purpose of the test is simply to show that when a type is unchecked, the type itself is not deleted (which was the case before)

protected/models/Dataset.php
alli83: the purpose here was just to fix the bug. the problem here is that we are not fetching datasetTypes but rather the type so If I call delete on the model, it will delete the type.

tests/acceptance/AdminDatasetForm.feature
kencho51: From what I got from the ticket #2138 is when someone tries to uncheck a dataset type from one dataset, it will be removed not just from that particular dataset, but also from other datasets as well.

So, the above scenarios I suggested are to make sure that a dataset type can be unchecked from one dataset, and other dataset wont be affected.

alli83: the main cause is that the dataset type itself has been removed.

alli83: I just added another test to double-check but the main cause is the disappearance of te type itself

pli888 (approved):
**praise**: The code in `protected/models/Dataset.php` fixes the bug by deleting the relevant row(s) in `dataset_type` table when checked Dataset Type checkboxes are unchecked.
**praise**: The new acceptance tests are a welcomed improvement for testing dataset type functionality on dataset admin pages. These tests are passing on my local dev environment.

This PR looks good to me now.

protected/models/Dataset.php
rija: @alli83, @kencho51,
I think the SQL is fine for this PR given it's high impact bug, but I also share  @kencho51's concern, it should just work with the Yii framework's ActiveRecord implementation, that's what it's for.
I think we are misusing it. I suspect this line is the culprit or part of the problem:
```
120 'datasetTypes' => array(self::MANY_MANY, 'Type', 'dataset_type(dataset_id,type_id)'),
```
but we can fix that in another PR.

alli83: I totally agree that some things need to be rewritten and are not quite right. But as we discussed, the PR should not mix refactoring with bug fixing

rija (approved):
Hi @alli83,

**praise**: bug fix works in my setup
**praise**: test coverage has been improved
**praise**: code is good enough

Happy to approve

kencho51 (approved):
Hi @alli83,

LGTM, happy to approve.

Issue #2099, opened 2024-11-20
CLOCKSS  permission file missing from web site preventing CLOCKSS systems to index our content


only1chunts:
## User story

>As a third party system
>I want to be allowed to access GigaDB content
>So that I index its content

## Acceptance tests

>Given  I need to assess I can index GigaDB content
>When I navigate to /lockss.txt
>Then I can see "CLOCKSS system has permission to ingest, preserve, and serve this open access Archival Unit."

## Additional infos

**Describe the bug**
From Meghan at CLOCKSS (users of our API)

> 
> Our developer took a look and said that it turns out that the permission URL is no longer working, so the crawler is no longer getting as far as the syntax problem.
> 
> This page used to contain the CLOCKSS permission statement
> https://gigadb.org/lockss.txt
> 
> Now, we receive the following error:
> CDbException
> CDbConnection failed to open the DB connection.
> 
> The permission statement that should be on that page is:
> CLOCKSS system has permission to ingest, preserve, and serve this open access Archival Unit.
> 
> Please let me know if more information or a call to discuss would be helpful.
> 
> Thanks,
> Meghan

rija:
A static file is missing, the support person actually specified in her email what should go in that file
Relevant documentation:

https://clockss.org/preserving-your-publications-in-clockss/

rija:
Fix is in PR #2185 2185

PR #2185, opened 2025-02-03
Fix: Add missing permission for CLOCKSS

rija:
# Pull request for issue: #2099

Add a publicly accessible static file on the website called `lockss.txt` which contain permission statement from:
https://clockss.org/preserving-your-publications-in-clockss/

## How to test?

Navigate to:

http://gigadb.gigasciencejournal.com/lockss.txt

It should show the permission text.

After pushing this PR to gitlab pipeline for your AWS environment, check on AWS deployment,  that the file is readable and has the permission text.

# How have functionalities been implemented?

static file added at the root of the website.

n/a

New acceptance scenario and a new user story user (Third party system)

Added a troubleshooting note for correct Gitlab user role needed to download variables

Remove Firefox installl from the JS container, as it is no longer  used
and caused the container to fail on build due to change to Firefox download URL.

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix #2099: Add missing permission for CLOCKSS
+
 ## v4.4.3 - 2025-02-03 - c62107e5e
 
 - Feat #474: Update FAQ "What journals are integrated with GigaDB?"
--- a/docs/SETUP_CI_CD_PIPELINE.md
+++ b/docs/SETUP_CI_CD_PIPELINE.md
@@ -424,4 +424,9 @@ The following Gitlab variables are needed for the acceptance run, both in the pi
 | AWS_SECRET_ACCESS_KEY | your secret key to AWS | All |
 
 
+## Troubleshooting
 
+### permissions
+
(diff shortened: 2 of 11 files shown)


kencho51 (approved):
Hi @rija,

praise: The static ockss.txt is working in both my local and staging environments.
praise: The acceptance is passing locally.
question(non-blocking): Why introduce ThirdPartySteps for the new acceptance test as there is I have not signed in step in CuratorSteps.php already.?
question(non-blocking): Whats the purpose of bringing in the nginx.production.conf?

No blocking issue, LGTM, happy to approve.

rija:
> Hi @rija,
> 
Hi @kencho51 

> question(non-blocking): Why introduce ThirdPartySteps for the new acceptance test as there is I have not signed in step in CuratorSteps.php already.? 

The "As a  (persona)" in the Feature's user story is not just for documentation, it defines the scope for  using custom steps for that feature file, so that every features that share the same "As a (persona)" can share the same steps.
Those steps are defined in `tests/_support/Steps/(persona).php`. 
Since this feature file uses a different persona (third party system) than existing features files, it does need its own Step files. (It cannot see the steps in the other step files).

> question(non-blocking): Whats the purpose of bringing in the nginx.production.conf?
> 

before, the rate limits configuration is global to all environments.  
It is not correct, as we want to set specific values to production to ensure staging and live are protected from DDOS,and different value for local dev (more lenient, so that repeated tests can be run without nginx activating rate limit defense)

Issue #1955, opened 2024-06-27
unable to edit users details via admin pages


only1chunts:
## User story

>As a curator  
>I want the user admin form to to not show captcha error
>So that I can update the details for the user

## Acceptance criteria

>Given I am on the user admin form for a user  
>When I change its details
>And I click Save  
>Then the details is saved to the database 
>And I am not asked to fill a captcha

>Given I am on the user admin form for a user  
>When I change its details
>And I click Save  
>Then the details is saved to the database 
>And I am not forced to provide a password for that user

## Additional Info

Currently the form requires that the admin person making the changes on the form must input a password for the user they are updating, this should NOT be mandatory as often we need to change something about the user details but we do not want to change their password for them. e.g. changing bastiens role from user to admin should not require us to change his password to something the admin person sets only for him to have to then go in and reset it to something private.

![image](https://github.com/gigascience/gigadb-website/assets/6037145/13bd8314-debd-4adb-ae82-e524f75c0892)

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
* [ ] The product is in a releasable state (i.e. the increment has not broken anything)

tuli:
We are unable to assign Bastien an Admin role.
Please can this issue be addressed @rija ?

rija:
Hi @tuli, @only1chunts 
I have manually changed Bastien's role in the database (he will need to logout and login again), 
so he's unblocked for now while this ticket goes through the usual process.

molcretb:
Hi Rija, thanks for the fix, I just logged in and it works :)

alli83:
[...]

PR #2191, opened 2025-02-11
Fix 1955 edit user

alli83:
# Pull request for issue: #1955 

add a regex for password
password is optional when I want to edit an user
captcha not required for update

## How to test?

Given I am on the user admin form for a user
When I change its details
And I click Save
Then the details is saved to the database
And I am not asked to fill a captcha
Given I am on the user admin form for a user
When I change its details
And I click Save
Then the details is saved to the database
And I am not forced to provide a password for that user

test added

### DIFF
--- a/protected/controllers/UserController.php
+++ b/protected/controllers/UserController.php
@@ -125,32 +125,28 @@ class UserController extends Controller {
     public function actionUpdate() {
         $user = $this->loadUser();
         $this->performAjaxValidation($user);
-        #if (!Yii::app()->user->checkAccess('updateOwnUser', array('user'=>$user))) {
-        #    Yii::log(__FUNCTION__."> Unauthorized", 'debug');
-        #    throw new CHttpException(403, 'You are not authorized to perform this action.');
-        #}
 
-
-
-
-        if (isset($_POST['User'])) {
-            $user->attributes = $_POST['User'] ;
-            $attrs = $_POST['User'];
-
-            $password = $user->password_new = $attrs['password'];
+        if ($attrs = Yii::$app->request->post('User')) {
+            $user->email = $user->username = strtolower(trim($attrs['email']));
+            $user->first_name = trim($attrs['first_name']);
+            $user->last_name = trim($attrs['last_name']);
+            $user->password_new = $attrs['password'];
             $user->password_repeat = $attrs['password_repeat'];
+            $user->role = $attrs['role'];
+            $user->affiliation = $attrs['affiliation'];
+            $user->preferred_link = $attrs['preferred_link'];
+            $user->newsletter = $attrs['newsletter'];
+            $user->terms = $attrs['terms'];
+
 
             if (!Yii::app()->user->checkAccess('admin')) {
                 $user->role = 'user';
             }
 
             if ($user->validate('update')) {
-                if ($password != '') {
-                    $user->encryptPassword();
-                }
+                $user->encryptPassword();
 
                 if ($user->save(false)) {
-
                     Yii::app()->user->setFlash('notice', 'Updated');
                     $this->redirect(array('user/show/id/'.$user->id));
                 }
@@ -163,6 +159,7 @@ class UserController extends Controller {
                 Yii::log(__FUNCTION__."> validation failed", 'warning');
             }
         }
+
         $user->password = $user->password_repeat = '';
         $this->render('update', array('model'=>$user));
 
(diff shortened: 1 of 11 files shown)


alli83:
imho, the separate model for changing one's own password is not required. I kept it because the goal here is to fix the bug but I believe it can be rewritten

rija:
@alli83 can you add which GitHub issue (issue number to they can be linked automatically) this PR is for, thanks.

protected/controllers/UserController.php
```diff
@@ -125,32 +125,28 @@ protected function performAjaxValidation($model) {
...
             if ($user->validate('update')) {
-                if ($password != '') {
-                    $user->encryptPassword();
-                }
+                $user->encryptPassword();
 
                 if ($user->save(false)) {
-
                     Yii::app()->user->setFlash('notice', 'Updated');
                     $this->redirect(array('user/show/id/'.$user->id));
```
kencho51: Hi @alli83,

**issue:** When the form is updated, the page will be redirected to a  broken page like this:
image

**suggestion:** Would it be better to redirect page to the view page as below:
image

kencho51 (changes requested):
Hi @alli83 

**praise:** The user form can be edit now.
**issue:** When I executed ` % ./tests/unit_functional_runner`, got this error:
```
There was 1 failure:

---------
1) DatasetAsXmlTest: Upload status validation
 Test  tests/unit/DatasetAsXmlTest.php:testUploadStatusValidation
Failed asserting that two strings are equal.
- Expected | + Actual
@@ @@
-'http://foo2.com'
+'http://foo.com'
#1  /var/www/tests/unit/DatasetAsXmlTest.php:67
#2  DatasetAsXmlTest->testUploadStatusValidation

FAILURES!
Tests: 254, Assertions: 792, Failures: 1.
```
**issue:** As mentioned in the comment, after save, the page would be redirected to `'user/show/id/'.$user->id` which looks broken.
**suggestion:** Maybe better redirect it view page as `'user/view/id/'.$user->id`, that would make it more consistent as the other pages.
**praise:** The fix related to the password is appreciated which makes it more secure, and it got covered by tests. Actually it could go into the technical debt section in the PR description.

kencho51 (approved):
Hi @alli83,

**praise:** After save, the the page is redirected to view page, instead of a broken page.
**suggestion (non-blocking): ** For the failing unit test `DatasetAsXmlTest: Upload status validation`, adding this line `./ops/scripts/setup_devdb.sh` could temp fix it, it wiil reset dev database with the `gigadb_testdata` dataset.

As no blocking issues, happy to approve.

rija (commented):
Hi @alli83,

**question**: both locally and on my Gitlab pipeline I have the two same acceptance tests failing:
```
Acceptance Tests (2) ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
✖ EditUser: Ensure I can edit a user without changing the password (3.57s)
✖ EditUser: Ensure I can edit a user with matching password (2.25s)
```

Am I missing something?

alli83:
@rija oops, a # character was missing.

rija (changes requested):
Hi @alli83,

**praise**: locally, the user form workflow works fine, both in update (with or without password changes) and when creating a new account
**issue**: On AWS deployment, when creating a new user account, the user form  throws upon submission a server error `Address in mailbox given [] does not comply with RFC 2822, 3.6.2.`  from `/var/www/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php(355)`

despite the email address inputed in the form being perfectly valid.

**suggestion**: I cannot pinpoint what change in this PR causes the error, but on staging and live the error doesn't happen (neither locally but that's maybe because it's local dev).
With the empty square bracket in the error message, it seems like the email address is not passed on to the newly created instance of User class. Although I can see this PR makes changes to that class, I can't see what's wrong.

**praise**: On AWS deployment, I can update a user account with and without password changes

alli83:
> `Address in mailbox given [] does not comply with RFC 2822, 3.6.2.` from `/var/www/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Mime/Headers/MailboxHeader.php(355)`

@rija is senderEmail defined: "from"?

rija (approved):
Hi @alli83 

praise: after correcting my variables, the error was gone

Happy to approve

Issue #1751, opened 2024-03-11
Implement app service to provide access to dataset upload status


rija:
## User story

>As a developer  
>I want all usage of upload status make call to methods or constants on a dedicated application service 
>So that whenever statuses need to be changed or amended there is only on place in the codebase to do so

## Acceptance criteria

>Given  there is a list of upload status
>When  a subsystem needs to use  "AuthorReview" as an upload status
>Then it makes a call to a service's method with "AuthorReview" as parameter that returns "AuthorReview"

## Additional Info

this is about centralising the code that manipulate the upload status.
There is a separate ticket #1755 for persisting them in the database

The idea is to ensure the logic of what is the id and what is the label of an upload status and the retrieval of the latter is handled in one place, a Yii2 service in `gigadb/app/services/UploadStatusService.php`

In this ticket a call to the service will return the same hardcoded label, so that current workflow still works.
But once #1755 is implemented, it will return the Id.

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
* [ ] The product is in a releasable state (i.e. the increment has not broken anything)

only1chunts:
Just a pause for thought, could the upload statuses be a table in the
database? Not sure what advantages that would have (if any) just throwing
it up as a possibility.
We will need both external and internal documentation of what each status
means, so potentially being able to make subtle adjustment to the name for
clarity in the future may be useful, meaning the consistent thing would be
the database ID rather than the name.
The database could include a description column.

rija:
Hi @only1chunts 

Could do. I once wondered why they were not when file types and dataset types are.
[...]

PR #2195, opened 2025-02-12
feature: add a dedicated service to handle upload status transition

alli83:
# Pull request for issue: #1751 

As a developer
I want all usage of upload status make call to methods or constants on a dedicated application service
So that whenever statuses need to be changed or amended there is only on place in the codebase to do so

## How to test?

Given there is a list of upload status
When a subsystem needs to use "AuthorReview" as an upload status
Then it makes a call to a service's method with "AuthorReview" as parameter that returns "AuthorReview"

yii2 services created

test added


rija:
Hi @alli83, 

Can you add a `CHANGELOG.md` and open this PR for review ?
I'm going to move it up in the "Reviewing required" column of the PRs project.
Thanks.

Issue #2235, opened 2025-03-10
Add a checkbox on manuscript admin page to indicate whether it is pre-print or not


rija:
## User story

>As a curator
>I want to able to add pre-print manuscript links
>So that we can link a pre-print version of a manuscript to a dataset

## Acceptance criteria

>Given I have pre-print link for a given manuscript to add to a GigaDB dataset
>When I go to the admin page for Dataset Manuscripts
>And I enter the identifier for the manuscript
>And I tick the checkbox "This link is a pre-print"
>Then the manuscript is saved as a pre-print manuscript

## Additional Info

the dataset page counterpart  is:
* [x] #2236

## Product Backlog Item Ready Checklist

* [ ] Business value is clearly articulated
* [ ] Item is understood enough by the IT team so it can make an informed decision as to whether it can complete this item
* [ ] Dependencies are identified and no external dependencies would block this item from being completed
* [ ] At the time of the scheduled sprint, the IT team has the appropriate composition to complete this item
* [ ] This item is estimated and small enough to comfortably be completed in one sprint
* [ ] Acceptance criteria are clear and testable
* [ ] Performance criteria, if any, are defined and testable
* [ ] The Scrum team understands how to demonstrate this item at the sprint review

## Product Backlog Item Done Checklist

* [ ] Item(s) in increment pass all Acceptance Criteria
* [ ] Code is refactored to best practices and coding standards
* [ ] Documentation is updated as needed
* [ ] Data security has not been compromised (with particular reference to the personal information we hold in GigaDB)
* [ ] No deviation from the team technology stack and software architecture has been introduced
* [ ] The product is in a releasable state (i.e. the increment has not broken anything)

alli83:
@rija @only1chunts is the manuscript supposed to be saved as a external link when created or updated? Because I don't see a connection between manuscript and external link

Would it be possible to have more details for " is saved as a pre-print manuscript". Do I need to create an external link and connect it to the manuscript?

only1chunts:
@alli83 the manuscript table is different to the external_link table, so no there should not be external_links added for manuscripts, the details are saved to the manuscript table.
@rija can confirm the technical side of things, but I'm guessing there will need to be a new column added to the manuscript table to indicate "pre-print" status of the manuscript.
In terms of the datacite metadata, then these pre-print manuscript links should be added to the datacite metadata in the same was as the normal manuscript links do at the moment, so no change of action there.
[...]

PR #2249, opened 2025-03-24
feat 2235: add a checkbox for manuscript considered a pre print link

alli83:
# Pull request for issue: #2235 

Add a checkbox to determine whether if a manuscript should be considered a pre-print link

- used yii2 models for DatasetLog and Manuscript. 
- Keep yii1 views and yii1 CForm => create a wrapper for my yii2 models in oder to be able to use CForm from yii1. 
- Fixed problem "data corrupted" on CI only for legacy_test job and not for prod environment

## How to test?

 I am on "/adminManuscript/create" or "/adminManuscript/update"
 And I should see a check-box field "is a pre-print link"

migration file created in order to add a column to manuscript table with a default value false

acceptance added

data corrupted for legacy_test job => fixed by switching to a temp directory to avoid data corruption caused by shared/persistent volumes only for legacy_tests job and not live


kencho51 (approved):
Hi @alis,

**praise:** The `is_pre_print` checkbox is created.
**praise:** The `is_pre_print` checkbox can be checked and un-checked.
**suggestion: (non blocking)** To make the test more consistent, the `Manuscript[identifier]` should be something like `10.1186/s13742-015-0064-7`.
**praise:** All tests are passing.

LGTM, happy to approve.

alli83:
@kencho51 ok, I'll update the test while rebasing the branch

rija (commented):
Hi @alli83,

**praise**: functionality works locally 
**note**:  The classes  in `protected/models/` are legacy and we are gradually migrating them to Yii2 model classes  in `gigadb/app/models/` but Manuscript is not there yet
**question**: Maybe this PR is the right time to create `gigadb/app/models/Manuscript.php` . Then you could modify `protected/controllers/AdminManuscriptController.php` to make use of that Yii2 class?

alli83:
@rija  I can do it. I will also need to create DatasetLog yii2 model for the behaviors and afterSave/afterUpdate/afterDelete behaviors too. But ofc I can work on it.

update: I have to rework the views

alli83:
@rija @only1chunts could you please confirm the constraints for the Manuscript and DatasetLog models so that I can implement them correctly in the yii2 models? thanks

only1chunts:
> @rija @only1chunts could you please confirm the constraints for the Manuscript and DatasetLog models so that I can implement them correctly in the yii2 models? thanks

@rija I'm leaving this one to you as I dont understand what @alli83 needs.

rija:
Hi @alli83,

for dataset log: 
entries can be added in dataset log only if the dataset is set to "Published"
(as for all the other upload status, it is the curation log that's being used)

for Manuscript:
both identifier, and dataset_id (which link the manuscript to an existing dataset) must not be null

alli83:
@rija there is no identifier in datasetLog

gigadb/app/models/DatasetLog.php
```diff
@@ -0,0 +1,125 @@
...
+
+    public static function createDatasetLogEntry(
+        int    $datasetId,
+        string $fileName,
+        ?string $fileModel = null,
+        ?int    $modelId = null,
+        ?int    $fileId = null
+    ): bool {
+        $datasetLog = self::makeNewInstanceForDatasetLogBy($datasetId, $fileName, $fileModel, $modelId, $fileId);
+        $datasetLog->message = $fileName . ': file attribute deleted';
```
alli83: It just a copy paste from the old model. but I noticed that it doesn't make sense to overwrite the message with 'File attribute deleted' (it could be something else). Could someone explain what was the logic behind this approach?

rija: Hi @alli83,
`createDatasetLogEntry` should just be  a generic purpose way of creating dataset log entry.
Therefore, I reckon the signature for `createDatasetLogEntry` should just be just (dataset_id, message)
and the line that's overwriting message with a concatenation of string should be performed in `DeleteFileAttributeAction` where createDatasetLogEntry is called (that's actually the only place that calls that method, make me wonder how many ways we have in the code to insert a dataset_log record.)

rija (changes requested):
Hi @alli83,

**praise**: tests are passing
**praise**: I really like how we can use the yii2 models by wrapping it with in a Yii1 proxy
**praise**: functionality works locally
**issue**: On AWS, got this error when navigating to the manuscript create or update form:
```
Failed to instantiate component or class "GigaDB\behaviors\DatasetRelatedTableBehavior".
```

**suggestion**: Adding a new entry in the psr-4 section of the composer file might not be enough?
@kencho51 had the exact same problem in one of his PR where he introduced an Yii2 model but I don't remember how it was fixed

**praise**: code looks good

alli83:
> DatasetRelatedTableBehavior

it is already in the PSR-4 section

kencho51:
Hi @alli83,

> **suggestion**: Adding a new entry in the psr-4 section of the composer file might not be enough?

Regarding this error, it is probably the file `gigadb/app/behaviors/DatasetRelatedTableBehavior.php` is not found in the production container, you need to bake it into  `ops/packaging/Production-Dockerfile` as:
```
COPY ./gigadb/app/behaviors /gigadb/app/behaviors/
```

Also refactor the `ops/configuration/php-conf/composer.json.dist` to include the path in the PSR4 section as:
```
"psr-4": {
            "common\\models\\" : "/app/common/models",     
            "backend\\models\\" : "/app/backend/models",
            "GigaDB\\models\\" : "/gigadb/app/models",
            "GigaDB\\services\\" : "/gigadb/app/services",
            "GigaDB\\behaviours\\": "/gigadb/app/behaviors"
        }
```
 
 After that, the new component `behaviours` should be available to use.

rija (changes requested):
Hi @alli83,

**suggestion**: To fix the issue mentioned in my previous comment, and as mentioned by @kencho51,  in `ops/packaging/Production-Dockerfile`,
add this line:
```
COPY ./gigadb/app/behaviors /gigadb/app/behaviors/
```
after line 199

rija (approved):
Hi @alli83 

praise: functionality is now fully working on my AWS deployment

Based on that, happy to approve

Issue #1814, opened 2024-04-22
RSS feed not working


only1chunts:
**Describe the bug**
On external pages that want to show a feed of our most recently published datasets, e.g. [GigaByte journal](https://gigabytejournal.com/) and [GigaScienceJournal](https://academic.oup.com/gigascience) the RSS feed they use no longer works (since we released 4.2.4 on live).

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'https://academic.oup.com/gigascience' or 'https://gigabytejournal.com/'
2. Scroll down to set of 4 RSS feed boxes
![image](https://github.com/gigascience/gigadb-website/assets/6037145/868a2f0f-2230-42eb-8823-2084ebdfd97d)

![image](https://github.com/gigascience/gigadb-website/assets/6037145/290d9248-3d00-4ae4-acf6-d245d5a14d09)

**Additional context**
These were working before the switch to using the AWS deployment. If there is a different way those external sites should be presenting or using our site then we need to inform them what to do.
As far as I am aware, those are the only 2 places using the RSS feed.

rija:
Maybe linked to #1799

rija:
We need to find the URLs being called.
we will also scour the logs of gigadb.org web server for failed API/RSS calls

only1chunts:
@rija the link being used by GigaSciencejournal and gigabytejournal is https://gigadb.org/rss/latest

rija:
Hi @only1chunts,

thanks for supplying the link. 
That link is no longer the correct link for accessing the web site RSS feed
The correct URL is: 
https://gigadb.org/site/feed

kencho51:
Hi @only1chunts,

The RSS feed boxes seem to have been fixed: 
![image](https://github.com/gigascience/gigadb-website/assets/64770635/e0a67897-2899-4c70-8983-616f9b12714d)

![image](https://github.com/gigascience/gigadb-website/assets/64770635/d1c53072-4732-45d3-ad57-7ab51192dcae)

Does that mean GigaSciencejournal and gigabytejournal have updated the RSS URL to `https://gigadb.org/site/feed` from their sides?

only1chunts:
I have no idea what they have done, but I can confirm it is working on both gigasciencejournal.com and gigabyte journal pages again now, so we can close this ticket 👍

pli888:
[...]

PR #2405, opened 2025-07-15
fix: Use Laminas\Feed for RSS feed

alli83:
# Pull request for issue: #1814 

- Use Laminas\Feed to handle RSS feeds

## How to test?

visit https://gigadb.org/rss/latest

Add Add Feed Writer Generator Extension in protected/feed/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


alli83:
in this PR https://github.com/gigascience/gigadb-website/commit/ee40e2c075b65ddd562fe9f7164ca2969ad98d51 you mentioned Suin\RSSWriter . It's used in a service but not in the controller. But the library is not updated anymore since 2017...
You deleted search method etc, extension files etc but there is no replacement. Should we use the yii extension or Laminas\Feed (even if, it's more for yii2) or php natif like XMLwriter?

